PR #22, opened 2017-07-25
Travis fixed

clebeaupin:


.travis.yml
```diff
@@ -2,19 +2,24 @@ esudo: false
...
   only:
   - master
+  - develop
 language: node_js
 node_js:
 - '6'
 before_install:
 - 'pwd && ls -als'
+- curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.23.2
+- export PATH="$HOME/.yarn/bin:$PATH"
```
danielweck: I wonder why this is needed. Yarn is automatically available it seems:
https://github.com/edrlab/r2-streamer-js/blob/develop/.travis.yml

https://travis-ci.org/edrlab/r2-streamer-js

clebeaupin: This is because travis is only shipped with yarn version 0.17 and r2-streamer-js requires yarn >= 0.23 
https://github.com/edrlab/r2-streamer-js/blob/develop/package.json#L26

danielweck: Thanks for the "heads-up", I did not realise TravisCI provided such an old default Yarn version `0.17` (when latest is `0.27.5`).
I am trying this in `r2-streamer-js` (note: no explicit version, should fetch latest):
https://github.com/edrlab/r2-streamer-js/commit/5ff0b49ee5db7648f8e52175e2ce75a15e7cbed6
...let's see:
https://travis-ci.org/edrlab/r2-streamer-js/builds/257732983

danielweck: Works fine :)

PR #41, opened 2018-02-15
reader style and db settings

Scarsniik:


src/renderer/components/ReaderApp.tsx
```diff
@@ -289,36 +528,36 @@ export default class ReaderApp extends React.Component<undefined, ReaderAppState
...
 
-        if (_publication.Spine && _publication.Spine.length) {
-            console.log(_publication.Spine);
+        if (publication.Spine && publication.Spine.length) {
+            // console.log(publication.Spine);
             const links: IStringMap = {};
-            _publication.Spine.forEach((spineItemLink) => {
-                links[spineItemLink.Href] = publicationJsonUrl + "/../" + spineItemLink.Href;
+            publication.TOC.forEach((spineItemLink) => {
+                links[spineItemLink.Title] = publicationJsonUrl + "/../" + spineItemLink.Href;
```
danielweck: `spineItemLink.Title` is not guaranteed to be unique (and in some cases is undefined / null), therefore I would suggest to use the `spineItemLink.Href` key instead.

danielweck:
If I am not mistaken, this PR is now unnecessary as `feature/next-v1` contains its code (manually merged from the associated branch). So I think this PR can be closed...but I am not 100% sure, @Scarsniik @clebeaupin ?

danielweck:
I am doing some housekeeping in the issue tracker. I see that this PR was closed (unmerged), but if I am not mistaken the PR code was manually cherry-picked into the destination branch `feature/next-v1`. I think that my review comment (above) about the `Link.Title` mapping should be addressed, as this is a bug with some EPUBs (which have duplicate or nil titles on their TOC links). See current code:
https://github.com/edrlab/readium-desktop/blob/feature/next-v1/src/renderer/components/ReaderApp.tsx#L545

danielweck:
https://github.com/edrlab/readium-desktop/blob/a1d086da2c4d09e9eb87260414fb9c4923f32ab9/src/renderer/components/ReaderApp.tsx#L545-L547

PR #55, opened 2018-02-16
redux actions

Scarsniik:


src/renderer/components/ReaderApp.tsx
```diff
@@ -151,10 +151,10 @@ const queryParams = getURLQueryParams();
 
 // tslint:disable-next-line:no-string-literal
 const publicationJsonUrl = queryParams["pub"];
-const publicationJsonUrl_ = publicationJsonUrl.startsWith(READIUM2_ELECTRON_HTTP_PROTOCOL) ?
+const newPublicationJsonUrl = publicationJsonUrl.startsWith(READIUM2_ELECTRON_HTTP_PROTOCOL) ?
```
danielweck: This change is incorrect. It is very important that `publicationJsonUrl` remains exactly as passed in the `pub` querystring URL parameter for the `index_reader.html` renderer process. Then, `newPublicationJsonUrl_` (with the underscore suffix) is only used to compute the publication path / filename, by decoding its base64 parameter.

src/renderer/components/ReaderApp.tsx
```diff
@@ -151,10 +151,10 @@ const queryParams = getURLQueryParams();
...
     convertCustomSchemeToHttpUrl(publicationJsonUrl) : publicationJsonUrl;
 
-const pathBase64 = publicationJsonUrl_.replace(/.*\/pub\/(.*)\/manifest.json/, "$1");
+const pathBase64 = newPublicationJsonUrl.replace(/.*\/pub\/(.*)\/manifest.json/, "$1");
```
danielweck: ...and just to be clear: this is also incorrect :)  (please add `_` again, or change the variable name if this is confusing)

clebeaupin: _ is forbidden by linter

PR #58, opened 2018-02-16
preprocessor directives centralization, better HTTP port handling for renderer processes (app/library + reader views)

danielweck:


danielweck:
Hello @clebeaupin could you please check the code diff, and authorize / do the merge?
This is aligned/synched with the latest version of `feature/next-v1`, so this is ready to merge.

This PR brings no change to external API, just internal cleanup of code that handles the WebPack / DefinePlugin preprocessor directives, and as a bonus also: configurable HTTP ports for the renderer processes (library/app and reader views).

I tested `npm run start`, `npm run start:dev` and `npm run start:devex`.

Once this is merged, I can start figuring out how to tie the `r2-xxx-js` "dev" mode detection: currently only runtime `process.env.NODE_ENV`, no build-time override (no preprocessor directives), which is not good for security reasons in the final dist product.

PR #144, opened 2018-05-02
navigator.epubReadingSystem

### LINKED ISSUE #120: Support for EPUB3 `navigator.epubReadingSystem`
This should be implemented in navigator, but app should be able to specify the actual contents of the injected `navigator.epubReadingSystem` object.
http://www.idpf.org/epub/31/spec/epub-contentdocs.html#app-epubReadingSystem

Scarsniik:
fixes #120

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -478,13 +478,13 @@
       "integrity": "sha512-a2+YeUjPkztKJu5aIF2yArYFQQp8d51wZ7DavSHjFuY1mqVgidGyzEQ41JIVNy82fXj8yPgy2vJmfIywgESW6w==",
       "requires": {
         "@types/connect": "3.4.32",
-        "@types/node": "10.0.0"
+        "@types/node": "10.0.2"
       },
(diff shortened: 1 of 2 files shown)


danielweck:
Just a note:

Interestingly, `@types/node` is v10 as a sub-dependency (for other NPM packages) in `package-lock.json`, whereas `readium-desktop` (as well as all the `r2-xxx-js` packages) must stick to v8 (see:
https://github.com/edrlab/readium-desktop/blob/011a75e8663c6d58df6a5f4d298bce2eb99603c9/package.json#L158 ), otherwise TypeScript compilation fails (notably: because of the URL typings).

Thankfully, NPM `node_module` organisation + resolving algorithm means that our own v8 devDependency gets a higher priority.

To be precise, v9 is still compatible, but v8 is logically aligned with Electron's NodeJS runtime v8.2.1: https://github.com/electron/electron/blob/v1.8.6/.node-version

PR #189, opened 2018-12-12
Latest r2-xxx-js packages (ReadiumCSS streamer injection, layout/rendering performance improvements, Base64 encoding/escape fix) + added missing license headers

danielweck:
Now works, see below. Please merge this PR as soon as possible to continue developing using the correct `package-lock.json` references.

> IMPORTANT, DO NOT MERGE! Compiles and runs fine, but I am not able to import EPUBs in the bookshelf, so I cannot test the navigator. I am going to hunt for a regression bug in `package-lock.json` ...

PR #191, opened 2018-12-12
added missing license headers

danielweck:
Note that this PR targets the branch `feature/r2-xxx-js_latest`, which is expected to be merged soon into `feature/new-db` (see https://github.com/readium/readium-desktop/pull/189 ).

If you prefer to target `feature/new-db` directly (and separately), then let's close this PR and delete the associated branch (`feature/r2-xxx-js_latest_license-headers`).

Then, simply use the following command line utility to update the license headers in any other branch:

`cd readium-desktop && node ../r2-utils-js/tools/license_header.js ./src`

Assuming the filesystem:
```
/
|___ readium-desktop
|___ r2-utils-js
```


danielweck:
PR now updated to match the latest `feature/r2-xxx-js_latest` (which itself is sync'ed with `feature/new-db`, ready to be merged).

PR #196, opened 2018-12-19
Several important fixes (see full description)

danielweck:
* Suggested code fix for `store.getState().reader.config.value` in both main and renderer processes (this is undefined/null when the app first runs, and when settings have not been changed by the user yet) ... I added console logging to visualize the problem without debugging.
* `r2-xxx-js` package updates: fixed support for LCP CRL (Certificate Revocation List)
* due to updated `r2-xxx-js` components:  removed useless ReadiumCSS boilerplate (now handled in directly navigator) + slight change in `navigator.epubReadingSystem` API (now "push", not "pull" anymore as the value is static / never changes ... sorry for this non-semver change, but because there is only one consumer app at the moment I wanted to keep it simple)
* added TODO comments about centralizing ReadiumCSS main/renderer store fetch (I will leave this to the Redux/React experts, right now the code works, but there is code duplication, so next time the `renderer` process code is updated, I worry that the `main` process will be out of sync ;)
* in renderer process / BrowserWindow: `window.READIUM.debug(true)` switches on the visual debugging of navigator (reading locations)
* the web console is redirected into shell output for easier debugging, with full debug() coloring support (reason: there is an Electron "preload" bug which makes the whole application crash on exit when the web inspector has been open, so console debugging is a useful alternative option). as usual with console debug: use the blackboxing settings to exclude JS files (same with WebPack logger)
* added missing debug() colors in dev mode (`npm run start:dev`)


danielweck:
I added a proposed fix for missing `debug()` colours, see https://github.com/readium/readium-desktop/issues/197

PR #202, opened 2018-12-29
NPM updates and convertMultiLangStringToString() for Subject.Name

danielweck:
Tested with OPDS2 Feedbooks feed:
https://catalog.feedbooks.com/catalog/index.json

...as well as OPDS1:
https://www.feedbooks.com/catalog.atom

PR #212, opened 2019-01-17
possibility to open lcp publications

Scarsniik:


src/main/converter/opds.ts
```diff
@@ -57,7 +57,7 @@ export class OpdsFeedViewConverter {
         const title = convertMultiLangStringToString(metadata.Title);
         const authors = convertContributorArrayToStringArray(metadata.Author);
         const publishers = convertContributorArrayToStringArray(metadata.Publisher);
-        const tags = metadata.Subject.map((subject) => subject.Name);
+        const tags = metadata.Subject.map((subject) => subject.Name as string);
```
danielweck: This forced type coercion to `string` is incorrect. The correct implementation is in this older Pull Request:
https://github.com/readium/readium-desktop/pull/202

src/main/streamer.ts
```diff
@@ -122,6 +122,8 @@ function computeReadiumCssJsonMessage(publication: Publication, link: Link | und
         typeScale: readiumCSSDefaults.typeScale,
 
         wordSpacing: settings.wordSpacing,
+
+        reduceMotion: false,
```
danielweck: I suggest to use `readiumCSSDefaults.reduceMotion`, as in the original Pull Request: https://github.com/readium/readium-desktop/pull/202

src/main/converter/opds.ts
danielweck: Fixed in commit: https://github.com/readium/readium-desktop/pull/212/commits/5ec111bf7f326759c1e39ac643da82c4786943a7

src/main/streamer.ts
danielweck: Fixed in this commit: https://github.com/readium/readium-desktop/pull/212/commits/74cbb6be8232b8d41e360efd72d2c4cec05a2439

src/common/redux/actions/lcp.ts
```diff
@@ -31,7 +30,7 @@ export enum ActionType {
     StatusUpdateError = "LCP_STATUS_UPDATE_ERROR",
 }
 
-export function checkUserKey(publication: Publication, hint: string): Action {
+export function checkUserKey(publication: PublicationDocument, hint: string): Action {
```
clebeaupin: PublicationDocument does not exist in common.
Use PublicationView

PR #221, opened 2019-03-20
Feature/find tag

manuellagrand:
this pull request include the temporary traduction of each layout and the implementation of tag search form with auto-suggestion


src/common/services/translator.ts
```diff
@@ -28,8 +28,8 @@ const initI18n = i18n.init({
 
 export enum AvailableLanguages {
     en = "English",
-    fr = "Français",
-    de = "Deutch",
+    fr = "French",
```
danielweck: What is the rationale for changing from French to English here?

manuellagrand: the reason is to standardize languages in english which is the default language.

manuellagrand: i made some changes in translation files in order to translate each languages when it is selected in settings

danielweck:
Let's say that I am a Japanese user, the UI is in English which is the default locale when I install the app. Now I need to find the menu / options dialog where I can switch to Japanese. I follow a tutorial to reach the UI (I cannot read English fluently so I follow steps / instructions). When I get there, I expect to see my language in my own script / locale, not English. In my opinion. What do you think?

llemeurfr:
I understand your concern @danielweck. When I change the language in e.g. an iphone, I see both informations, ex. "Espanõl" (target language) + (below) "Spanish" or "Espagnol" (i.e. the current language). We'll work on obtaining that.

src/renderer/components/catalog/GridView.tsx
```diff
@@ -52,8 +51,16 @@ export default class GridView extends React.Component<GridViewProps, undefined>
                             </section>
                         );
                 })}
-                <AddEntryForm />
+            <button className={styles.tag_add_button}>
+                <span>
+                    <Link to="/settings/tags" style={{color: "#1a1a1a"}}>
```
clebeaupin: Do not put inline styles

src/renderer/components/catalog/AddEntryForm.tsx
```diff
@@ -7,86 +7,112 @@
...
+        console.log("tag: " + this.state.tag);
+    }
     private submit(e: any) {
+
+        if (this.state.tag === "") {
+            alert("tag input is empty.");
+            return;
+        }
         e.preventDefault();
+        console.log("props" + this.props);
```
clebeaupin: Remove all logs
Use log for debugging but do not commit them

src/renderer/components/catalog/AddEntryForm.tsx
```diff
@@ -104,12 +130,19 @@ export default withApi(
...
-                moduleId: "publication",
-                methodId: "updateTags",
+                moduleId: "catalog",
+                methodId: "updateEntries",
             },
         ],
     },
 );
+
+export const translate = withTranslator(AddEntryForm);
```
clebeaupin: translator is already included in withApi so you don't need it to use it

src/renderer/components/catalog/AddEntryForm.tsx
```diff
@@ -7,86 +7,112 @@
...
+    addEntry?: (entry: any) => void;
+    getentry?: CatalogEntryView[];
+    tags?: string[];
 }
 
 export class AddEntryForm extends React.Component<AddEntryFormProps, AddEntryFormState> {
     private selectRef: any;
+    // How to translate temporarily
+    @lazyInject("translator")
+    private translator: Translator;
```
clebeaupin: Stop using lazy inject.
Instead use withApi or withTranslator HOC

src/renderer/components/catalog/ListView.tsx
```diff
@@ -12,24 +12,32 @@ import { CatalogEntryView } from "readium-desktop/common/views/catalog";
...
 import PublicationListElement from "readium-desktop/renderer/components/publication/PublicationListElement";
 
-import AddEntryForm from "./AddEntryForm";
+import { Link } from "react-router-dom";
 
 import * as styles from "readium-desktop/renderer/assets/styles/myBooks.css";
+import { TranslatorProps, withTranslator } from "../utils/translator";
 
-interface ListViewProps {
+import { lazyInject } from "readium-desktop/renderer/di";
```
clebeaupin: Remove lazyInject if it used to autowired Translator

PR #226, opened 2019-03-25
Feature/all books design

manuellagrand:
removal of "allbooks" button in next of each selection & move it to sub navigation bar, below "Preferences".


llemeurfr:
Manuel merged by mistake. Seems rights are still not in place.

PR #234, opened 2019-04-16
Version bumps for r2-xxx-packages, now compatible with Electron 3 and 4

danielweck:
Known remaining problem since migration to Electron v1+ (regression bugs / breaking changes in v3 and v4): single key stroke shortcuts for "page turn" (arrow left/right) => app-level events are not raised when keyboard focus is inside the webview (which is now in fact an out-of-process iframe inside a shadow DOM inside the Electron's webview tag). I will fix this in the next `r2-navigator-js` release.
Otherwise, CSS column pagination and scroll mode now works again, and most keyboard / mouse handling seems to work as before.


danielweck:
I have now updated this PR to match the latest `develop` branch, and to import the latest `r2-xxx-js` packages. I tested `readium-desktop` with this PR. Notably, there is a fix for the Electron / webview regression for left/right arrow keys when the keyboard focus is inside the publication document. There are other Electron/Chromium regressions in fixed-layout FXL documents, but I have not yet found workarounds (this will be addressed in a future package update).

danielweck:
I've just updated the latest `r2-xxx-js` packages, notably: now the "Resize Sensor" dependency (`css-element-queries`) is fetched from the proper NPM registry (used to be GitHub repository).
Looking at the full `package-lock.json`, I see one remaining GitHub dep: `github:danielzzz/node-ping`. It looks like we are now pretty much free of NPM-related bugs then ;)

danielweck:
FYI, with this PR I am getting this TypeScript error (probably due to a change in `package-lock.json`):
```
[2] ERROR in [at-loader] ./src/renderer/components/settings/CatalogEntrySettings.tsx:172:13 
[2]     TS2326: Types of property 'entryToUpdate' are incompatible.
[2]   Type '{ id: number; title: string; tag: string; }' is not assignable to type '{ id: number; title: string; }'.
[2]     Object literal may only specify known properties, and 'tag' does not exist in type '{ id: number; title: string; }'.
```

danielweck:
TypeScript `3.5.1` stricter rules, now fixed. See:
https://github.com/readium/readium-desktop/commit/3e36af097c7d6f5d1aa1da60fd1a54d1552e5d0c

PR #253, opened 2019-04-30
add translations in reader settings

manuellagrand:
this PR contains the translation of default font in reader settings.


src/renderer/components/reader/ReaderOptions.tsx
```diff
@@ -412,6 +412,13 @@ export default class ReaderOptions extends React.Component<Props, State> {
         );
     }
 
+    private checkFontName(font: string, __: any): string {
+        if (font && font === "Default font") {
```
clebeaupin: Could you use a constants for "Default font"?

const DEFAULT_FONT_NAME = "Default font"

Put this constant ahead of your class

PR #269, opened 2019-05-28
add aria-label translation

manuellagrand:
This is the add of translation of aria-label


src/renderer/components/opds/SearchForm.tsx
```diff
@@ -12,8 +12,11 @@ import * as styles from "readium-desktop/renderer/assets/styles/header.css";
 import SVG from "readium-desktop/renderer/components/utils/SVG";
 
 import { RouteComponentProps, withRouter } from "react-router-dom";
+import { TranslatorProps, withTranslator } from "../utils/translator";
```
clebeaupin: Pas d'import relatif

PR #311, opened 2019-06-15
[fix] information component: async loading md + replacing static version in variable

panaC:
In information component the information locale md is loaded in sync. I have added promise to load file with async method.

I changed the version number by dynamic print with version in package.json

test on ubuntu lts 18 and not on mac-os/win


src/renderer/components/settings/Information.tsx
```diff
@@ -17,10 +17,11 @@ import { setLocale } from "readium-desktop/common/redux/actions/i18n";
 
 import { TranslatorProps, withTranslator } from "readium-desktop/renderer/components/utils/translator";
 
-import * as fs from "fs";
-
 import * as commonmark from "commonmark";
 
+import * as packageJson from "readium-desktop/package.json";
+import { readFile } from "readium-desktop/utils/readFileAsync";
```
clebeaupin: You should use util.promisify()

panaC: thank you, I used it now

PR #330, opened 2019-06-20
change no tag message

manuellagrand:
Change the message that indicates no tag have not still added to the book


danielweck:
@llemeurfr shall we merge this trivial code change?

PR #331, opened 2019-06-20
correct some translation

### LINKED ISSUE #286: Improper label translations
In the book information windows:

In english, "language" -> "Language"; "identifier" -> "Identifier". 
In french, "langue" -> "Langue"; "identifiant" -> "Identifiant".

manuellagrand:
Correct some translation & added missing translation

fix #286

### DIFF
--- a/src/resources/locales/en.json
+++ b/src/resources/locales/en.json
@@ -102,7 +102,9 @@
       "return": "Do you really want to return the licence for this publication?",
       "deleteOpds": "Do you really want to delete this opds feed?",
       "renew": "Do you really want to renew the licence for this publication?",
-      "closeModalWindow": "Close this modal window"
+      "closeModalWindow": "Close this modal window",
+      "alreadyAdd": "It seems that you have already added this publication to your library.",
+      "sure": "Are you sure you want to add it again ?"
     },
     "opds": {
       "download": "Download",
@@ -135,11 +137,11 @@
       },
       "menu": {
         "aboutBook": "About this book",
-        "addLib": "Add this book to the librairy",
+        "addLib": "Import the extract",
         "goBuyBook": "Buy this book",
         "goLoanBook": "Borrow this book",
         "goSubBook": "Subscribe",
-        "addTeaser": "Add extract to the librairy"
+        "addTeaser": "Add extract to the library"
       },
       "breadcrumbRoot": "Catalogs"
     },
@@ -163,7 +165,7 @@
         "title": "Settings",
         "theme": {
           "title": "Theme",
-          "predefined": "Predefined themes",
+          "predefined": "Predifined themes",
           "myTheme": "My themes",
           "create": "Create a theme",
           "name": {
@@ -174,18 +176,18 @@
         },
         "text": "Text",
         "left": "Left",
-        "justify": "Justified",
+        "justify": "Justify",
         "column": {
           "title": "Column",
           "auto": "Automatic",
           "one": "1 column",
-          "two": "2 columns"
+          "two": "2 column"
         },
         "spacing": "Spacing",
         "margin": "Margin",
-        "wordSpacing": "Word spacing",
+        "wordSpacing": "Words spacing",
         "lineSpacing": "Line spacing",
-        "letterSpacing": "Letter spacing",
+        "letterSpacing": "Letters spacing",
         "accessibility": "Accessibility",
         "disposition": {
           "title": "Disposition",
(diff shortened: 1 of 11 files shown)


src/resources/locales/en.json
```diff
@@ -135,11 +138,11 @@
       },
       "menu": {
         "aboutBook": "About this book",
-        "addLib": "Add this book to the librairy",
+        "addLib": "Import the extract",
```
danielweck: Why "extract" and not "book"?

src/resources/locales/en.json
```diff
@@ -135,11 +138,11 @@
...
         "goBuyBook": "Buy this book",
         "goLoanBook": "Borrow this book",
         "goSubBook": "Subscribe",
-        "addTeaser": "Add extract to the librairy"
+        "addTeaser": "Add extract to the library"
```
danielweck: What's the difference with `"addTeaserToLib": "Add the extract book to library"`?

src/resources/locales/en.json
```diff
@@ -163,7 +166,7 @@
         "title": "Settings",
         "theme": {
           "title": "Theme",
-          "predefined": "Predefined themes",
+          "predefined": "Predifined themes",
```
danielweck: The correct spelling is `Predefined`

src/resources/locales/en.json
```diff
@@ -174,18 +177,18 @@
...
         "text": "Text",
         "left": "Left",
-        "justify": "Justified",
+        "justify": "Justify",
         "column": {
           "title": "Column",
           "auto": "Automatic",
           "one": "1 column",
-          "two": "2 columns"
+          "two": "2 column"
```
danielweck: Shouldn't this remain a plural word?

src/resources/locales/en.json
llemeurfr: A matter of available space; one is in the "book info" view (thre is space) they other in a more constrained space.

src/resources/locales/en.json
llemeurfr: This is used only for extracts. The issue us the "addLib" property name which does not reflect that.

src/resources/locales/en.json
llemeurfr: you're right, I didn't request this. @manuellagrand ?

src/resources/locales/en.json
llemeurfr: For sure!

src/resources/locales/en.json
danielweck: okay. thank you.

src/resources/locales/en.json
```diff
@@ -174,18 +177,18 @@
...
         },
         "spacing": "Spacing",
         "margin": "Margin",
-        "wordSpacing": "Word spacing",
+        "wordSpacing": "Words spacing",
         "lineSpacing": "Line spacing",
-        "letterSpacing": "Letter spacing",
+        "letterSpacing": "Letters spacing",
```
llemeurfr: Bad move. I don't think it is plural. @danielweck ?

src/resources/locales/en.json
danielweck: I see thank you. Shall we leave the l10n key as it is, to avoid modifying consumer code at this stage?

src/resources/locales/en.json
danielweck: I think singular works better, both as a technical term and as a human-readable label.
https://developer.mozilla.org/en-US/docs/Web/CSS/word-spacing
https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing

src/resources/locales/en.json
```diff
@@ -174,18 +177,18 @@
...
           "title": "Column",
           "auto": "Automatic",
           "one": "1 column",
-          "two": "2 columns"
+          "two": "2 column"
         },
         "spacing": "Spacing",
         "margin": "Margin",
-        "wordSpacing": "Word spacing",
+        "wordSpacing": "Words spacing",
```
danielweck: I think singular works better, both as a technical term and as a human-readable label.
https://developer.mozilla.org/en-US/docs/Web/CSS/word-spacing
https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing

src/resources/locales/en.json
llemeurfr: Yes, we'll make a tour of badly named i10n properties later, there are several which don't make sense in English.

danielweck:
Note: this PR introduced a bug (typo in locale resource key):
https://github.com/readium/readium-desktop/issues/431

PR #351, opened 2019-06-20
fixes keyboard focus outline inconsistencies, differentiates mouse (pointing device) vs. keyboard interaction

danielweck:
WIP Work in Progress ...

Seems to all work fine as an improvement to the current method, but I am not totally happy with outline clipping (this problem exists in the current implementation too)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1096,9 +1096,9 @@
       "dev": true
     },
     "@types/react": {
-      "version": "16.8.21",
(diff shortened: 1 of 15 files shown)


danielweck:
I am going to merge this, as the only remaining issue (outline clipping) already exists with the current codebase (this will be fixed in a separate PR that addresses UI component spacing)

PR #352, opened 2019-06-21
[fix #257] close library and reader when detached and close button clicked

### LINKED ISSUE #257: Closing  the app should close detached windows
when the home window is detached of the reader window and we close the home window, it should close all windows open

panaC:
I have
- ordered import
- set type for mainWindow
- fix closeCallback

fixes #257 

test on ubuntu lts 18


panaC:
see PR of Manu

Issue #334, opened 2019-06-20
Keyboard left/right arrow keys: UI must capture interaction, otherwise the page turn command is triggered


danielweck:
This is especially noticeable in the right-hand-side slide-out panel with UI controls such as the horizontal sliders. But in fact, the keyboard event listener for left/right arrow keys should not react at all when popup dialogs or overlay menus are presented.

danielweck:
https://github.com/readium/readium-desktop/blob/97905cc5d39aead1e0634bb338525c03bc27fc24/src/renderer/components/reader/Reader.tsx#L311-L319

danielweck:
The usual web/DOM techniques of `preventDefault` and `stopPropagation` should be used to control this behaviour. Basically, UI controls that consume arrow key events and that overlay other controls should capture the user input and not let it propagate into lower-down consumers.

danielweck:
Fix for this issue is in preparation: https://github.com/readium/readium-desktop/pull/360

PR #360, opened 2019-06-22
WIP: fixes the left/right arrow keyboard events the reader view

danielweck:
Fixes https://github.com/readium/readium-desktop/issues/334

=> disables keyboard event listener when side menus are open, and modal popup dialog too (although this still needs some work, because we need to detect dialog close in order to restore keyboard arrows)

### DIFF
--- a/src/renderer/components/dialog/DialogManager.tsx
+++ b/src/renderer/components/dialog/DialogManager.tsx
@@ -228,6 +228,11 @@ export class DialogManager extends React.Component<DialogManagerProps, undefined
 const mapDispatchToProps = (dispatch: any) => {
     return {
         closeDialog: () => {
+            // TODO: this is a short-term hack.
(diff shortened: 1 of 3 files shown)


danielweck:
This solution works, but the implementation needs to be improved. In a nutshell: a custom DOM event named `"Thorium:DialogClose"` is dispatched on the `window.document` specific to the BrowserWindow instance within which a popup modal dialog is closed. Just like with the slide-out menus on the right hand side, this allows setting the local component state to bypass keyboard left/right arrow events.
A more permanent solution would be to subscribe to Redux action type `ActionType.CloseRequest`, but I am not sure how to implement this so that the scope is limited to a single window / reader view instance (i.e. not application-wide). @clebeaupin @Scarsniik what do you think?

danielweck:
I am merging this fix. Not an ideal solution from an architectural standpoint, but works perfectly fine otherwise.

PR #378, opened 2019-07-01
[fix #329] rm publication dir

### LINKED ISSUE #329: Removed publications are not deleted from the filesystem
For example on Windows10:
`C:\Users\danie\AppData\Roaming\EDRLab.ThoriumReader\publications\bc908720-7292-4019-b0d0-806fa2ce6258\book.epub`
...or MacOS:
`/Users/danielweck/Library/Application Support/EDRLab.ThoriumReader/publications/b8f0bb5f-f74c-4b30-ad6d-f6845c21df39/book.epub`

=> the cover image (if any), the `book.epub` and the containing folder (UUID) remain in the app-private storage area, so it looks like this Thorium-specific space will grow indefinitely.

panaC:
- transform rmDir to async version
- link delete fct in publication

fixed #329

### DIFF
--- a/src/main/api/publication.ts
+++ b/src/main/api/publication.ts
@@ -47,7 +47,7 @@ export class PublicationApi {
 
     public async delete(data: any): Promise<void> {
         const { identifier } = data;
-        await this.publicationRepository.delete(identifier);
+        await this.catalogService.deletePublication(identifier);
     }
 
     public async findAll(): Promise<PublicationView[]> {
--- a/src/main/storage/publication-storage.ts
+++ b/src/main/storage/publication-storage.ts
@@ -15,7 +15,7 @@ import { streamToBufferPromise } from "@r2-utils-js/_utils/stream/BufferUtils";
 import { IZip } from "@r2-utils-js/_utils/zip/zip.d";
 
 import { File } from "readium-desktop/common/models/file";
-import { getFileSize, rmDirSync } from "readium-desktop/utils/fs";
+import { getFileSize, rmDirAsync } from "readium-desktop/utils/fs";
 
 import { PublicationView } from "readium-desktop/common/views/publication";
 
@@ -73,8 +73,8 @@ export class PublicationStorage {
         return files;
     }
 
-    public removePublication(identifier: string) {
-        rmDirSync(this.buildPublicationPath(identifier));
+    public async removePublication(identifier: string) {
+        await rmDirAsync(this.buildPublicationPath(identifier));
     }
 
     public getPublicationEpubPath(identifier: string): string {
--- a/src/main/storage/publication-storage.ts
+++ b/src/main/storage/publication-storage.ts
@@ -15,7 +15,7 @@ import { streamToBufferPromise } from "@r2-utils-js/_utils/stream/BufferUtils";
 import { IZip } from "@r2-utils-js/_utils/zip/zip.d";
 
 import { File } from "readium-desktop/common/models/file";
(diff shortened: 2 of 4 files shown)


src/main/api/publication.ts
```diff
@@ -47,7 +47,7 @@ export class PublicationApi {
 
     public async delete(data: any): Promise<void> {
         const { identifier } = data;
-        await this.publicationRepository.delete(identifier);
+        await this.catalogService.deletePublication(identifier);
     }
```
danielweck: I think we should ask @clebeaupin about this code / logic change. I am concerned that there might have been a specific reason to invoke `publicationRepository.delete()` instead of `catalogService.deletePublication()`.

src/main/storage/publication-storage.ts
```diff
@@ -15,7 +15,7 @@ import { streamToBufferPromise } from "@r2-utils-js/_utils/stream/BufferUtils";
 import { IZip } from "@r2-utils-js/_utils/zip/zip.d";
 
 import { File } from "readium-desktop/common/models/file";
-import { getFileSize, rmDirSync } from "readium-desktop/utils/fs";
+import { getFileSize, rmDirAsync } from "readium-desktop/utils/fs";
```
danielweck: Can you explain the motivation for switching from the `sync` to the `async` `rmDir` variant in the call to `removePublication()`? Was that a *necessary* change in addition to switching from `publicationRepository.delete()` to `catalogService.deletePublication()`? Did you test on Windows, Mac and Linux? (filesystem operations are often inconsistent from one OS to another)

src/main/api/publication.ts
danielweck: @panaC in your estimation, is there a risk of regression because of the switch to a different API? Or do you think this is a safe code / logic change?

src/main/storage/publication-storage.ts
danielweck: Naturally, I understand the impact of "sync" vs. "async" in terms of blocking the Node main thread during the few cycles that are necessary to delete an entire folder (albeit in this particular case: just 3 files, so probably insignificant).

So, here I am asking specifically about fixing the bug: https://github.com/readium/readium-desktop/issues/329
Reading your comment in the issue discussion, it looks like you also came to the conclusion that strictly-speaking, the move from "sync" to "async" was not necessary to fix the bug, and that in fact only the API switch from `publicationRepository.delete()` to `catalogService.deletePublication()` was needed. Is that correct?

panaC: I removed my works

src/main/api/publication.ts
panaC: I think it's a logic change 😐

danielweck: Sure, but do you think it is a *safe* one? :)
There must be a reason why Leo or Cyrille decided to use the original API, instead of the one you are replacing it with. Thus why I am concerned about potential regression bugs. We need to assert that this is a safe change with no nasty side effects.
Ping @clebeaupin @Scarsniik

src/main/storage/publication-storage.ts
danielweck: I see, okay. So just to be sure: the bug is fixed *without* the need to switch to async versions of the filesystem functions, right?

panaC: yes it's works

danielweck:
API changed reviewed by @clebeaupin (thank you) ... let's merge! :)

PR #379, opened 2019-07-01
[fix #367] export align in list-menu

### LINKED ISSUE #367: "export" button is horizontally offset relative to the "about" and "delete" buttons in list view
ThoriumExportButton

panaC:
I changed only 2 lines of css


danielweck:
I tested this, it seems to work just fine without any regressions elsewhere (the `list_menu` class is only used in one UI place).
Merging.

PR #380, opened 2019-07-01
[fix #369] change i18n bookmarks

### LINKED ISSUE #369: Bookmarks are displayed in the EPUB "landmarks" nav list, they should have their own reserved section
ThoriumBookmarksNav

panaC:


src/renderer/components/reader/ReaderMenu.tsx
```diff
@@ -78,7 +78,7 @@ export class ReaderMenu extends React.Component<Props, State> {
                 disabled: !publication.LOI || publication.LOI.length === 0,
             },
             {
-                title: __("reader.marks.landmarks"),
+                title: __("reader.marks.bookmarks"),
                 content: this.createLandmarkList(),
```
danielweck: The function `this.createLandmarkList()` seems to produce a list of *bookmarks*, which is inconsistent with the function name.

danielweck:
There is some confusion about the actual bug in the current UI.

In EPUB, there are 3 major types of authored navigation structures:
* table of contents (TOC)
* page list (authored list of `epub:type` page-breaks)
* landmarks (which can be `epub:type` list-of-illustration, etc.)

Then, reading systems add their own "bookmarks" and "annotations" lists, which are not authored in the original publications, but added as navigation overlay. See the screenshots of Thorium and r2-testapp-js in https://github.com/readium/readium-desktop/issues/369

src/renderer/components/reader/ReaderMenu.tsx
danielweck: TODO in separate issue

Issue #359, opened 2019-06-22
"Fiche info" popup dialog is not modal (mouse interaction above the blurred greyed area is allowed)


danielweck:
Note how I was able to switch to a different menu ("preferences") using the top UI:

![ThoriumModalDialog](https://user-images.githubusercontent.com/625201/59966632-fe503800-9516-11e9-9a24-2f526b524c89.png)

danielweck:
Similar issue in the reader view, here shown with the right-hand-side slide-out menu overlaying the modal dialog.

![ThoriumModalDialogMenuOverlay](https://user-images.githubusercontent.com/625201/59966816-728bdb00-9519-11e9-92bc-04f64c032690.png)

panaC:
related to #370

Scarsniik:
fixed in 416b6ae

danielweck:
Thanks @Scarsniik ! :)
However, if I am not mistaken your fix is only for the library window, could you please check the reader view as well? (see the above screenshot, which shows the slide-out right-hand-side menu incorrectly displaying over the modal popup dialog)

danielweck:
I posted review comments in the PR:
https://github.com/readium/readium-desktop/pull/381

Scarsniik:
Yes, it works with the reader

PR #381, opened 2019-07-01
fix modal

Scarsniik:


danielweck:
Fixes https://github.com/readium/readium-desktop/issues/359

src/renderer/components/layout/LibraryLayout.tsx
```diff
@@ -46,21 +46,22 @@ class LibraryLayout extends React.Component<LibraryLayoutProps, undefined> {
         }
 
         return (
-            <>
+            <div
+                style={ this.props.dialogOpen ? {filter: "blur(1px)"} : {} }
+            >
```
danielweck: if I am not mistaken, this fix is only for the library window, could you please check the reader view as well? See the following screenshot which shows the slide-out right-hand-side menu incorrectly displaying over the modal popup dialog: https://github.com/readium/readium-desktop/issues/359#issuecomment-504683032

src/renderer/assets/styles/bookDetailsDialog.css
```diff
@@ -16,7 +16,7 @@
 
 .c_dialog_background {
     position: fixed;
-    z-index: 101;
+    z-index: 102;
```
danielweck: should a similar implementation strategy be used for all popup modal dialogs? (for example, yes/no/cancel confirmations, LCP passphrase input, etc.)

src/renderer/components/layout/LibraryLayout.tsx
danielweck: @Scarsniik said: "Yes, it works with the reader"
https://github.com/readium/readium-desktop/issues/359#issuecomment-507579537

danielweck: Tested by me, works fine in reader view too.

src/renderer/assets/styles/bookDetailsDialog.css
danielweck: I tested a few dialogs, it all seems to work consistently.

PR #401, opened 2019-07-09
reader slide-out menu, radio buttons for page layout settings

### LINKED ISSUE #350: display settings (in reader view) should behave like language choice and theme selection (radio with single active option)
Right now, the column etc. preferences are accessible via keyboard tab navigation, instead of arrow key which is standard practice for radio "buttons"

manuellagrand:
this PR is about the change of the display section in reader menu in order to make it more accessible

fixes #350


src/renderer/components/reader/ReaderOptions.tsx
```diff
@@ -197,90 +197,131 @@ export class ReaderOptions extends React.Component<Props> {
...
+                <ul className={styles.center_in_tab}>
+                    <li>
+                        <label htmlFor={"radio-" + styles.option_gauche}
+                        className={this.getButtonClassName("align", "left")}>
+                            <SVG svg={LeftIcon} title={__("reader.settings.left")}/>
+                            {__("reader.settings.left")}
+                        </label>
+                        <input
+                        id={"radio-" + styles.option_gauche}
+                        name="alignement"
```
danielweck: English is `alignment` :) (no `e`)

src/renderer/components/reader/ReaderOptions.tsx
```diff
@@ -197,90 +197,131 @@ export class ReaderOptions extends React.Component<Props> {
...
+                    </li>
+                    <li>
+                        <label htmlFor={"radio-" + styles.option_justif}
+                        className={this.getButtonClassName("align", "justify")}>
+                            <SVG svg={JustifyIcon} title={__("reader.settings.justify")}/>
+                            {__("reader.settings.justify")}
+                        </label>
+                        <input
+                        id={"radio-" + styles.option_justif}
+                        name="alignement"
```
danielweck: English is `alignment` :) (no `e`)

src/renderer/components/reader/ReaderOptions.tsx
danielweck: PS: I am aware that there are many other French words (such as `option_gauche` and `option_colonne`), and it must be very tempting to fix this now in these small PRs ... but hopefully this will be addressed in a future refactoring :)

danielweck:
I tested this PR, see screenshot below:

ThoriumPaginationLayoutStyleRadios

Is it possible to hide the native radio buttons, like with the language and colour selectors?

Thanks! :)

Issue #335, opened 2019-06-20
Error when pressing escape to close the right-hand-side slide-out menu


danielweck:
```
AccessibleMenu.tsx:99 Uncaught TypeError: this.props.focusMenuButton is not a function
    at AccessibleMenu.handleKey (AccessibleMenu.tsx:99)
handleKey @ AccessibleMenu.tsx:99`
```
![ThoriumFocusEscapeKeyError](https://user-images.githubusercontent.com/625201/60912428-34aee680-a27d-11e9-9475-f3c78fa8d15a.png)

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/336

panaC:
fixed

danielweck:
Hello @panaC the bug still exists in the develop branch, could you please specify the PR that addresses this issue? Thanks! :)

panaC:
![Screenshot 2019-07-03 at 17 51 36](https://user-images.githubusercontent.com/3198865/60606240-4b57c800-9dbb-11e9-8fd6-28b01fb8af85.png)

my fault, sorry, but what is the purpose of this function ? @Scarsniik

PR #402, opened 2019-07-09
catch focus on tab menu reader when is closed

### LINKED ISSUE #336: After pressing escape to close the right-hand-side slide-out menu, outline is not shown on currently focused button
Related issue?
https://github.com/readium/readium-desktop/issues/335

manuellagrand:
this PR allow to find back the focus in reader when we closed settings and navigation menu

fixes #336

### DIFF
--- a/src/renderer/components/reader/ReaderHeader.tsx
+++ b/src/renderer/components/reader/ReaderHeader.tsx
@@ -34,6 +34,8 @@ import { TranslatorProps, withTranslator } from "readium-desktop/renderer/compon
 import ReaderMenu from "./ReaderMenu";
 import ReaderOptions from "./ReaderOptions";
 
+import {createRef} from "react";
+
 interface Props extends TranslatorProps {
     menuOpen: boolean;
     mode?: ReaderMode;
@@ -52,10 +54,10 @@ interface Props extends TranslatorProps {
 }
 
 export class ReaderHeader extends React.Component<Props, undefined> {
-    private enableFullscreenRef: any;
-    private disableFullscreenRef: any;
-    private catchSettingFocus: any;
-    private catchMenuFocus: any;
+    private enableFullscreenRef = createRef<HTMLButtonElement>();
+    private disableFullscreenRef = createRef<HTMLButtonElement>();
+    private catchSettingFocus = createRef<HTMLButtonElement>();
+    private catchMenuFocus = createRef<HTMLButtonElement>();
 
     public constructor(props: Props) {
         super(props);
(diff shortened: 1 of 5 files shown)


danielweck:
Hello, does this fix this issue too?
https://github.com/readium/readium-desktop/issues/335

src/renderer/components/reader/ReaderHeader.tsx
```diff
@@ -54,6 +54,8 @@ interface Props extends TranslatorProps {
 export class ReaderHeader extends React.Component<Props, undefined> {
     private enableFullscreenRef: any;
     private disableFullscreenRef: any;
+    private catchSettingFocus: any;
+    private catchMenuFocus: any;
```
danielweck: To avoid using `any`, could we instead declare the type of the private members by assigning them a proper "ref", for example `private catchMenuFocus = createRef<HTMLElement>();` (`import { createRef } from 'react'`), then instead of `ref={(ref) => this.catchMenuFocus = ref}` in the JSX DOM, we could simply use `ref={this.catchMenuFocus}`, and the usage of the private member can be safeguarded by checking for undefined/null `current` value, e.g.: `if (this.catchMenuFocus.current) { this.catchMenuFocus.focus(); }`. What do you think?

src/renderer/components/reader/ReaderHeader.tsx
```diff
@@ -54,6 +54,8 @@ interface Props extends TranslatorProps {
 export class ReaderHeader extends React.Component<Props, undefined> {
     private enableFullscreenRef: any;
     private disableFullscreenRef: any;
+    private catchSettingFocus: any;
```
danielweck: For consistency with `enableFullscreenRef` and `disableFullscreenRef`, I suggest `catchSettingFocus` and `catchMenuFocus` are renamed with the `xxRef` suffix naming convention. For example: `navigationMenuButtonRef` and `settingsMenuButtonRef`. Thoughts?

danielweck:
Regarding issue #335, I can now confirm that the bug persists:

ThoriumFocusEscapeKeyError

However I tried your PR and it seems to work really well (I tested hybrid mouse / keyboard interaction in order to hit potential edge cases, but it all seems solid).

src/renderer/components/reader/ReaderHeader.tsx
manuellagrand: yes it's more relevant so i changed  my variables name

danielweck:
Thank you for also addressing #335. Now I am wondering:
Both the `ReaderMenu` and `ReaderOptions` menus (which are composed from `ReaderHeader`) are built with `SideMenu`, which is itself built with `AccessibleMenu`. The same dependency chain exists for `Menu` -> `MenuContent` -> `AccesibleMenu`. The difference is that with the latter, `this.props.focusMenuButton` is passed along the chain of React components, whereas with the former, it isn't. This PR is creating ad-hoc logic to handle the "re-focusing" on the calling site for the popup / slide-out menus (i.e. buttons), but shouldn't we instead leverage the existing framework feature `this.props.focusMenuButton`? The only existing concrete implementation of this function is:
https://github.com/readium/readium-desktop/blob/23a0b6438f78e3096f6fb808cf46a64127e771f2/src/renderer/components/utils/menu/Menu.tsx#L124-L128
...would a similar approach work as a replacement for the ad-hoc solution in this PR?
If so, we could avoid duplicating logic, and instead re-use existing building blocks.

PR #403, opened 2019-07-10
Command Line Interface

### LINKED ISSUE #85: Opt-in EPUB file extension association ("double-click from filesystem to open publication")
CBZ too?

This should work on all 3 target distribution platforms Windows, Linux and MacOS

panaC:
Add a CLI on Readium

- allow to import and open a book with a path
  - Works with 'open with' on windows
  - fixes #85 -> add 'FileAssocitation'in package.json

- allow to import a new opds feed with title and url
- allow to read a publication with a partial title search
- allow to import a new publication with silent mode
- silent mode to import a feed or a publication without open window app

Test on W10 : OK
I added 'msi' target in package.json to test on windows install 'open with'
Test on Mac-OS : OK with `npx electron .` and not with the release package (error when installation)

-- 11/07

### Test

Test on W10 : The MSI installer works.
   - To read an epub publication : right click to epub file and select open-with 'Thorium reader'

Test on MAC : 
   -  The build on MAC dosen't work but with npm run start:dev : it's work

No still test on Linux

### CLI

Thorium_reader /my/epub 
  - imports and read my epub

Thorium_reader --import /my/epub
  - imports and opens library window

Thorium_reader --import /my/epub --silent
  - imports and do not opens window

Thorium_reader --opds title=url
  - imports opds url with the name supply or per default print the TLD
  - with --silent don't open the window

Thorium_reader --read 'mon title''
  - opens the first publication returned by a database search

---

update 23/07

new CLI :

```
Thorium Reader <option> [args]

Commands:
  Thorium Reader opds <title> <url>  import opds feed
  Thorium Reader import <path>       import epub or lpcl file
  Thorium Reader read <title>        searches already-imported publications with
                                     the provided TITLE, and opens the reader
                                     with the first match
  Thorium Reader [path]              default command                   [default]

Positionals:
  path  publication in absolute or relative path                        [string]

Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]

```

I have test on W10 and MAC-OS


danielweck:
Great!
Could you please explain "allow to read a publication with a partial title"?

danielweck:
> I added 'msi' target in package.json to test on windows install 'open with'

danielweck:
> I added 'msi' target in package.json to test on windows install 'open with'

Note that using the `NSI` target for Windows would give us auto-update functionality, so perhaps a better choice than `MSI`. See:

https://www.electron.build/auto-update

Also note that although code-signing for `APPX` / Windows store deployment is not needed (thanks to the Microsoft developer account which is already a trusted software provider), code-signing would however be highly-desirable for `MSI`/`NSI` installers, in order to remove the security warning that users get when installing Thorium.

https://www.electron.build/code-signing

package.json
```diff
@@ -98,11 +98,23 @@
...
+    "fileAssociations": [
+      {
+        "ext": "epub",
+        "role": "Viewer",
+        "mimeType": "application/epub+zip"
+      },
+      {
+        "ext": "lcpl",
+        "role": "Viewer",
+        "mimeType": "application/lcpl+zip"
```
danielweck: Correction: `application/vnd.readium.lcp.license.v1.0+json`

panaC: fixed

panaC:
ready to merge

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,120 @@
...
+            // extract and save the title and url from opdsFeed
+            // title=http://myurl.com or get TLD and set url
+            const feed = (argv.opds as string).split("=");
+            const url = feed.length === 2 ? feed[1] : feed[0];
+            const title = feed.length === 2 ? feed[0] : extractHostname(url, true);
+            const opdsRepository = container.get("opds-feed-repository") as OpdsFeedRepository;
+            return await opdsRepository.save({ title, url });
+        },
+        help: [
+            "--opds URL",
```
danielweck: Shouldn't this be `--opds TITLE=URL`?

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,120 @@
...
+            "stay on command line and don't open main window",
+        ],
+    },
+    {
+        name: "read",
+        fct: async ({ argv }) => {
+            // read the publication name
+            const publicationRepo = container.get("publication-repository") as PublicationRepository;
+            const publication = await publicationRepo.searchByTitle(argv.read as string);
+            if (publication && publication.length) {
```
danielweck: I think we should display a failure message, for when the publication cannot be found based on its title.

For example:
```
process.stdout.write(`There is not publication title match for "${argv.read}"\n`);
```

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,120 @@
...
+                    payload: {
+                        publication: {
+                            identifier: publication[0].identifier,
+                        },
+                    },
+                });
+            }
+        },
+        help: [
+            "--read STRING",
```
danielweck: What about `--read TITLE`?

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,120 @@
...
+            "open the reader on your publication",
```
danielweck: Suggestion - more explicit message:
"searches already-imported publications with the provided TITLE, and opens the reader with the first match"

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,120 @@
...
+    },
+    {
+        name: "import",
+        fct: async ({ argv }) => {
+            const catalogService = container.get("catalog-service") as CatalogService;
+            return await catalogService.importFile(argv.import as string);
+        },
+        help: [
+            "--import PATH",
+            "import epub or lpcl file",
```
danielweck: Must this be an **absolute** path on the filesystem? If not, is the **relative** path based on `process.cwd()`? (i.e. the command line `cwd` from where the app executable is invoked)

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,120 @@
...
+    help: string[];
+}
+
+export const cli: ICli[] = [
+    {
+        name: "_",
+        fct: async ({ argv }) => {
+            let notRead = true;
+
+            argv._.map(async (path) => {
```
danielweck: Strictly-speaking, `path` is not a reserved keyword, but it is a commonly-used NodeJS API (e.g. `path.join()`). I would therefore recommend using a different name for this local variable, for the avoidance of doubt.

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,120 @@
...
+        },
+        help: [
+            "--opds URL",
+            "import opds feed url",
+        ],
+    },
+    {
+        name: "silent",
+        fct: async (param) => {
+            param.quit = true;
```
danielweck: I am concerned about the app lifecycle when using `--silent`, because we have no control over **when** the application exits, when using to `param.quit = true`.
Instead, could we explicitly invoke `process.exit(code)` (with the appropriate exit `code`), **after** the point at which we can guarantee all application jobs have been completed (e.g. importing publications in the database and on the filesystem, opening windows, etc.)

src/utils/hostname.ts
```diff
@@ -0,0 +1,31 @@
+// ==LICENSE-BEGIN==
+// Copyright 2017 European Digital Reading Lab. All rights reserved.
+// Licensed to the Readium Foundation under one or more contributor license agreements.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END==
+
+// return Top Level Domain from URL
+export function extractHostname(url: string, tld: boolean) {
```
danielweck: I would strongly recommend using the NodeJS URL `hostname` API instead.
https://nodejs.org/api/url.html#url_url_hostname

danielweck:
For MacOS, when the application is already open, and "open with Thorium" is invoked, a special event is triggered. Should we handle it?
https://github.com/electron/electron/blob/master/docs/api/app.md#event-open-file-macos

src/main/cli/process.ts
```diff
@@ -0,0 +1,59 @@
...
+        await arg.map(async (command, idx) => {
+            const op = commandLine.find((c) => c.name === command);
+            if (op && (op.name !== "_" || argv._.length)) {
+                await op.fct(param);
+            }
+        });
+    }
+
+    if (param.quit) {
+        app.quit();
```
danielweck: Note the difference between `app.quit()` and `app.exit(code)`:
https://electronjs.org/docs/api/app#appquit
Because this is a CLI with the potential use-case of "automated publication imports" (`--silent` mode, no windows at all) I would recommend generating a proper exit code so that a script runner can pass/fail depending on the CLI termination code (for example, to chain batch imports).

src/main/cli/process.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ICli, ICliParam } from "readium-desktop/main/cli/commandLine";
+import { Arguments } from "yargs";
+
+declare const __APP_NAME__: string;
+declare const __APP_VERSION__: string;
+
+// Logger
+const debug = debug_("readium-desktop:cli");
+
+export async function processCommandLine(commandLine: ICli[], argv: Arguments): Promise<boolean> {
```
danielweck: Doing this code review, I am now realizing that this implements the layer of logic on top of the low-level `yargs` functionality. Have you considered using `meow` (or similar lib) to easily handle all the edge-cases of options parsing, help display, exit conditions / codes, and perhaps more importantly: Promise "unhandled rejections" which is a likely occurence in our case (lots of await/async without try/catch). See `hard-rejection` in https://github.com/sindresorhus/meow

src/main/cli/commandLine.ts
danielweck: `argv._.map()` ... do you mean `argv._.forEach()`?

danielweck: Note that `array.forEach(async () => { ... })` is **not safe** when using async/await Promises because the iteration will not be sequenced.
Instead, use `for (const p of argv._) { ... }` (you already have `fct: async` so there will be no linting error like `no-floating-promises` https://palantir.github.io/tslint/rules/no-floating-promises/ ).

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,120 @@
...
+                if (publication && notRead) {
+                    store.dispatch({
+                        type: readerActions.ActionType.OpenRequest,
+                        payload: {
+                            publication: {
+                                identifier: publication.identifier,
+                            },
+                        },
+                    });
+                    notRead = false;
```
danielweck: Sorry, I do not understand the role of `notRead`. Is this to ensure **only one** publication is loaded in a reader view? What about the impact of `--silent`? (will this just brutally force-quit after the reader window is opened?)

In this case I would suggest using a variable name that is more self-explanatory, for example `publicationOpenRequested=false` (then set to `true` after the `store.dispatch()` for `OpenRequest`).

panaC:
new version corrected

I will test tomorrow on windows

package.json
```diff
@@ -36,7 +36,9 @@
     "package:win": "npm run package:build && electron-builder --win --x64",
     "package:linux": "npm run package:build && electron-builder --linux",
     "package:mac": "npm run package:build && electron-builder --mac",
-    "package:all": "npm run package:build && electron-builder -mwl"
+    "package:all": "npm run package:build && electron-builder -mwl",
+    "cli:help": "npm run build && electron . -h",
```
danielweck: Why `npm run build` before? I think we should not automatically invoke the build task here (`cli:help` should be consistent with the `cli` script).

PS: if you **really** need to execute a task before launching the help command, you could create a new `pre` script in `package.json`, here for example `precli:help` (just add a `pre` prefix to the script name).

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,135 @@
...
+                store.dispatch({
+                    type: readerActions.ActionType.OpenRequest,
+                    payload: {
+                        publication: {
+                            identifier: publication[0].identifier,
+                        },
+                    },
+                });
+            } else {
+                process.stdout.write(`There is not publication title match for "${argv.read}"\n`);
```
danielweck: Sorry, my fault: `There is not` should be `There is no`!

src/main/cli/process.ts
```diff
@@ -0,0 +1,68 @@
...
+            const commandRes = await Promise.all(commandPromiseTab);
+            const ifFalse = commandRes.indexOf(false);
+            returnCode = ifFalse < 0 ? 0 : 1;
+        } catch (e) {
+            returnCode = 1;
+        }
+    }
+
+    if (param.quit) {
+        app.exit(returnCode);
```
danielweck: This is fine in `--silent` mode, because no reader windows are opened.
Otherwise, we need to use `app.quit()` instead ( https://electronjs.org/docs/api/app#appquit ) so that the app subsystems are shutdown gracefully.
For example, there is a `app.on("will-quit", ...)` event listener in the "navigator" which handles webview sessions:
https://github.com/readium/r2-navigator-js/blob/95f15b54977b309ff3e14e971980753089fab0dc/src/electron/main/sessions.ts#L265

src/main/cli/commandLine.ts
danielweck: I am closing this comment as there are other more relevant and clearer remarks elsewhere.

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,135 @@
...
+            return returnValue;
+        },
+        help: [],
+    },
+    {
+        name: "import",
+        fct: async ({ argv }) => {
+            const catalogService = container.get("catalog-service") as CatalogService;
+            await catalogService.importFile(argv.import as string);
+            return true;
```
danielweck: For MacOS, when the application is already open, and "open with Thorium" is invoked, a special event is triggered:
https://github.com/electron/electron/blob/master/docs/api/app.md#event-open-file-macos

There is already placeholder code (but no functionality attached to it):
https://github.com/readium/readium-desktop/blob/541e48bbe45407a8918018e0076d4ca5ebecd68b/src/main.ts#L254-L261

For this PR to be complete for MacOS (double-click on EPUB file, or "open with"), I think we should also invoke `catalogService.importFile()` in the event listeners. What do you think?

panaC:
ready to review

danielweck:
Hello Pierre, to see pending/unresolved review items, you can use the "speech bubble" in the right-side corner. See screenshot:
ThoriumGitHubReview

src/main/cli/process.ts
danielweck: Note that changing the current ad-hoc logic (command line parameter handling) in favour of a more robust mechanism (for example using Meow, but there are others) would require a lot of refactoring and testing. So I am going to resolve this comment for now, and will file a separate issue in due time.

danielweck:
Sorry, I meant "the issue about the absolute v. relative path"

panaC:
- we can use app.exit because no window is start : https://electronjs.org/docs/api/app#appexitexitcode
- The path must be absolute

- I test on mac, the 'open with' works
but I can't run the command line program
The .app can't be executed by the shell but only with `open Thorium reader`

src/main/cli/commandLine.ts
danielweck: @panaC said:
> The path must be absolute
https://github.com/readium/readium-desktop/pull/403#issuecomment-512746474

danielweck: Okay, so can we please have this information in the help text, for example "import epub or lpcl file (must be absolute path)"

src/main/cli/process.ts
danielweck: @panaC said:

> we can use app.exit because no window is start

https://github.com/readium/readium-desktop/pull/403#issuecomment-512746474

src/main/cli/process.ts
```diff
@@ -0,0 +1,70 @@
...
+    };
+    debug("param: ", param);
+    let returnCode = 0;
+
+    // assign arg and remove $0
+    const arg = Object.keys(argv);
+    arg.splice(arg.indexOf("$0"), 1);
+
+    // execute all commands
+    const commandPromiseTab = arg.map((command) => {
```
danielweck: Some items inside `commandPromiseTab` might be `undefined`, because not all code paths inside the iterated function return a value.
I think the intent of this code routine is to "filter" (not to return a one-to-one "map"), therefore I would suggest using `Array.filter()` with an inclusion predicate, instead of `Array.map()`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter
(this will eliminate the ambiguity for future readers of the source code)

src/main.ts
```diff
@@ -255,9 +263,26 @@ app.on("open-url", (event: any, url: any) => {
...
+                type: readerActions.ActionType.OpenRequest,
+                payload: {
+                    publication: {
+                        identifier: publication.identifier,
+                    },
+                },
+            });
+        }
+    } catch (e) {
+        process.stderr.write(`Publication error for "${filePath}"\n`);
```
danielweck: The app is already up and running here. Use `debug()`

src/main/cli/process.ts
danielweck: So `param.quit` is guaranteed to be true **only** when `--silent` is set. That's good.

Now, let's see what happens with:

1) `EXE --silent --read=TITLE` (in my case: `npm run cli -- --silent --read=Accessible` to open the existing "Accessible EPUB 3" book in the library)
2) `EXE PATH_TO_EPUB --silent` (in my case: `npm run cli -- /Volumes/EPUBs/childrens-literature.epub --silent`)

I tweaked your code so that the app is forcefully killed with `app.exit()` after a short timeout (plus a console message so I can see what happens):
```
    if (param.quit) {
        process.stderr.write(`KILL...\n`);
        setTimeout(() => {
            app.exit(returnCode);
        }, 3000);
        return true;
    }
```

In both cases 1) and 2), the app opens the publication successfully in the reader UI, and then the app is killed forcefully with `app.exit()` (no `app.quit()` clean destroy). In my test this happens after my 3s timeout, but in the current PR this occurs immediately (which is not acceptable either).

I suggest ignoring the `--silent` command line parameter when the app successfully opens its UI. Otherwise (e.g. missing TITLE, or missing PATH_TO_EPUB), then `app.exit()` with the appropriate failure code.

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,135 @@
...
+            }
+            return false;
+        },
+        help: [
+            "--opds TITLE=URL | URL",
+            "import opds feed url",
+        ],
+    },
+    {
+        name: "silent",
```
danielweck: Tests (not edge cases, just normal use cases):

1) `npm run cli -- /PATH/TO/EPUB --silent` opens the publication (reader UI) and then forcefully kills the app with `app.exit()`. Note that in this PR the forceful killing occurs immediately after the app opens the publication (too quick to see), so for my test I added a kill-timeout, to actually **see** the UI being opened.
2) `npm run cli -- --silent /PATH/TO/EPUB` kills the app, but never opens the publication => so it looks like the order of parameters matters, but this seems to be a bug, not a feature.
3) `npm run cli -- --silent=true /PATH/TO/EPUB` => same as (1)
4) `npm run cli -- --silent true /PATH/TO/EPUB` => same as (1)
5) `npm run cli -- --silent=zzzz /PATH/TO/EPUB` => same as (1)
6) `npm run cli -- --silent zzzz /PATH/TO/EPUB` => same as (1)
7) `npm run cli -- --silent=false /PATH/TO/EPUB` => same as (1). Obvious bug: `silent=false` should definitely not set `param.quit=true` and subsequently kill the app.
8) the same tests can be run with `--read=TITLE`, for example `npm run cli -- --silent=false --read=Accessibility` (successfully opens "Accessible EPU3", then forcefully kills the app)

The CLI handling code in this PR seems quite fragile. This is why I suggested using a library (like Meow) which handles CLI param types, aliases, ordering, precedence, uncaught exceptions, etc. etc.

src/main/cli/process.ts
```diff
@@ -0,0 +1,70 @@
...
+        process.stdout.write(`Usage: ${argv.$0} [options][paths...]\n`);
+        process.stdout.write("\n");
+        process.stdout.write(`Options\n`);
+        commandLine.map((v) => {
+            if (v.help.length) {
+                process.stdout.write(` ${v.help[0]}\t\t${v.help[1] ? v.help[1] : ""}\n`);
+            }
+        });
+        app.exit(0);
+        process.exit(0);
```
danielweck: `process.exit(0);` is not necessary after `app.exit(0);`

panaC: done 64c63d2

src/main/cli/commandLine.ts
panaC: Why meow is better than yargs ?
https://npmcompare.com/compare/commander,meow,minimist,yargs

yargs is already present

src/main.ts
panaC: done 99a38af

src/main/cli/process.ts
panaC: filter function returns a string Array and not a Promise<boolean>.
And the undefined is handle in Promise.all()

src/main/cli/commandLine.ts
danielweck: But, the choice of library is just an implementation detail. What matters in this PR is the design, currently flawed in terms of the state chart / flow of actions (read-UI + silent -> force-kill, etc.).

There are also bugs like `--silent=false` => `silent` is in fact true in the code! ... this is why I propose Meow which would clarify / simplify the code to ensure correct types and default value handling.

src/main/cli/process.ts
danielweck: Sure, then you chain filter().map().map() (fluent design, chained functions).
Using `.map()` to return `undefined` values in the returned array is an **anti-pattern**, the code is misleading because the assumption is that there is a one-to-one "mapping".

If the goal of the routine is to exclude items, then `.filter()` must be used.
If you then need the preserved items to be mapped to a different type / value, then chain to `.map()`

danielweck: Also, I realize that the term "tab" (for "table" in French, if I am not mistaken) is misused and confusing here. In English I would use "array" or "list" ("table" sounds like a 2D matrix columns x rows)

danielweck: Resolving, as the code has now changed significantly.

src/main/cli/commandLine.ts
danielweck: Closing this comment, as the code has now changed significantly. There is now proper Args handling of command line parameters, but we will have to test parameter types (e.g. `silent` boolean, `--silent` vs. `--silent=true` vs. `--silent 1`, etc.), default parameter values, parameter combinations and ordering (e.g. `silent` with `read`), etc.

src/main/cli/process.ts
```diff
@@ -0,0 +1,111 @@
...
+
+declare const __APP_NAME__: string;
+
+// Logger
+const debug = debug_("readium-desktop:cli");
+
+export function cli(mainFct: () => void) {
+
+    yargs
+        .scriptName(__APP_NAME__)
```
danielweck: This is great! :)

May I suggest we add `.version(__APP_VERSION__)` as well?

https://github.com/yargs/yargs/blob/master/docs/api.md#versionversionboolean

panaC: I can do that
but 
`If no arguments are passed to version (.version()), yargs will parse the package.json of your module and use its version value.` is the same that __APP_VERSION

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+    return returnValue;
+}
+
+export async function cliImport(filePath: string) {
+    try {
+        const catalogService = container.get("catalog-service") as CatalogService;
+        if (!await catalogService.importFile(filePath)) {
+            throw new Error();
```
danielweck: Is this `throw` statement meant to reject the `cliImport` promise, or is it intentional for the `catch` immediately below?
* If the former, then this does not work (the `cliImport` promise will resolve with `false` instead of the desired behaviour)
* If the latter, then I would suggest `return false` instead of needlessly throwing an exception.

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,92 @@
...
+        store.dispatch({
+            type: readerActions.ActionType.OpenRequest,
+            payload: {
+                publication: {
+                    identifier: publication[0].identifier,
+                },
+            },
+        });
+    } else {
+        process.stdout.write(`There is no publication title match for "${title}"\n`);
```
danielweck: Use `process.stderr` instead of `process.stdout` (error condition, exist status code == 1)

src/main/cli/process.ts
```diff
@@ -0,0 +1,111 @@
...
+            (y) =>
+                y.positional("title", {
+                    describe: "title of your publication",
+                    type: "string",
+                })
+            ,
+            (argv) => {
+                mainFct();
+                app.on("will-finish-launching", async () => {
+                    await cliRead(argv.title);
```
danielweck: Tell me what you think about the UX:
if the `cliRead` promise returns false, or if it is rejected (`try`/`catch` or `.catch()`), I think we should `app.quit()` gracefully (not `app.exit()`) after a few milliseconds
Rationale: the user intentionally invoked the CLI with `read <title>`, so it seems unnecessary to keep the blank browser window opened in case of failure. Instead: show the error message in the console, and abort gracefully.

src/main/cli/commandLine.ts
panaC: it's for return false when the catch appear or importFile returns null

src/main/cli/commandLine.ts
panaC: done

src/main/cli/commandLine.ts
panaC: it's for return false when the catch appear or importFile returns null

src/main/cli/commandLine.ts
panaC: done

src/main/cli/process.ts
```diff
@@ -0,0 +1,111 @@
...
+                })
+            ,
+            (argv) => {
+                // const filePathArray = argv._.map((p) => path.resolve(p));
+
+                mainFct();
+                if (argv.path) {
+                    app.on("will-finish-launching", async () => {
+                        if (!await cli_(argv.path)) {
+                            debug("no publication to open");
```
danielweck: If I am not mistaken, this debug message is incorrect. It should say something like "import failed for at least one of the given publication paths" (regardless of whether the first successfully-imported publication was opened in the reader view)

src/main/cli/process.ts
panaC: It not quit with app.exit. The application is just launched on the library and not on the reader.
I will update the code with a try-catch and debug messages

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,92 @@
...
+    let returnValue = true;
+    const filePathArray = isArray(filePath) ? filePath : [filePath];
+
+    for (const fp of filePathArray) {
+        // import and read publication
+        const catalogService = container.get("catalog-service") as CatalogService;
+        const publication = await catalogService.importFile(fp);
+        const store = container.get("store") as Store<RootState>;
+        if (publication) {
+            if (!publicationOpenRequested) {
```
danielweck: @llemeurfr @mutterback
Is there a use case for the Thorium CLI to bulk-import the given filepath parameters? (for instance, eKitabu catalog provisioning?)
For example: `thorium.exe PATH1 PATH2 PATH3 ... PATHXX` or `thorium.exe FOLDER`

If so, this would seem like the right place to take into account a `silent` boolean flag, which would skip the step "open the first successfully-imported publication into a reader view" (note that with the current PR code, subsequent successfully-imported publications are not opened)

Also, the combination of `paths` parameter + the `silent` flag would negate the need for the special `import` command (which is the equivalent of `silent` === `true`).

src/main/cli/process.ts
panaC: I will say "import and read failed for at least one of the given publication paths"

src/main/cli/process.ts
danielweck: `__APP_VERSION__` is a build-time "preprocessor directive" global.

Yargs can fetch `package.json` at runtime if it knows where it is, but its default location may not work in all contexts (ASAR, `__dirname` inside the app JS bundle, etc.).

So, better provide the version reliably, via the existing `__APP_VERSION__` build-time global.

src/main/cli/process.ts
```diff
@@ -80,7 +80,13 @@ export function cli(mainFct: () => void) {
             (argv) => {
                 mainFct();
                 app.on("will-finish-launching", async () => {
-                    await cliRead(argv.title);
+                    try {
+                        if (!await cliRead(argv.title)) {
+                            debug("no publication to read");
+                        }
+                    } catch (e) {
+                        debug("An error was appears on cliRead");
```
danielweck: > An error was appears on ...

Correction: "An error occurred during ..."

src/main/cli/process.ts
```diff
@@ -99,8 +105,12 @@ export function cli(mainFct: () => void) {
...
                 if (argv.path) {
                     app.on("will-finish-launching", async () => {
-                        if (!await cli_(argv.path)) {
-                            debug("no publication to open");
+                        try {
+                            if (!await cli_(argv.path)) {
+                                debug("import failed for at least one of the given publication paths");
+                            }
+                        } catch (e) {
+                            debug("An error was appears on default CLI");
```
danielweck: > An error was appears on ...

Correction: "An error occurred during ..."

src/main/cli/commandLine.ts
panaC: In no case "thorium.exe PATH1 PATH2 PATH3 ... PATHXX" is the import command

I will just add an option to add multiple publication in a same time and/or a glob path.

It will be `./Thorium import PATH1 PATH2 ...``
or/and `./Thorium 'myPath/**/*.epub' to import all the folder


The command ./Thorium PATH1 is mainly for "open with" on windows/linux/mac

src/main/cli/process.ts
danielweck: So just to clarify: your personal opinion about the UX is that: if the user invokes `thorium.exe TITLE` from the console (or from an OS shortcut), and the TITLE does not match any publication => the application should remain open (library view)?

If so, I just want to clarify that I personally feel the app should shut down gracefully, with `process.stderr.print` and either `app.exit(code)` (abort), or `app.quit()` (clean exit, just in case some destroy/teardown functionality depends on the Electron `quit` events)

What do you think?

src/main/cli/process.ts
panaC: done

src/main/cli/process.ts
panaC: done

src/main/cli/commandLine.ts
llemeurfr: I'm ok for adding a bulk import, as e.g. `./Thorium import 'myPath/**/*.epub'`.
It will allow for instance the import of every IPDF samples in one command. 

I don't see a use case for importing the content of multiple folders.

danielweck: So right now, if a user invokes the command line `thorium.exe PATH1 PATH2 PATH3 ... PATHXX` (assuming all paths exist) =>
1) all paths are imported
2) PATH1 is open in the reader view

So why have a separate `import` CLI function? This seems redundant. Why not provide a `silent` boolean flag (default `false`) so that users can bulk-import a list of publications without triggering the "open reader for first successful import"?

danielweck: PS: +1 to Glob support.
https://github.com/sindresorhus/globby
or
https://github.com/isaacs/node-glob

src/main/cli/commandLine.ts
danielweck: In that case `return false`. Throwing an exception is (1) unnecessary, and more importantly (2) misleading (because it implies that the intent is to reject the promise, when in fact it is to resolve it with `false`).

src/main/cli/commandLine.ts
panaC: The Idea behind défault command is to have a fallback. Normally on défault mode : it have to just read and not import to library .. but we can't doing this yet

danielweck: Yes this would be great: temporarily open/read a publication without importing into the library ... but I think this would require a pretty significant refactoring for LCP support, reading location/bookmarks/annotations etc. (anything that requires the reader state to be managed via the DB)

llemeurfr: ... or simply delete the publication when the app is closed, maybe.

danielweck: Yes, I think this is the most pragmatic / realistic option. This would probably cause UX confusion though (e.g. publication in bookshelf whilst app is open, then disappears at next restart) ... unless we have a special flag which hides the "transient" EPUBs from the library.

panaC: You can remove publication after reading, but that implies to track the closing or 'goto library' events from reader window.

The issue is more deeper, how implement both duplicate or replace the publication ? we need the publication identifier. When the user import several ebooks they actually duplicated.

danielweck: Yes, see Laurent's suggestions here: https://github.com/readium/readium-desktop/issues/438

src/main/cli/process.ts
panaC: an error message in react can be displayed to the user .. but for me, in any case, it must leave.
When a user double-clicks on an epub, he will not see thorium at all.

the launch principle of thorium is the opening of the library then the reader.

danielweck: > Otherwise, when a user double-clicks on an epub, he will not see thorium at all.

Right, that makes sense. So for example `thorium.exe TITLE` could be used via an icon shortcut on the user's desktop, it is not necessarily a command line / shell invocation. In which case, the user must be able to get some GUI feedback, instead of just `stderr`. What about Electron's own built-in `dialog.showErrorBox()`?
https://electronjs.org/docs/api/dialog#dialogshowerrorboxtitle-content

panaC: done

src/main/cli/process.ts
panaC: done

src/main/cli/commandLine.ts
panaC: done

src/main/cli/process.ts
```diff
@@ -0,0 +1,133 @@
...
+                if (argv.path) {
+                    app.on("ready", async () => {
+                        try {
+                            if (!await cli_(argv.path)) {
+                                const errorTitle = "Import Failed";
+                                const errorMessage = `Import failed for the publication path : ${argv.path}`;
+                                debug(errorMessage);
+                                dialog.showErrorBox(errorTitle, errorMessage);
+                            }
+                        } catch (e) {
```
danielweck: Suggestion: `process.stderr.write()` and `dialog.showErrorBox()` of the caught exception `e`.

This exception can (for example) be the *uncaught* promise rejection from `const publication = await catalogService.importFile(filePath);` which is called inside the `cli_()` function. So it is important to *not* swallow the context of the error, and to report it appropriately. Do you think `stderr` is necessary as well as `dialog.showErrorBox()`?

src/main/cli/process.ts
```diff
@@ -0,0 +1,133 @@
...
+                    realpath: true,
+                }) || [];
+                const promise = cliImport(pathArray);
+                promise.then((isValid) => {
+                    if (isValid) {
+                        app.exit(0);
+                        return;
+                    }
+                    app.exit(1);
+                }).catch(() => {
```
danielweck: Suggestion: `process.stderr.write()`of the caught promise rejection (need to pass the `err` param in the `.catch()` callback).

This caught error can (for example) be the *uncaught* promise rejection from `if (!await catalogService.importFile(fp))` which is called inside the `cliImport()` function. This is why I suggest reporting the full error context to the CLI user.

src/main/cli/process.ts
```diff
@@ -0,0 +1,133 @@
...
+                    describe: "title of your publication",
+                    type: "string",
+                })
+            ,
+            (argv) => {
+                mainFct();
+                app.on("ready", async () => {
+                    try {
+                        if (!await cliRead(argv.title)) {
+                            debug("no publication to read");
```
danielweck: Should there be a `dialog.showErrorBox()` here as well?

src/main/cli/process.ts
```diff
@@ -0,0 +1,133 @@
...
+                        }
+                    } catch (e) {
```
danielweck: Same remark as other comments about reporting the exception / promise rejection in `stderr` and `dialog.showErrorBox()`.

danielweck: PS: the potential uncaught promise rejection in this case is `const publication = await publicationRepo.searchByTitle(title);` called from within the `cliRead()` function.

src/main.ts
```diff
@@ -51,403 +34,82 @@ if (_PACKAGING !== "0") {
 // Logger
 const debug = debug_("readium-desktop:main");
 
-// Parse command line
-const processArgs = yargs.argv;
-
+// Global
```
danielweck: In this PR, `main.ts` gets divided into several constituent parts, which I think is a good idea (makes the code clearer to read and maintain). 

However, we now have to be **very** careful before merging, because there have been changes to `main.ts` pushed to `develop` since this PR was started:
https://github.com/readium/readium-desktop/commits/develop/src/main.ts

For example, the `console` disabling:
https://github.com/readium/readium-desktop/commit/c2296e761b40da0b779bb458b033a3df68144e14#diff-8cfead41d88ad47d44509a8ab0a109ad
...should show-up "automatically" just above this PR comment (when looking at the code diff), once `develop` is merged back into this PR's branch.

BUT, potential changes made to other parts of `main.ts` that are now separated into `createWindow.ts` or `init.ts` will not automatically be "ported" there. For example, the `i18next` locale sync is now in `init.ts`, but it is behind the tip of the `develop` branch:
https://github.com/readium/readium-desktop/commit/ffe98cb490bbfe89ff0200951dd07112deaa93fc#diff-8cfead41d88ad47d44509a8ab0a109ad
(search for the string `const lang = Object.keys(AvailableLanguages).find((l) => l === loc) || "en";` in this PR code diff)

Having analysed the Git commit history of `main.ts`, I think there are no other instances of breaking changes. @panaC what do you think?

webpack.config-preprocessor-directives.js
```diff
@@ -29,6 +29,7 @@ const nodeModuleRelativeUrl = (isPackaging === "1") ?
 
 const data = {
     __APP_VERSION__: JSON.stringify(version),
+    __APP_NAME__: JSON.stringify(build.productName),
```
danielweck: Please update the PR's branch to sync with `upstream/develop`, see: https://github.com/readium/readium-desktop/commit/de8e69b2513c67adacf5b4b3bf3dcdf8cb2132ee

panaC: done

package.json
```diff
@@ -156,6 +171,10 @@
     "connected-react-router": "^6.4.0",
     "debug": "^4.0.1",
     "font-awesome": "^4.7.0",
+    "glob": "^7.1.4",
+    "html-loader": "^0.5.5",
+    "html-webpack-plugin": "^3.2.0",
+    "i": "^0.3.6",
```
danielweck: Why are `html-webpack-plugin`, `html-loader`, and `i` added in this PR? (sorry if I am missing something obvious ... I can see that `glob` is introduced by this PR, but I do not see the reason for the other NPM packages)

danielweck: Mhmm, some of these packages already exist in the `devDependencies`:
https://github.com/readium/readium-desktop/blob/bb36f5b545e351654ed07fef0f55fde635de5c01/package.json#L234-L236
...but in this PR they are added to `dependencies`, there is something wrong here!
(and what about `i`?)

danielweck: Ah, I now understand. The recent sync you did in this PR's branch with the develop branch wasn't quite correct. These 4 packages must be removed from `dependencies` (in this PR). See: https://github.com/readium/readium-desktop/pull/504

panaC: done

src/main/cli/process.ts
panaC: done

src/main/cli/process.ts
panaC: done

src/main/cli/process.ts
panaC: done

src/main/cli/process.ts
panaC: done

package.json
```diff
@@ -156,6 +171,7 @@
     "connected-react-router": "^6.4.0",
     "debug": "^4.0.1",
     "font-awesome": "^4.7.0",
+    "glob": "^7.1.4",
```
danielweck: In your PR, can you please remove `"glob": "^7.1.4"` from `devDependencies`, otherwise this PR introduces a duplicate package in `dependencies`.

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,82 @@
...
+        const catalogService = container.get("catalog-service") as CatalogService;
+        if (!await catalogService.importFile(fp)) {
+            returnValue = false;
+        }
+    }
+    return returnValue;
+}
+
+export async function cliOpds(title: string, url: string) {
+    // extract and save the title and url from opdsFeed
```
danielweck: Is that a todo? If so, can we please have a `TODO:` prefix in the comment.

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,82 @@
...
+    // title=http://myurl.com or get TLD and set url
```
danielweck: What is `TLD` please?

src/main/cli/commandLine.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (hostname) {
+        const opdsRepository = container.get("opds-feed-repository") as OpdsFeedRepository;
+        await opdsRepository.save({ title, url });
+        return true;
+    }
+    return false;
+}
+
+export async function cliRead(title: string) {
+    // read the publication name
```
danielweck: I do not understand this comment. Is this a todo?

src/main/cli/process.ts
```diff
@@ -0,0 +1,142 @@
...
+                    describe: "url opds feed",
+                    type: "string",
+                })
+            ,
+            (argv) => {
+                const promise = cliOpds(argv.title, argv.url);
+                promise.then((isValid) => {
+                    if (isValid) {
+                        app.exit(0);
+                        return;
```
danielweck: Just a thought (tell me what you think): as a CLI user, if the operation succeeds I get a blank response (no information in the console). Should we output a shell message like `process.stdout.write("OPDS import done." + EOL);`?

src/main/cli/process.ts
```diff
@@ -0,0 +1,142 @@
...
+            (argv) => {
+                const pathArray = glob.sync(argv.path, {
+                    absolute: true,
+                    realpath: true,
+                }) || [];
+                const promise = cliImport(pathArray);
+                promise.then((isValid) => {
+                    if (isValid) {
+                        app.exit(0);
+                        return;
```
danielweck: Same remark as for the OPDS CLI: as a CLI user, if the operation succeeds I get a blank response (no information in the console). Should we output a shell message like `process.stdout.write("Publication(s) import done." + EOL);`?

src/main/cli/process.ts
```diff
@@ -0,0 +1,142 @@
...
+                    }
+                    process.stderr.write("OPDS URL not valid, exit with code 1" + EOL);
+                    app.exit(1);
+                }).catch((e) => {
+                    debug("import error :", e);
+                    process.stderr.write("An error occurred during import an ODPS feed, exit with code 1" + EOL);
```
danielweck: I think we need to add `process.stderr.write(e.toString() + EOL);`, because in production+packaged mode `debug()` is disabled.

src/main/cli/process.ts
```diff
@@ -0,0 +1,142 @@
...
+                promise.then((isValid) => {
+                    if (isValid) {
+                        app.exit(0);
+                        return;
+                    }
+                    process.stderr.write("No valid files, exit with code 1" + EOL);
+                    app.exit(1);
+                }).catch((e) => {
+                    debug("import error :", e);
+                    process.stderr.write("An error occurred during import, exit with code 1" + EOL);
```
danielweck: I think we need to add `process.stderr.write(e.toString() + EOL);`, because in production+packaged mode `debug()` is disabled.

panaC:
ready to merge

danielweck:
Hotfix: https://github.com/readium/readium-desktop/commit/376468474b056785fb620aea12db5cd836853d64

danielweck:
Hotfix: https://github.com/readium/readium-desktop/commit/8f9141943efbf654bc18e4a272edfcc2fe395c47

PR #404, opened 2019-07-10
stricter code checks (TypeScript and Lint)

danielweck:
This PR removes a few `any` types, removes unused variables, etc. (see linting rules)

Note that strict `null`/`undefined` checks are implemented in a separate PR: https://github.com/readium/readium-desktop/pull/475


danielweck:
Related information: https://code.visualstudio.com/blogs/2019/05/23/strict-null

danielweck:
Related information: https://code.visualstudio.com/blogs/2019/05/23/strict-null

danielweck:
This PR is reasonably consensual, I am hoping we will reach an agreement to merge it.
I created a separate PR for the much less trivial code changes required to achieve `strictNullChecks`:
https://github.com/readium/readium-desktop/pull/475

danielweck:
Note that TypeScript 3.6 will provide strongly-typed iterators and generators, which means that the returned values of Redux-Saga `yield call()` will not have to be typed explicitly. https://github.com/microsoft/TypeScript/issues/31639

danielweck:
Hello Pierre @panaC if you have time, I would appreciate your input / review on this PR. The code alterations are mostly lint-related, there are no changes in execution logic. I think that stricter type checking and linting is needed in the Thorium codebase, so that we can continue implementing improvements / refactoring more efficiently.
Thank you! :)

Note that the other PR https://github.com/readium/readium-desktop/pull/475
is much more tricky to integrate (in fact, at this stage, impossible), because of the added strictness for nullable types which breaks compilation. We can ignore this PR for now.

PR #405, opened 2019-07-10
[fix] add icon in reader window

panaC:
- add an icon line in new window object

### DIFF
--- a/src/main/redux/sagas/reader.ts
+++ b/src/main/redux/sagas/reader.ts
@@ -66,6 +66,7 @@ async function openReader(publication: Publication, manifestUrl: string) {
             webSecurity: true,
             webviewTag: true,
         },
+        icon: path.join(__dirname, "assets/icons/icon.png"),
     });
     const winRegistry = container.get("win-registry") as WinRegistry;
     const appWindows = winRegistry.getWindows();


src/main/redux/sagas/reader.ts
```diff
@@ -66,6 +66,7 @@ async function openReader(publication: Publication, manifestUrl: string) {
             webSecurity: true,
             webviewTag: true,
         },
+        icon: path.join(__dirname, "assets/icons/icon.png"),
```
danielweck: This mirrors the `main.ts` library window, right?
https://github.com/readium/readium-desktop/blob/209b4513844848e1d19c606a1d186d85eb89741b/src/main.ts#L92

Just to make double-check: are we 100% sure that `__dirname` works in all build modes?
* development mode `npm run start:dev`
* production mode `npm run start`
* production + packaged mode `npm run package:mac|win|linux`

I think the answer to my question is "yes, it's fine", mainly because `__dirname` works in other cases where the resource may (or may not) be included in ASAR at production+packaging time, for example the LCP lib:
https://github.com/readium/readium-desktop/blob/209b4513844848e1d19c606a1d186d85eb89741b/src/main.ts#L61-L62

...but I am asking because we also have a precedent of dealing with relative / absolute paths when including ReadiumCSS:
https://github.com/readium/readium-desktop/blob/209b4513844848e1d19c606a1d186d85eb89741b/src/main/streamer.ts#L52-L58

...and worse even, the "preload" script path that is used to configure the navigator:
https://github.com/readium/readium-desktop/blob/209b4513844848e1d19c606a1d186d85eb89741b/src/renderer/components/reader/Reader.tsx#L493-L509

(basically, it also depends on the API that consumes the generated absolute path for the resource)

Note that there is also the special case of WebPack at development time, but this doesn't affect the logic of the window icon (I just wanted to mention this special case). See:
https://github.com/readium/readium-desktop/blob/209b4513844848e1d19c606a1d186d85eb89741b/src/main.ts#L99-L105
https://github.com/readium/readium-desktop/blob/209b4513844848e1d19c606a1d186d85eb89741b/src/main/redux/sagas/reader.ts#L107-L113

So, bottom line: my default assumption is that the icon will work just fine in all build modes, I have a high degree of confidence that there will not be any issues. I would still like to test this assumption though :)

PR #416, opened 2019-07-11
[fix] replace console.log by debug or console.error

panaC:
This PR fix the print of IPC API historic on stdio

I removed also other console.log.

I don't removed console.log in JSX

Tested on MAC-OS

### DIFF
--- a/src/main/streamer.ts
+++ b/src/main/streamer.ts
@@ -126,8 +126,7 @@ function computeReadiumCssJsonMessage(publication: Publication, link: Link | und
         reduceMotion: readiumCSSDefaults.reduceMotion,
     };
     const jsonMsg: IEventPayload_R2_EVENT_READIUMCSS = { setCSS: cssJson };
-    console.log("jsonMsg MAIN");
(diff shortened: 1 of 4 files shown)


src/main/streamer.ts
```diff
@@ -126,8 +126,7 @@ function computeReadiumCssJsonMessage(publication: Publication, link: Link | und
         reduceMotion: readiumCSSDefaults.reduceMotion,
     };
     const jsonMsg: IEventPayload_R2_EVENT_READIUMCSS = { setCSS: cssJson };
-    console.log("jsonMsg MAIN");
-    console.log(jsonMsg);
+    debug("jsonMsg MAIN", jsonMsg);
```
danielweck: Can we please remove this? (it was me who introduced this logging code)

Note that there is no console logging in the equivalent reader logic:
https://github.com/readium/readium-desktop/blob/86a83b7815970200f69ad3788582565752eca230/src/renderer/components/reader/Reader.tsx#L157-L158

PR #423, opened 2019-07-12
[fix #422] information in settings

### LINKED ISSUE #422: Information tab doesn't show description of Thorium in production+packaged mode
Information tab show no Thorium app description in any language.

information

panaC:
I import info file through webpack and not from a local src path

fix #422


src/renderer/components/settings/Information.tsx
```diff
@@ -5,25 +5,29 @@
...
+import * as info_fr from "readium-desktop/resources/information/fr.md";
+import Header from "./Header";
 
-import { readFile } from "fs";
+type Record<K extends keyof any, T> = {
+    [P in K]: T;
+};
 
-import { promisify } from "util";
+const info: Record<string, typeof import("*.md")> = {
```
danielweck: I think that defining a generic type just for the `lang` -> `info` mapping (which is effectively just a plain Javascript "indexed" object with string keys) introduces unnecessary complexity in the TypeScript source code.
Case in point, this line of code further down:
`info[locale] as unknown as string`

What about a simple type (name just for illustration purposes):
`interface StringToStringMap { [key: string]: string; }`

src/renderer/components/settings/Information.tsx
```diff
@@ -5,25 +5,29 @@
...
-import * as commonmark from "commonmark";
-
 import * as packageJson from "readium-desktop/package.json";
+import LibraryLayout from "readium-desktop/renderer/components/layout/LibraryLayout";
+import {
+    TranslatorProps, withTranslator,
+} from "readium-desktop/renderer/components/utils/translator";
+import * as info_de from "readium-desktop/resources/information/de.md";
+import * as info_en from "readium-desktop/resources/information/en.md";
+import * as info_fr from "readium-desktop/resources/information/fr.md";
```
danielweck: I think that using the compile-time WebPack loader for every single Markdown info file for each supported language is not a scalable solution. First of all, this source code will need to be updated when a locale is added. Secondly, the markdown files occupy memory unnecessarily.

Instead, I would suggest adding a compile-time WebPack directive to copy the info files into the app dist folder:
https://github.com/readium/readium-desktop/blob/6cf1e1b1f55b39a9a25aa77aac734e2418b8e3a0/webpack.config.main.js#L112-L129

Then, I would continue to use the following line of code:
```
let fileContent = await promisify(readFile)(`src/resources/information/${locale}.md`, {encoding: "utf8"});
```

...but I would use `__dirname` and `path.join()` to correctly reference `${locale}.md` in a cross-platform manner.

panaC: __dirname returns '/' and not the same main process path

i'm looking for a solution but I don't see it

panaC: https://medium.com/@dkerrious/electron-packager-file-path-issue-24f4bcbfc970

src/renderer/components/settings/Information.tsx
panaC: you're right, done

src/renderer/components/settings/Information.tsx
danielweck: For the renderer process, you can follow existing usage of `__dirname`:
https://github.com/readium/readium-desktop/blob/b2314a3947e534387183cf6bf37503d3874f5ce2/src/renderer/components/reader/Reader.tsx#L493-L509

danielweck: Note that in the particular case of "Thorium info Markdown files", you will not need `_NODE_MODULE_RELATIVE_URL`. But just for your information, this is used in the main process as well, so use the correct `node_module` location:

https://github.com/readium/readium-desktop/blob/541e48bbe45407a8918018e0076d4ca5ebecd68b/src/main/streamer.ts#L52-L58

src/renderer/components/settings/Information.tsx
```diff
@@ -47,15 +47,13 @@ export class LanguageSettings extends React.Component<Props, States> {
 
     public async componentDidMount() {
         const { locale } = this.props;
-
-        try {
-            let fileContent = await promisify(readFile)(`src/resources/information/${locale}.md`, {encoding: "utf8"});
+        if (info[locale]) {
+            let fileContent = info[locale] as unknown as string;
```
danielweck: Now that `info` is declared as `const info: { [key: string]: typeof import("*.md") }`, can this line become `let fileContent = info[locale]`? (why do we need explicit type coercion `as unknown as string`?) If not, can we just replace `typeof import("*.md")` with `string`?

src/renderer/components/settings/Information.tsx
```diff
@@ -47,9 +44,12 @@ export class LanguageSettings extends React.Component<Props, States> {
 
     public async componentDidMount() {
         const { locale } = this.props;
-
+        let filePath: string = path.join((global as any).__dirname, __INFO_MD_RELATIVE_URL__);
```
danielweck: Minor detail: rename `filePath` to `folderPath` (for the avoidance of doubt when reading the source code)

webpack.config-preprocessor-directives.js
```diff
@@ -27,13 +27,17 @@ const isPackaging = process.env.PACKAGING || "0";
 const nodeModuleRelativeUrl = (isPackaging === "1") ?
     "node_modules" : "../node_modules";
 
+const infoMdRelativeUrl = (isPackaging === '1') ?
+    "assets/md/information" : "dist/assets/md/information"
+
```
danielweck: After reviewing the current usage of build-time variables defined in `webpack.config-preprocessor-directives.js`, as well as other cases in the source code where we need to reference known (relative) paths, I feel that we should leave `webpack.config-preprocessor-directives.js` unchanged, and simply add the `[dist/]assets/md/information` logic in `src/renderer/components/settings/Information.tsx`, where we already test the `__PACKAGING__` condition.

panaC: Je comprend, mais actuellement webpack fonctionne,
Cyrille, m'a expliqué de faire comme ça, hier,

Je ne vais pas faire et défaire ce que je fais à chaque fois.

Nous avons des deadline à respecter.

La PR est testée sur MAC/PC et elle fonctionne.

Peux-tu la merger quand même a moins qu'il y ai qqe chose de vraiment bloquant.

Merci bcp Daniel.

danielweck: English please on GitHub :)

danielweck: * As a peer code reviewer I have a duty to highlight potential issues (e.g. regression bugs, architectural inconsistencies, etc.). Although the proposed solution "works", I observed that the addition of the `__INFO_MD_RELATIVE_URL__` "preprocessor directives" is out of place, from an architectural perspective. The proposed design is inconsistent with other usage in the codebase, because `__INFO_MD_RELATIVE_URL__` is domain-specific (i.e. particular to the logic of the Markdown "information file"), as opposed to; let's say; `__NODE_MODULE_RELATIVE_URL__` which is applicable to other use cases.
* As the developer currently responsible for approving + merging Pull Requests, I will follow the advice of other team members, and if the advice is: "this code works, we are under pressure because of a deliverable deadline, please disregard design issues, and merge as-is", then I will do so. However I stand by the fact that sloppy design decisions (even small ones) are a bad signal for current and future project contributors.

In cases where the problem is too complex to fix in time for a milestone release, sure, we can add a "TODO" comment in the source code, and file a separate actionable issue so that we can come back to it at a later stage.
In this particular case, it would take as much time to add a TODO and file an issue, as it would to fix the design problem in the first place.

llemeurfr: Re. the process, @danielweck is right: as reviewer, his task is to avoid evolutions which lead to bad design. So @panaC has to follow his advice.
Re. this PR, the initial info associated with the proposed modification is weak, so I can't take any position on its content.
Each PR must strike a balance between advantages, review time and deadlines.
Our focus in this sprint is accessibility, therefore our time should be spent on tasks related to accessibilty

panaC: Ok, you're all right, it's important to have a clean code base.
We are a team, you are the project manager: I prefer in this case that you give me a briefing before coding the solution of each issue

I will make the change.

danielweck: Thank you both for your feedback. Onwards and upwards! :)

src/renderer/components/settings/Information.tsx
```diff
@@ -5,25 +5,22 @@
...
-
+import LibraryLayout from "readium-desktop/renderer/components/layout/LibraryLayout";
+import {
+    TranslatorProps, withTranslator,
+} from "readium-desktop/renderer/components/utils/translator";
 import { promisify } from "util";
+import Header from "./Header";
+
+declare const __PACKAGING__: string;
+declare const __INFO_MD_RELATIVE_URL__: string;
```
danielweck: Unused constant.

src/renderer/components/settings/Information.tsx
```diff
@@ -47,9 +44,12 @@ export class LanguageSettings extends React.Component<Props, States> {
 
     public async componentDidMount() {
         const { locale } = this.props;
-
+        let folderPath: string = path.join((global as any).__dirname, "assets/md/information");
         try {
-            let fileContent = await promisify(readFile)(`src/resources/information/${locale}.md`, {encoding: "utf8"});
+            if (__PACKAGING__ === "0") {
+                folderPath = path.join(process.cwd(), "dist/assets/md/information");
```
danielweck: The string `assets/md/information` is duplicated, instead it should be defined once, for example:

```javascript
const infoFolderRelativePath = "assets/md/information";
// ...
let folderPath: string = path.join((global as any).__dirname, infoFolderRelativePath);
// ...
folderPath = path.join(process.cwd(), "dist", infoFolderRelativePath);
```

Other than that, this now looks clean to merge :)

panaC: Done

src/renderer/components/settings/Information.tsx
```diff
@@ -44,10 +44,12 @@ export class LanguageSettings extends React.Component<Props, States> {
...
     public async componentDidMount() {
         const { locale } = this.props;
-        let folderPath: string = path.join((global as any).__dirname, "assets/md/information");
+        const infoFolderRelativePath = "assets/md/information";
+
+        let folderPath: string = path.join((global as any).__dirname, infoFolderRelativePath);
         try {
             if (__PACKAGING__ === "0") {
-                folderPath = path.join(process.cwd(), "dist/assets/md/information");
+                folderPath = path.join(process.cwd(), infoFolderRelativePath);
```
danielweck: Shouldn't this be
```
`dist/${infoFolderRelativePath}`
```
?

danielweck: Or better even:
`folderPath = path.join(process.cwd(), "dist", "infoFolderRelativePath");`

danielweck: Please use the multi-param `path.join()` variant, as this will take care of forward/reverse slash path separator.

panaC: done

src/renderer/components/settings/Information.tsx
panaC: done

src/renderer/components/settings/Information.tsx
panaC: done

PR #428, opened 2019-07-15
Make book cover focusable & clickable

### LINKED ISSUE #424: Make book cover focusable and clickable
"read book" button is missing in drop-down book menu. we should add button to allow blind person to 
access at book more easier.
Currently they must open a modal window by "about this book" button in drop-down menu then open a book by "Read" button, it take too much time.

manuellagrand:
This PR make book cover focusable and clickable for ease blind person users to access books .

fixes #424


src/renderer/components/publication/PublicationCard.tsx
```diff
@@ -65,13 +65,17 @@ class PublicationCard extends React.Component<PublicationCardProps, PublicationC
...
             <div className={styles.block_book}
                 aria-haspopup="dialog"
                 aria-controls="dialog"
             >
                 <div className={styles.image_wrapper}>
-                    <a aria-hidden onClick={(e) => this.handleBookClick(e)}>
+                    <a
+                    tabIndex={0}
+                    onClick={(e) => this.handleBookClick(e)}
+                    onKeyPress={(e) => this.handleBookClick(e)}>
```
danielweck: We need to filter keyboard key-press events with `if (e.charCode === 13) { ... }` (`enter` key)

src/renderer/components/publication/PublicationCard.tsx
```diff
@@ -123,6 +124,19 @@ class PublicationCard extends React.Component<PublicationCardProps, PublicationC
...
+
+        if (e.key === "Enter") {
+            if (this.props.isOpds || !lcpReadable(publication, lsdStatus)) {
+                this.props.openInfosDialog(publication);
+            } else {
+                this.props.openReader(publication);
+            }
+        }
+    }
+
```
danielweck: Sorry for nit-picking, but this function duplicates pretty much the entire `handleBookClick()` body.

Can we instead filter the event type upstream of the function call? For example instead of:
`onKeyPress={(e) => this.handleEnterPressed(e)}`
...simply use:
`onKeyPress={(e) => { if (e.charCode === 13) { this.handleBookClick(e); } }}`

Note that I am suggesting using `e.charCode === 13` instead of `e.key === "Enter"`, in order to be consistent with other cases in the codebase.

PS: maybe replace `any` with `React.SyntheticEvent` in `handleBookClick(e: any)`? (this is just from the top of my head ... I haven't actually run the TypeScript compiler on this :)

PR #429, opened 2019-07-15
Fix/give focus back

### LINKED ISSUE #425: After pressing escape key to close the information modal window, outline is not visible on currently focused button
outline is not visible after pressing escape key to close the information modal window.

focus should come back to the information button.

currently the focus is lost.

manuellagrand:
This PR fix the focus lost when the modal window is closed.

fix #425


src/renderer/components/reader/Reader.tsx
```diff
@@ -717,6 +719,7 @@ const mapStateToProps = (state: RootState, __: any) => {
     return {
         reader: state.reader.reader,
         mode: state.reader.mode,
+        infoOpen: state.dialog.open,
```
danielweck: If I am not mistaken, `state.dialog.open` is from the global Redux `store.getState()`, which means it will be true for any open dialog, not just the "info" dialog.
So, here is a suggestion (please review): we need a new entry like `state.reader.infoDialog`, which we can set to true in `displayPublicationInfo` of `mapDispatchToProps` (where we already `setState` on `shortcutEnable`). Then, we need to reset `state.reader.infoDialog` to false in the event listener for `Thorium:DialogClose` (where we already also reset `setState` on `shortcutEnable`). What do you think?

src/renderer/components/reader/ReaderHeader.tsx
```diff
@@ -75,6 +77,11 @@ export class ReaderHeader extends React.Component<Props, undefined> {
                 this.enableFullscreenRef.current.focus();
             }
         }
+
+        if (this.props.infoOpen !== oldProps.infoOpen &&
+            this.props.infoOpen === false) {
+                this.infoMenuButtonRef.current.focus();
```
danielweck: I agree with this logic, but I just want to point out that when we will implement keyboard shortcuts so that users can quickly invoke main features - like the book info panel, for example with `CTRL i` - then the keyboard focus will not necessarily have been on the info button (the popup dialog was triggered by a combination of keystrokes whilst the keyboard focus was somewhere else in the UI).
Anyway, this is a global problem we will have to solve if we decide to *not* steal the focus when invoking keyboard shortcuts.
Right now, let's keep it simple.

src/renderer/components/reader/Reader.tsx
danielweck: Ah, even simpler:
`infoOpen: state.dialog.open && state.dialog.type === DialogType.PublicationInfoReader`

As you can see, you can test for the specific type of dialog currently opened in the browser window, which is the current local context for the global Redux `store.getState()` (therefore `state.dialog.open` is already specific to the partiular instance of the reader view)

Issue #435, opened 2019-07-16
Language preference not saved when app is restarted


kevinmainairungu:
The language preference is not saved once app is restarted.

Steps to Reproduce:
1.	Launch the application.
2.	Select ‘Settings”.
3.	Select “French”.
4.	Close the application.
5.	Launch the application

danielweck:
Thank you for reporting.
I thought persistence was already implemented, as part of this PR:
https://github.com/readium/readium-desktop/commit/024dc4e64c645dc150432a232e0b65912dcb10d6

It looks like the locale configured in the store is not persistent / saved into DB. To be investigated.

danielweck:
`store.dispatch(setLocale(lang));`
=>
We seem to already have everything in place in the Redux actions:
https://github.com/readium/readium-desktop/blob/develop/src/common/redux/actions/i18n.ts

(search for all occurrences of `ActionType.Set` to find the code paths)

danielweck:
I think there is a typo in the Redux reducer:
https://github.com/readium/readium-desktop/blob/877e2fee1280535bbc14e03094298a80ab998fd0/src/common/redux/reducers/i18n.ts#L23-L26
(missing `i18n` key, should be `state.i18n.locale`)

kevinmainairungu:
@danielweck thank you for the update.

kevinmainairungu:
@danielweck kindly when are you going to make changes to this? Merge fix.

danielweck:
Work in Progress...
https://github.com/readium/readium-desktop/pull/436/files

kevinmainairungu:
@danielweck thanks

PR #436, opened 2019-07-16
WIP (DO NOT MERGE): persisting language locale

danielweck:
first stab at persisting language locale (Redux Saga boilerplate).

Fixes https://github.com/readium/readium-desktop/issues/435

### DIFF
--- a/src/renderer/components/settings/Information.tsx
+++ b/src/renderer/components/settings/Information.tsx
@@ -34,7 +34,7 @@ interface States {
     placeholder: any;
 }
 
-export class LanguageSettings extends React.Component<Props, States> {
+export class AppInformation extends React.Component<Props, States> {
     private parsedMarkdown: string;
 
     public constructor(props: Props) {
@@ -87,4 +87,4 @@ const mapDispatchToProps = (dispatch: any) => {
     };
 };
 
-export default connect(mapStateToProps, mapDispatchToProps)(withTranslator(LanguageSettings));
+export default connect(mapStateToProps, mapDispatchToProps)(withTranslator(AppInformation));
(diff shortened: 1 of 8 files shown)


danielweck:
@Scarsniik @clebeaupin Could you please help with this one?

Please take a look at the code diff in this PR. As you can see I tried to use the `ConfigRepository` to save the `state.i18n.locale` into persistent storage (alongside the reader settings, etc.).

My understanding of the Redux / Saga boilerplate necessary to make this work is limited (unfortunately my PR does not successfully save the locale in the DB document).

I would appreciate if you could educate me (and the other devs on the team) about the Redux Saga plumbing that is required in Thorium for this simple kind of feature.

For example, this shared knowledge will be useful so that we eventually replace the code that saves the window dimensions / coordinates into a JSON file on the filesystem! (`windowRectangle.json`) See https://github.com/readium/readium-desktop/pull/312

CC @panaC @manuellagrand

src/renderer/components/settings/Information.tsx
```diff
@@ -34,7 +34,7 @@ interface States {
     placeholder: any;
 }
 
-export class LanguageSettings extends React.Component<Props, States> {
+export class AppInformation extends React.Component<Props, States> {
     private parsedMarkdown: string;
```
danielweck: Note that this language-related typo is fixed in this PR, but we could also externalize into a separate PR.

danielweck:
@Scarsniik do you have your own design approach / ideas to solve this problem? Is my PR going in the right direction? Thanks! :)

panaC:
I will see if you want , Cyrille has explained me how the db works..

danielweck:
Great, yes please, I would love your input.
Note that my lack of "expertise" is more about the Redux/Saga flow + state synchronization, rather than about the database per-se. But knowing the DB infrastructure will certainly help! :)
Thanks!

danielweck:
This PR is superseded by https://github.com/readium/readium-desktop/pull/471

Issue #389, opened 2019-07-04
Horizontal scrolling is not blocked


panaC:
![image.png](https://images.zenhubusercontent.com/5d1a0adb1449416ed898590b/a6821af5-0006-4c0f-83fd-0d3a8ef6c677)

It's a bug or a feature ?

danielweck:
I noticed that too during testing with the keyboard. I am not sure how I managed to scroll sideways, I did not manage to reproduce consistently.

Scarsniik:
fixed in 2f4412d

danielweck:
PR: https://github.com/readium/readium-desktop/pull/439

danielweck:
I am not able to reproduce this bug. Any tips?
(MacOS)

PR #439, opened 2019-07-17
fix horizontal scroll in library

Scarsniik:


danielweck:
Fixes: https://github.com/readium/readium-desktop/issues/389

danielweck:
This PR breaks the "skip link", which does not appear anymore when keyboard-tabbing.

Screenshot before the PR code change:

ThoriumSkipLink

PR #440, opened 2019-07-17
Fix/order tab focus

### LINKED ISSUE #337: Incorrect keyboard tabindex in icon/button group in top-right corner
Keyboard tab navigation into these buttons is confusing, partially reversed.

manuellagrand:
This PR fix and re-order buttons in header of the reader to have logical movement of focus .

fixes #337


src/renderer/assets/styles/reader-app.css
```diff
@@ -285,6 +285,55 @@
...
+
+      & .right {
+        float:right;
+        border-left:1px solid #c1c1c1;
+        border-right:none;
+      }
+
+      & .blue {
+        background-color: #2B5366;
+      }
```
danielweck: The CSS in the codediff above looks like a copy-paste from `.main_navigation ul`:

https://github.com/readium/readium-desktop/blob/3bab0fd6d84ce1006e155a2885f1d93259586240/src/renderer/assets/styles/reader-app.css#L245-L284

Let's avoid duplicating CSS.
The only CSS rule that your proposed nested `ul.menu_option` CSS selector adds is: `float: right;`.
So, we can simply use the `,` CSS selector combinator, and apply the `float` as follows:

```
  & > ul, & ul.menu_option {
    &.menu_option {
      float: right;
    }
```

(line 240, inside `.main_navigation`, replace `& > ul {` with the above snippet of CSS code)

PR #441, opened 2019-07-17
add see more button in publication info

Scarsniik:


src/resources/locales/en.json
```diff
@@ -89,7 +89,9 @@
     "progressDownload": "Downloading",
     "endDownload": "Download completed",
     "failedDownload": "Download Failed",
-    "canceledDownload": "Download Canceled"
+    "canceledDownload": "Download Canceled",
+    "seeMore": "See more",
+    "seeLess": "See less"
```
danielweck: @panaC @manuellagrand do any of you speak basic German, to add "See more/less" to https://github.com/readium/readium-desktop/blob/develop/src/resources/locales/de.json ?

danielweck: @Scarsniik could you please add the German strings? Then we can merge this PR.

PR #444, opened 2019-07-18
Fix/drag and drop

### LINKED ISSUE #426: Drag and drop from filesystem: should only accept recognized file types (EPUB, OPDS, LCPL)
Here is what happens when a PNG image file is drag+dropped onto the library view:

ThoriumFileDrop

Code reference:
https://github.com/readium/readium-desktop/blob/541e48bbe45407a8918018e0076d4ca5ebecd68b/src/renderer/components/App.tsx#L52-L65

manuellagrand:
This PR fix drag and drop issue and prevent to the add of file with the wrong extension

fixes #426


src/renderer/components/App.tsx
```diff
@@ -64,6 +66,19 @@ export default class App extends React.Component<any, undefined> {
         ));
     }
 
+    public cleanTabFile(files: File[]): File[] {
+        const tab: File[] = [];
+
+        files.map((file) => {
+            if (file.name.search(".epub") !== -1 ||
+                file.name.search(".opds") !== -1 ||
+                file.name.search(".lcpl") !== -1) {
```
danielweck: Please use `path.extname()` https://nodejs.org/api/path.html#path_path_extname_path

src/renderer/components/App.tsx
```diff
@@ -64,6 +66,19 @@ export default class App extends React.Component<any, undefined> {
         ));
     }
 
+    public cleanTabFile(files: File[]): File[] {
+        const tab: File[] = [];
+
+        files.map((file) => {
+            if (file.name.search(".epub") !== -1 ||
+                file.name.search(".opds") !== -1 ||
```
danielweck: 1) Note that OPDS feeds do not necessarily end with the `.opds` file extension. For example: https://www.feedbooks.com/catalog.atom
2) I do not think OPDS feeds can be drag-dropped onto the Thorium UI anyway, see error:
```
[2]   readium-desktop:main#redux/sagas/api TypeError: Cannot read property 'resources' of null
[2]   readium-desktop:main#redux/sagas/api     at PublicationViewConverter.convertDocumentToView (/Volumes/500GB/Code/readium-desktop/dist/main.js:2891:47)
[2]   readium-desktop:main#redux/sagas/api     at newDocs.map (/Volumes/500GB/Code/readium-desktop/dist/main.js:2494:67)
[2]   readium-desktop:main#redux/sagas/api     at Array.map (<anonymous>)
[2]   readium-desktop:main#redux/sagas/api     at PublicationApi.<anonymous> (/Volumes/500GB/Code/readium-desktop/dist/main.js:2493:28)
[2]   readium-desktop:main#redux/sagas/api     at Generator.next (<anonymous>)
[2]   readium-desktop:main#redux/sagas/api     at fulfilled (/Volumes/500GB/Code/readium-desktop/dist/main.js:2400:58)
[2]   readium-desktop:main#redux/sagas/api     at process._tickCallback (internal/process/next_tick.js:68:7) +0ms
```

src/renderer/components/dialog/FileImport.tsx
```diff
@@ -31,41 +31,46 @@ export class FileImport extends React.Component<FileImportProps, undefined> {
...
+            const paths = this.props.files.map((file: File) => {
+                return file.path;
+            });
+            this.props.importFiles({ paths });
+            this.props.closeDialog();
+        }
     }
 
-    private buildBasicFileImportList() {
+    private buildBasicFileImportList(__: any) {
```
danielweck: No need to add the function parameter `__: any`, as you can see inside the function body there is already usage of `this.props.__`.

src/renderer/components/App.tsx
```diff
@@ -50,11 +50,13 @@ export default class App extends React.Component<any, undefined> {
 
     // Called when files are droped on the dropzone
     public onDrop(acceptedFiles: File[]) {
+        const tabfile: File[] = this.cleanTabFile(acceptedFiles);
```
danielweck: Just out of interest, why the term "tab"? (sorry if I am missing something obvious ;)

src/renderer/components/App.tsx
```diff
@@ -64,6 +66,19 @@ export default class App extends React.Component<any, undefined> {
         ));
     }
 
+    public cleanTabFile(files: File[]): File[] {
```
danielweck: Coding advice: rather than "clean", use a more self-explanatory term such as "filterImportFilesByAllowedExtensions" (this is just an example, feel free to adapt :)

src/renderer/components/App.tsx
```diff
@@ -64,6 +66,19 @@ export default class App extends React.Component<any, undefined> {
...
+        const tab: File[] = [];
+
+        files.map((file) => {
```
danielweck: Do not use `Array.map()` if the code does not use the return value of the function call. Instead, iterate the list of files using `Array.forEach()` or `for (const item of array) { ... }`.

src/resources/locales/fr.json
```diff
@@ -108,6 +108,7 @@
     "available": "Nouvelle version !"
   },
   "dialog": {
+    "importError": "une erreur s'est produite, vérifiez l'extension de votre fichier (.epub,.opds ou.lcpl)",
```
danielweck: Space characters missing in `.epub,.opds ou.lcpl`

src/resources/locales/de.json
```diff
@@ -81,6 +81,7 @@
         "available": "Update verfügbar!"
     },
     "dialog": {
+      "importError": "ein Fehler ist aufgetreten, überprüfen Sie Ihre Dateiendung (.epub,.opds oder.lcpl).",
```
danielweck: Spaces missing in `.epub,.opds oder.lcpl`

src/renderer/components/App.tsx
```diff
@@ -50,11 +50,13 @@ export default class App extends React.Component<any, undefined> {
...
     // Called when files are droped on the dropzone
     public onDrop(acceptedFiles: File[]) {
+        const tabfile: File[] = this.cleanTabFile(acceptedFiles);
+
         this.store.dispatch(
             dialogActions.open(
                 DialogType.FileImport,
                 {
-                    files: acceptedFiles.map((file) => {
+                    files: tabfile.map((file) => {
```
danielweck: This is a minor coding remark: instead of introducing a new function `cleanTabFile()`to filter out some values from the original array of dropped files, you could simply use the `Array.filter()` function with a simple function predicate that excludes non-recognized file extensions, and chain the result (`acceptedFiles.filter(...).map(...)`). See:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

src/renderer/components/App.tsx
manuellagrand: tabfile is for table of file. i can change this variable name if it isn't enough relevant.

danielweck: You mean an "array" / "list"?
In English, a "table" is a matrix (or something you have dinner at ;)

danielweck: 2D matrix = rows+columns

src/renderer/components/App.tsx
```diff
@@ -105,7 +112,7 @@ export default class App extends React.Component<any, undefined> {
                         </Dropzone>
                     </div>
                 </ConnectedRouter>
-            </Provider>
+            </Provider >
```
danielweck: Minor: is this space character intentional?

src/renderer/components/App.tsx
```diff
@@ -54,7 +57,11 @@ export default class App extends React.Component<any, undefined> {
             dialogActions.open(
                 DialogType.FileImport,
                 {
-                    files: acceptedFiles.map((file) => {
+                    files: acceptedFiles.filter((file) => {
+                            return (path.extname(file.path) === ".epub" ||
+                            path.extname(file.path) === ".lcpl");
```
danielweck: Once last nit-picking comment :)

`path.extname(file.path)` is evaluated twice at each iteration. Instead, just store the string value in a constant variable scoped to the function body:

```javascript
files: acceptedFiles.filter((file) => {
    const ext = path.extname(file.path);
    return (ext === ".epub" || ext === ".lcpl");
}
```

PR #446, opened 2019-07-18
Fix/key list

manuellagrand:
this PR fix warning key issue in opds list.


danielweck:
Incrementing the `index` value by an arbitrary value (`3000`) is not an appropriate fix (`key={index + 3000}` in `src/renderer/components/opds/FeedList.tsx`).

The culprit for the console warning message is the duplicated `key` == `0` generated by this code in the same source file (because `-0` == `0`):

```
{[...Array(6).keys()].map((__, index) => {
    return <div key={-index}></div>;
})}
```

See screenshot of the React developer tools.

ThoriumOPDSReactkey

src/renderer/components/opds/FeedList.tsx
```diff
@@ -44,9 +44,8 @@ export class FeedList extends React.Component<OpdsListProps, null> {
                 <ul>
                     { this.props.feeds.map((item, index) => {
                         return (
-                            <li>
+                            <li key={item.title}>
```
danielweck: Using `item.title` is not a valid solution, as there is no guarantee of unicity. See screenshot:

ThoriumOPDSFeedUniqueKeys

danielweck: Suggestion:
```
key={`feed-${index}`}
```

danielweck: Or:
```
key={item.identifier}
```

PR #447, opened 2019-07-18
[Fix #346] Skip link focus the first element of the list

### LINKED ISSUE #346: skip-to-content keyboard accessibility link: should highlight target when clicked (temporary flashing outline)
Note that this is how the navigator handles document targets. I will see if I can use the same techniques in the UI.

Scarsniik:


danielweck:
I am not sure that the UX proposed in this Pull Request is standard practice (i.e. forcing focus on the first button etc. item inside a given "main content" container).

For example, GitHub's "Skip to Content" link (invisible until keyboard-tabbed, just like Thorium) forces focus on a dedicated `div` element: `<div id="start-of-content" class="show-on-focus"></div>`

Another example: MDN's "Skip to main content" link (same invisible/visible strategy) points to a `main` element: `<main id="content" role="main"> ...` ( https://developer.mozilla.org/en-US/ )

In other words, the expectation is that the keyboard / screen reader user is placed in a position that allows him/her to start exploring the content "within" (or "below"), forward of the current focus (which is not necessarily tabbable, by the way).
Conversely, this PR forces focus/tab one step ahead, which removes important contextual information during the UX (especially with screen readers). For instance, in the library view the user immediately lands on the button menu for the first publication in the "recent" list (with no indication of context: just a labelled button without an introductory section "heading"). Instead, the user should land at the very top of the "main content", ready to explore what follows next.

What do you think?

danielweck:
An explanation: why I think this PR addresses a different problem than issue 346:
https://github.com/readium/readium-desktop/issues/346#issuecomment-512851031

Scarsniik:
I think I don't really understand the issue. Before my PR, the focus was put just before the content and you can focus the content with just a tab

danielweck:
The standard practice (see my more complete comment above) seems to be: when the "skip link" is activated / clicked, land the user focus on a "main" container which then allows the user to keyboard-tab forward, into the actual content. This PR goes one step too far (I think), by prematurely landing the user onto isolated controls (such as buttons, hyperlinks, form inputs, etc.), thereby removing the navigation context that users normally have when landing "at the top of" a main content area (e.g. `main` element). This is especially true for screen readers which scan HTML markup forward, in a much more fine-grained manner than keyboard-users (who are limited to keyboard-focusable elements only).

Scarsniik:
But it was already like this. The focus was just at the start of the main section

danielweck:
I know :) This is why I am asking.

danielweck:
Closing this PR based on consensus (non-standard UX)

Issue #303, opened 2019-06-14
keyboard navigation - choosing a language


llemeurfr:
I should be able to select a language by (this is my interpretation, maybe for 100% accessibility purpose it should be different): 
- tab to Settings + return
- tab to UI language + return
- down arrow to select a language + return (it's a sort of drop-down menu isn't it?)

Currently: 
- tab to Settings + return
- tab to UI language 
- tab to Information
- tab to English
- down arrow to select a language, automatically selected

which cannot be good.

llemeurfr:
Plus in the Settings view the "skip to content" link jumps to I don't know where. I think it should jump to the "UI language" label.

danielweck:
Laurent, I filed a separate issue about "skip to content" https://github.com/readium/readium-desktop/issues/346

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/350

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/365

llemeurfr:
After careful thinking, I came to the conclusion that the UX is too complex for the need. We have currently in Settings a set of two tabs, the tab panel of the "UI language" tab being a set of radio buttons. Accessibility cannot be good in this case. 

We will therefore simplify the UX:

1/ remove the nav containing the two tabs ("UI language" and "Information").
2/ keep the "Language choice" label and the set of radio buttons in this screen (other settings may come later).

I found a reference document about how to deal with accessibility of radio buttons at https://www.w3.org/TR/wai-aria-practices/examples/radio/radio-1/radio-1.html. 

Please follow these guidelines as much as possible (use of role radiogroup, radio, aria-checked). There are two variants, I don't know what is best. Please advise. 

The "Information" tab must be moved elsewhere, as it is difficult to find in Settings.
As a start, let's insert it a the bottom of the My Books screen, below My Tags. 
- The title of the section, same size as My Tags, is "About Thorium".
- below, the software version (V1.0.0.beta.2) in normal size, bold
- below, a link "More information" which opens a modal box containing the content of the .md

Scarsniik:
Fixed in #461

PR #461, opened 2019-07-22
Move the about section in the main page

Scarsniik:
fixe issue #303


danielweck:
The "about Thorium" / "more information" section is located below the bookshelf, at the very bottom. It is not easily discoverable / reachable once publications are imported. Is that okay? (see screenshots)

ThoriumAbout1

... scroll ... (or keyboard-tab)

ThoriumAbout2

danielweck:
* The "more information" box does not evidently look like a modal popup dialog, because of the lack of left/right margins.
* There is a vertical scrollbar in the main app area, which should not exist (needs overflow hidden?)
* The information text is horizontally-centered, which makes the bullet points look very awkward (I suggest left-align, or justify).

See screenshot:

ThoriumAbout3

llemeurfr:
Yes this is what I requested to Leo. I didn’t want it to bloat the navigation bar and didn’t want it to stay in Settings (it is not). 
We may rethink that later but for now it solves a large accessibility issue we had before. And people scroll. 

My only remark so far : the alignment of the elements is discutable, as is the Lack of space between the carousel and the title of the section.

llemeurfr:
And I agree with Daniel’s comments about the modal window.

danielweck:
When the "more information" dialog is closed (escape-key or cross-button-click or mouse-click-outside), the keyboard focus should return to the calling site (button).

danielweck:
There is now a code conflict (please synchronize with the tip of the develop branch :)

danielweck:
Side note -- I have found "about-window" pretty useful (other Electron projects):
https://github.com/rhysd/electron-about-window#screen-shots

...but it is only good for short content (description summary, copyright, version(s), links).

So, although it would partly solve the "viewport" problem, it does not solve the "calling site" issue (in other projects, there is a menu bar with an "about" menu item).

llemeurfr:
I'd like to see "V1.0.0-beta.2"

PR #475, opened 2019-07-24
WIP (DO NOT MERGE) attempt to identify and solve of strictNullChecks errors (many)

danielweck:
This code changes in this PR are much less trivial than https://github.com/readium/readium-desktop/pull/404
(because code logic is affected, due to `null` and `undefined` having very different meaning and applications)

In my analysis of the current codebase, there are many instances of unhandled null/undefined values, including promises that should reject instead of resolving, etc.


danielweck:
Note that TypeScript 3.6 will provide strongly-typed iterators and generators, which means that the returned values of Redux-Saga `yield call()` will not have to be typed explicitly. https://github.com/microsoft/TypeScript/issues/31639

PR #481, opened 2019-07-26
http request update

### LINKED ISSUE #463: OPDS feeds, set Accept-Language
Some OPDS feeds (e.G. Feedbooks) use the http Accept-Language property to select a localized feed for the user. 

Accept-Language should be set to the language chosen by the user in the app.

panaC:
In the base code, they are 2 httpRequest based on 'request' -> I removed them and I created an httpGet in common/utils directory.

I have updated all the code with this new httpGet

httpGet take in parameter the url and options typed and returns the httpRequestResponse type.

I added a new header : Accepted-Language with the language in app from redux-state.

It's works great on Win and Mac.

On feedbooks opds feed, the language is automatically updated in function of the app language.

fix #463


src/common/utils/http.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const store = container.get("store") as Store<RootState>;
+    const locale = store.getState().i18n.locale;
+    const requestOptions = Object.assign(
+        {},
+        { url },
+        {
+            headers: {
+                "User-Agent": "readium-desktop",
+                "Accept-Language": `${locale}-${locale.toUpperCase},en-US;q=0.7,en;q=0.5`,
```
danielweck: I think the function should be invoked inside the string template literal, i.e. `${locale.toUpperCase()}` instead of `${locale.toUpperCase}`.

danielweck: If `locale` == `fr` or `de`, then this works:
`${locale}-${locale.toUpperCase()}`
=>
`fr-FR`
...but note that there are many other `fr` language codes, such as `fr-CA`, `fr-BE`, `fr-LU`, etc.

However, `locale` == `en` results in `en-EN`, which is incorrect (this would normally be `en-US` or `en-GB`, but there are many others, e.g. `en-CA`, `en-AU`, etc.). 

Also note that the full language code can be lower-case.

References:
https://tools.ietf.org/html/rfc7231#section-5.3.5
https://tools.ietf.org/html/bcp47

src/common/utils/http.ts
```diff
@@ -0,0 +1,33 @@
...
+import { promisify } from "util";
+
+export type httpResponse = request.Response;
+
+export function httpGet(url: string, options?: request.CoreOptions): Promise<request.Response> {
+
+    const store = container.get("store") as Store<RootState>;
+    const locale = store.getState().i18n.locale;
+    const requestOptions = Object.assign(
+        {},
```
danielweck: Can this empty object `{}` be removed?

src/common/utils/http.ts
panaC: I have removed fr-FR

We handle only the first language locale.

It's works on feedbooks

panaC: I know

We can't add a best language code than i18n.locale state in App.

what do you think ?

src/common/utils/http.ts
panaC: `Object.assign(target, ...sources)` it's the target

src/common/utils/http.ts
```diff
@@ -0,0 +1,33 @@
...
+        { url },
+        {
+            headers: {
+                "User-Agent": "readium-desktop",
+                "Accept-Language": `${locale}-${locale.toUpperCase},en-US;q=0.7,en;q=0.5`,
+            },
+        },
+        options,
+    );
+    return promisify<request.CoreOptions, request.Response>(request)(requestOptions);
```
danielweck: I think that the original `httpGet` promise-wrapper code for the vanilla `request` API was incorrect because HTTP response status codes were not handled properly (e.g. `404` with a perfectly valid `body`, no `error`). I think using `promisify` perpetuates this.

I propose this fix (note that the `request` lib follows HTTP `3xx` redirects by default, so no need to handle this explicitly):

```javascript
return new Promise((resolve, reject) => {

request({
url,
method: "GET",
encoding: undefined, // note that null means Buffer, undefined means .toString(undefined), which defaults to UTF8
}, (error, response, body) => {
    if (error) {
        reject(error);
        return;
    }
    if (!response) {
        // this should never happen because error is falsy,
        // but we fail-fast to move onto the next response check ...
        reject(new Error(`No HTTP response?! ${url}`));
        return;
    }
    if (response.statusCode < 200 || response.statusCode >= 300) {
        const b = body ? JSON.stringify(body) : "";
        reject(new Error(`HTTP response status code: ${url} => ${response.statusCode} => ${b}`));
        return;
    }
    // response.statusCode >= 200 && response.statusCode < 300
   if (!body) {
        reject(new Error(`HTTP no body?! ${url} => ${response.statusCode}`));
        return;
   }

   resolve(body); // this should be UTF8 string, given the default encoding option
});

});
```

danielweck: Naturally, this means removing `.body` from `opdsFeedData.body`, `lsdStatusBodyResponse.body`, `lcplResponse.body`, etc. in other modified code in this PR.

src/common/utils/http.ts
danielweck: You can target `{ url }`, or even better, group it with `headers` to avoid unnecessary multiple parameters:

```
        {
            url,
            headers: {
                "User-Agent": "readium-desktop",
                "Accept-Language": `${locale},en-US;q=0.7,en;q=0.5`,
            },
        },
```

src/common/utils/http.ts
panaC: done

src/common/utils/http.ts
danielweck: Let's just use `${locale}` and assume the app's configured locale will always be simple language codes like `en`, `de` and `fr`.

panaC: already done, thanks

src/common/utils/http.ts
danielweck: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

panaC: already done, thanks

src/common/utils/http.ts
```diff
@@ -0,0 +1,44 @@
...
+            method: "GET",
+            encoding: undefined,
+            headers: {
+                "User-Agent": "readium-desktop",
+                "Accept-Language": `${locale},en-US;q=0.7,en;q=0.5`,
+            },
+        },
+        options,
+    );
+    const response = await promisify<request.CoreOptions, request.Response>(request)(requestOptions);
```
danielweck: This does not handle the `error` parameter callback of the original `request` API - which, when wrapped in `util.promisify`, becomes a promise rejection (which must be explicitly captured via `.catch()` or `try {} catch {}`).

Please refer to my proposal again, and see how the edge cases are more explicitly covered. Using `util.promisify` only adds confusion / obfuscation of the original `request` API.

PS: I think `httpGet` can reliably return `Promise<string>`, given `encoding: undefined` (not `null`) in the `request` options. This is currently our main use-case in the codebase, let's not enable `Buffer` for now.

src/common/utils/http.ts
```diff
@@ -0,0 +1,44 @@
...
+        {
+            url,
+            method: "GET",
+            encoding: undefined,
+            headers: {
+                "User-Agent": "readium-desktop",
+                "Accept-Language": `${locale},en-US;q=0.7,en;q=0.5`,
+            },
+        },
+        options,
```
danielweck: I've just realized: we must ensure that certain options cannot be overridden by the `options` function parameter.
1) `options.url` (otherwise, this creates confusion with the first function parameter `url`)
2) `options.method` (otherwise, the function cannot be called `httpGet()`)
3) `options.headers["User-Agent"]` (this is debatable, but I personally think API consumers of `httpGet()` should not be given a choice here)
4) `options.headers["Accept-Language"]` (same as above)
5) `options.encoding` (I propose that we guarantee that `httpGet()` returns `Promise<string>` (as per my other comment about `util.promisify`), so `encoding` should not be overridden by `httpGet()` callers)

Any other thoughts?
https://github.com/request/request#requestoptions-callback

danielweck: PS: I see two ways to prevent overriding:
1) `delete`:
```javascript
if (options) {
  if (typeof options.url !== "undefined") {
    delete options.url;
  }
  if (typeof options.encoding !== "undefined") { // note that this covers `null` too, as typeof null === "object"!
    delete options.encoding;
  }
}
```

2) `Object.assign()` override (place high-precedence object as last parameter)

src/renderer/utils.ts
```diff
@@ -22,8 +22,8 @@ export function buildOpdsBrowserRoute(
 
 // Encode without padding
 // And apply RFC 4648
-export function encodeB64(data: any) {
-    let encoded = btoa(data);
+export function encodeB64(data: string) {
+    let encoded = btoa(unescape(encodeURIComponent(data)));
```
danielweck: I need to review this carefully, based on the actual usage of the `encodeB64()` function.
I speak from extensive experience -- in developing the `r2-xxx-js` packages in particular, where Base64 and URL/URI encoding is critical, and error-prone -- working with the subtleties of `unescape`/`escape`, `decodeURIComponent`/`encodeURIComponent` and also `decodeURI`/`encodeURI`).

I will come back to this later (sorry cannot do it right now).

src/common/utils/http.ts
panaC: - No, it' handle error because promisify throw error with the promise . your example is not explicit either. No ?
- Why promisify add confusion : it's just to return a promise on a function with a callback at the end of the function ?
- Why return body and not response ? It allow to check response inside code and to access directly on body ? Cyrille has begin like this in code.

src/renderer/utils.ts
panaC: I need this to works : https://stackoverflow.com/questions/23223718/failed-to-execute-btoa-on-window-the-string-to-be-encoded-contains-characte

src/common/utils/http.ts
panaC: I will use the second method

src/common/utils/http.ts
danielweck: `await` without an explicit `try`/`catch` (or the functional `.catch()` equivalent) is by definition an "unhandled promise rejection" (which there are several in the `readium-desktop` code, unfortunately). Exception handling in the context of promises has different semantics and design patterns than regular cascaded exceptions used to control unexpected side effects.

danielweck: I look at the original `httpGet()` and I see `resolve(body);`, not `resolve(response);` (where `body` is guaranteed to be type `string` is `options.encoding` is `undefined` (not `null`)). I would like the code to continue using this design feature, as it makes it simpler for API consumers. In a future revision of the code, if there is a justification for returning the full HTTP `response`, then we may do it (note that this can be useful for streaming the HTTP payload, in which case the `response.body` is not available in full).

src/common/utils/http.ts
danielweck: I think it is a better choice too, because `delete` would modify the passed object, which might break the API consumer / client side of `httpGet()`.

src/common/utils/http.ts
panaC: - The throw inside an async function return catch(e) with the error in parameter ? No ? I didn't see why you don't like promisify ?

- there was 2 requestGet/httpGEt : https://github.com/readium/readium-desktop/blob/develop/src/main/redux/sagas/update.ts

src/common/utils/http.ts
panaC: done

src/common/utils/http.ts
panaC: httpGet returns response.body now

panaC:
New issue linked to this PR #484 

any idea ?

src/common/utils/http.ts
```diff
@@ -0,0 +1,48 @@
...
+import * as request from "request";
+import { promisify } from "util";
+
+export async function httpGet(url: string, options?: request.CoreOptions): Promise<string> {
+
+    const store = container.get("store") as Store<RootState>;
+    const locale = store.getState().i18n.locale;
+    const headers = Object.assign(
+        {},
+        options.headers,
```
danielweck: `options` can be `undefined` (`options?: request.CoreOptions`).

`options.headers` can be `undefined` / `null`, but `Object.assign()` gracefully ignores:
"Note that Object.assign() does not throw on null or undefined source values."
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/common/utils/http.ts
```diff
@@ -0,0 +1,48 @@
...
+        {
+            "User-Agent": "readium-desktop",
+            "Accept-Language": `${locale},en-US;q=0.7,en;q=0.5`,
```
danielweck: Strictly-speaking this is problematic, so I must flag it up (however, note that in practice the `readium-desktop` code is unlikely to trigger the edge-case):

Callers of the function `httpGet()` can pass `options.headers` = `{ "user-agent":  "MyOverriddenUserAgent" }` (note the lower-case key ... HTTP headers are case-insensitive). This will result in the concatenation of `"readium-desktop"` and `"MyOverriddenUserAgent"`, because the `request` lib utilises `caseless` in order to ignore case variants (see https://github.com/request/request/blob/212570b6971a732b8dd9f3c73354bcdda158a737/request.js#L159 and https://github.com/request/caseless/blob/98079105e15e04308d1f3becc3e8efab8ddd7f0e/index.js#L59 ).

One solution would be to force-lower-case the object/dictionary keys of a `options.headers` clone, and to use lower-case `"user-agent"` and `"accept-language"` in order to ensure this takes precedence in `Object.assign()`.

src/common/utils/http.ts
```diff
@@ -0,0 +1,48 @@
...
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END==
+
+import { container } from "readium-desktop/main/di";
+import { RootState } from "readium-desktop/main/redux/states";
+import { Store } from "redux";
+import * as request from "request";
+import { promisify } from "util";
+
+export async function httpGet(url: string, options?: request.CoreOptions): Promise<string> {
```
danielweck: if `options.json` === `true` (or truthy), then `httpGet()` returns a promise of type `any` (raw JSON), not `string`. Therefore, I suggest the function signature explicitly returns `Promise<string | any>`, with a code comment to explain the reason for `any` in the TypeScript union type.

The alternative option would be to `delete options.json`, but I feel uncomfortable about this, because this breaks the API contract of the original function, and requires adapting the client-side consumer code (i.e. to add explicit `JSON.parse()`).

What do you think?

src/common/utils/http.ts
panaC: done

src/common/utils/http.ts
panaC: i didn't know about that, I removed the returns type

src/common/utils/http.ts
panaC: done

src/common/utils/http.ts
danielweck: I see you removed the TypeScript return type in the function signature, why not use the union type `Promise<string | any>`? (yes, `string` is redundant with `any`, but there is an API comment to explain `json:true` => `any`).

panaC: if you want, yes it's redundant

panaC: Screenshot 2019-07-29 at 16 05 57

the interest is really limited anyway.

danielweck: Thanks. This way, both the TypeScript compiler and the human programmer have an explicit reference to the `json` edge case.

src/renderer/utils.ts
danielweck: FYI, related issue: https://github.com/readium/readium-desktop/issues/486
(let's handle this in a separate Pull Request)

src/common/utils/http.ts
panaC: yes and no the string return type option is only available when the dev see the file. The information panel, so also the typescript compiler,  overload string type with any.
Cf image above.

Yes the human programmer can have an explicit reference but more with the comment on function returns type

src/renderer/utils.ts
danielweck: This Pull Request (481) is about fixing / harmonizing the HTTP request utility functions.

I suggest that the Base64-related code changes (which impact the OPDS React router routes) should be removed, and treated in a separate Pull Request. Could you please file an issue to explain what the problem is exactly (reproducible bug), or alternatively just create a Pull Request that explains the issue itself. Thank you.

panaC: Ok, but my PR become obsolete without this

panaC: How create a new issue, if the bug is reproducible only with a specific accept-language on request for me ?

src/common/utils/http.ts
danielweck: Well, instead of `any` we can use the following `AnyJson` type:

`declarations.ts`:
```
export type AnyJson =  JsonPrimitives | JsonArray | JsonMap;
export type JsonPrimitives = string | number | boolean | null;
export interface JsonMap {  [key: string]: AnyJson; }
export interface JsonArray extends Array<AnyJson> {}
```

panaC: Thanks i apply it

src/renderer/utils.ts
danielweck: I understand the original logical coupling of HTTP request <-> Base64 encoding/decoding, due to how the Base64 bug surfaced in your tests. It is now also apparent that there should be two separate well-scoped Pull Requests of single responsibility. We must be able to test and validate the Base64 code alterations outside of the context of the refactoring of the HTTP API, which - as you have observed - revealed several subtle side effects.

In another PR (I will file issues later) we must properly handle `httpGet()` promise rejections (`try`+`catch`) and `finally` on the client side, so that error recovery from the `request` `error` parameter (e.g. request timeout), or from totally unexpected exceptions (which also reject the promise) can be used to manage the UI state (e.g. progress bar). I saw that the Redux-Saga `yield call` already handles rejection correctly, but others do not (resulting in "unhandled promise rejections").

Let's have a separate "clean" PR (based on `develop`, using the harmonized/centralized HTTP utility function) to address the Base64 encoding/decoding (as you probably saw, there is already a separate issue about broken handling of `+` characters). This will be easier to test and rollback if necessary.

src/common/utils/http.ts
danielweck: Do you think we can use `typings.d.ts`? (that is what I meant by `declarations.d.ts`)
https://github.com/readium/readium-desktop/blob/develop/typings.d.ts

src/renderer/utils.ts
panaC: ok , let's do it

But explain me the particularity of encodeB64 and decodeB64 : it's used only for the opdsBrowserRoute. the drift is limited

danielweck: Sure, but for example in other `r2-xxx-js` packages, we use an `encodeURIComponent()` implementation that conforms to RFC3986 (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent ) ...so, better isolate these types of code changes.

src/common/utils/http.ts
panaC: we can't use the typing declaration with interface , no ?
I created a new file.ts especially for this

panaC: We have to create a new module otherwise

danielweck: Ok, thanks :)

panaC:
I changed all conflict lines about yesterday discussion

Be careful @danielweck to resolve encode 64 and decode B64 issues quickly after the merge, please

panaC:
Why tslint throw errror ? @danielweck 

ERROR: src/typings/json.d.ts:8:1 - 'namespace' and 'module' are disallowed

src/renderer/utils.ts
danielweck: New issue: https://github.com/readium/readium-desktop/issues/522

src/common/utils/http.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Store } from "redux";
+import * as request from "request";
+import { promisify } from "util";
+
+/**
+ * @param url url of your GET request
+ * @param options request options
+ * @returns body of url response. 'String' type returned in many cases except for options.json = true
+ */
+export async function httpGet<T extends AnyJson = string>(url: string, options?: request.CoreOptions): Promise<T> {
```
danielweck: Hello @panaC you will be interested to know that the `request.CoreOptions` TypeScript type is incorrect when using stricter checking rules. I implemented a fix in my Pull Request:
https://github.com/readium/readium-desktop/pull/404/commits/79dbc9ed86646beced0421631ee215004d24324a

PR #474, opened 2019-07-24
Fix/translate

manuellagrand:
fix some translation issues

PR #483, opened 2019-07-26
Fix/manage close window

### LINKED ISSUE #454: MacOS: Apple key + W does not close windows

manuellagrand:
This PR add "Command+W" management for closing windows

fixes #454


src/main.ts
```diff
@@ -234,6 +234,9 @@ app.on("ready", async () => {
     debug("ready");
     initApp();
 
+    globalShortcut.register("CommandOrControl+w", () => {
```
danielweck: The proposed solution is incorrect. `globalShortcut` is for "global" shortcuts :)
(system-wide)
See:
https://electronjs.org/docs/api/global-shortcut

The solution is to re-instate the typical "window" menu which has built-in accelerator shortuts, i.e. `role: "close"`, `role: "front"`, `role: "minimize"`, `role: "togglefullscreen"`, etc. and for MacOS (darwin) also in the first "about" menu `role: "services"`, `role: "hide"`, `role: "hideothers"`, `role: "unhide"`, `role: "quit"`.

See:
https://electronjs.org/docs/api/menu#examples

danielweck:
Also note that the number of unrelated commits in this PR indicate that your local copy of the `develop` branch needs to be updated from the `upstream` remote.
Use `git fetch upstream` and update your `fix/AddTagDialog` branch with `git merge upstream/develop` (resolve conflicts as appropriate, if any).

danielweck:
I think a clue as to why your PR contains abnormal / unrelated commits is because of PR 474 (I can see it in your commit history) which had to be "reverted" using a forced-push in order to rewrite the polluted Git commit history in the develop branch. See:
https://github.com/readium/readium-desktop/pull/474#issuecomment-514671174

That being said, your pull requests / local feature branches should always be "forked" off from `upstream/develop`, which was cleaned immediately after the erroneous PR merge (474). Perhaps you created your local branch from your own local out-of-date copy of `develop`?

danielweck:
Given the erroneous state of the commit history, and the fact that the proposed solution to #454 is incorrect, I take the decision to close this PR.

PR #498, opened 2019-07-31
Fix cases where the navigation bar disappears

Scarsniik:


danielweck:
Let's merge this early, so people can test. Personally I am not able to reproduce the issue on MacOS (my main computer), but I managed to reproduce on Windows 10 using the store-deployed app.

danielweck:
Hello @Scarsniik I am afraid this workaround introduces a regression:
https://github.com/readium/readium-desktop/issues/505

Issue #476, opened 2019-07-24
German untranslated OPDS strings


danielweck:
![ThoriumOPDSGerman](https://user-images.githubusercontent.com/625201/61831833-03cad600-ae67-11e9-9c6f-190836bfe3a2.png)

danielweck:
Note: I thought there was a fallback to English configured in `i18next`, but this is visibly not the case as shown by `opds.addForm.namePlaceholder` in the screenshot.

danielweck:
In the PR that discovers missing locale keys, I have also implemented a consistent fallback to English:
https://github.com/readium/readium-desktop/pull/506

danielweck:
This issue is superseded by https://github.com/readium/readium-desktop/issues/515

Issue #397, opened 2019-07-08
Localization JSONs: compare languages to detect missing keys, remove obsolete / unused keys


danielweck:
e.g. `svg.contentTable`

Also:
https://github.com/readium/readium-desktop/blob/024dc4e64c645dc150432a232e0b65912dcb10d6/src/resources/locales/en.json#L217
(there may be many more ... how do we automate the discovery of obsolete l10n labels?)

danielweck:
Related: https://github.com/readium/readium-desktop/issues/396

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/413

danielweck:
Potential tools:
https://www.i18next.com/overview/plugins-and-utils#extraction-tools

http://i18next.github.io/i18next-scanner/
https://github.com/i18next/i18next-parser

danielweck:
In another project, I use the NPM package `i18next-json-sync` (CLI utility) to ensure all locales (in our case here: `fr` and `de`) are consistent with a master language (in our case here: `en`). 

Missing JSON keys are added by the CLI utility, redundant keys are removed.

Note that keys are automatically sorted so that the tool can compare canonical-ish representations. So, the initial pass creates a big diff, just for the re-ordering.

https://github.com/jwbay/i18next-json-sync

For the avoidance of doubt regarding the large initial commit (key sorting), I used the following shell script to sort the locale JSONs (using `jq`, a well-established native/binary CLI JSON manipulation tool):

```
#!/bin/bash

## https://stedolan.github.io/jq/manual/
JQ_PARAM="def recsort: if type==\"array\" then [.[]|recsort]|sort elif type==\"object\" then (.[]) |= recsort else . end; recsort"

find ./packages -maxdepth 100 -type f -name "*.json" -regex "\./src/resources/locales/[^/]*\.json" -print0 | while IFS= read -rd '' relativefilepath; do
    filename=`basename "${relativefilepath}"`
    echo "${relativefilepath} ===> ${filename}"

    jq -S "${JQ_PARAM}" --indent 4 "${relativefilepath}" > "${relativefilepath}.tmp" && mv "${relativefilepath}.tmp" "${relativefilepath}"
    git --no-pager diff "${relativefilepath}"
done

git status
```

Then, I commit the JSON-sorting changes to my local Git repository clone, and I invoke the `i18next-json-sync` CLI utility to track key addition/removal:

`package.json`
```json
  "scripts": {
[...]

PR #506, opened 2019-07-31
JSON locale checking (requires canonical key sorting), discovered lots of missing German keys, and a few redundant

danielweck:
Fixes https://github.com/readium/readium-desktop/issues/397

Fixes https://github.com/readium/readium-desktop/issues/476


danielweck:
@panaC the part that is now missing is the scanning of source code (AST or regexp, whichever works reliably) in order to discover all the localization keys, and verify that the primary `en` locale contains them (no missing, no redundant keys). The secondary locales can be verified against the primary easily using the utilities in this PR.
So, have you made any progress with regards to searching the source code for all the keys?

danielweck:
I am happy to report that `en` and `fr` are well-aligned (no missing keys, no unused keys). However `de` is in bad shape, as shown in this commit:
https://github.com/readium/readium-desktop/pull/506/commits/efa423a859042592160a37d93e9fcb26db00aec0
(tip: search for all empty strings `""`, and this returns all the missing locale keys)

danielweck:
Note that in order to deal with empty string values `""` which indicate missing JSON keys (there are many in the German `de` locale), I patched the translator code:

https://github.com/readium/readium-desktop/pull/506/commits/f50a234b2e5a0945ee000cc699fefa92dc587aa4

Now we have consistent fallback to English when there are missing keys, or when the values are empty strings.

danielweck:
I created a separate issue about scanning the source code: https://github.com/readium/readium-desktop/issues/509

danielweck:
Hello Pierre @panaC could you please send me your thoughts / review on this PR, I would like to merge it early so that any upcoming addition to the locales is based on the cleaned-up format (checked JSON keys) Thank you! :)

panaC:
No problems, I'm always in seeking a solution to correctly parse the source code locale

Just a question: why use yarn and not npm ?

danielweck:
Ah thank you, `yarn` usage slipped through the net! (from another project) I will correct that immediately :)

danielweck:
Ok, I am going to merge.
I will also file an issue about replacing (in a future revision) the current shell script which creates a canonical / key-sorted version of the locale JSONs, with a cross-platform NodeJS solution (hopefully a simple one-liner NPM script based on an existing package). The reason behind using the native shell / CLI utility (which is itself cross-platform, but requires external install) was to rely on a well-tested implementation, so that our locales do not get screwed-up by some version `0.0.x` lib found on NPM ;)

danielweck:
New issue about canonical JSON key sorting (replace `jq` shell executable):
https://github.com/readium/readium-desktop/issues/514

danielweck:
New issue specifically about the missing German keys:
https://github.com/readium/readium-desktop/issues/515

PR #513, opened 2019-07-31
[clean] reader.tsx around queryParam

panaC:
proposal to clean queryParam in reader.tsx

### DIFF
--- a/src/renderer/components/reader/Reader.tsx
+++ b/src/renderer/components/reader/Reader.tsx
@@ -159,8 +159,8 @@ const computeReadiumCssJsonMessage = (): IEventPayload_R2_EVENT_READIUMCSS => {
 };
 setReadiumCssJsonGetter(computeReadiumCssJsonMessage);
 
-const publicationJsonUrl = queryParams.pubId.startsWith(READIUM2_ELECTRON_HTTP_PROTOCOL) ?
-    convertCustomSchemeToHttpUrl(queryParams.pubId) : queryParams.pubId;
+const publicationJsonUrl = queryParams.pub.startsWith(READIUM2_ELECTRON_HTTP_PROTOCOL) ?
+    convertCustomSchemeToHttpUrl(queryParams.pub) : queryParams.pub;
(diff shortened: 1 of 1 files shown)


src/renderer/components/reader/Reader.tsx
```diff
@@ -462,7 +455,7 @@ export class Reader extends React.Component<ReaderProps, ReaderState> {
             // https://github.com/electron/electron/blob/v3.0.0/docs/api/breaking-changes.md#webframe
             // publicationJsonUrl is READIUM2_ELECTRON_HTTP_PROTOCOL (see convertCustomSchemeToHttpUrl)
             // publicationJsonUrl_ is https://127.0.0.1:PORT
-            response = await fetch(publicationJsonUrl_);
+            response = await fetch(publicationJsonUrl);
```
danielweck: Can we please rename `publicationJsonUrl` as `queryParams.pub`, and `publicationJsonUrl_` as `publicationJsonUrl` in the comment immediately above this line of code, thanks :)

danielweck: There is a regression in this PR, because `installNavigatorDOM()` is still called with `publicationJsonUrl`, when it should have been adjusted to the `READIUM2_ELECTRON_HTTP_PROTOCOL` `queryParams.pub` URL. I am working on a fix...

danielweck:
Regression in this PR: R2 navigator uses the localhost / 127.0.0.1 URL instead of the special URL protocol which isolates per-publication domain.
See:
https://github.com/readium/readium-desktop/pull/513/files#r383369057

danielweck:
Regression fixed in https://github.com/readium/readium-desktop/commit/36fbca41c4307b4a0703164cd9873bc7364d47ea

src/renderer/components/reader/Reader.tsx
danielweck: Fixed via https://github.com/readium/readium-desktop/commit/36fbca41c4307b4a0703164cd9873bc7364d47ea

Issue #509, opened 2019-07-31
Localization JSONs: scan the source code to create a complete list of all locale keys, and verify master English locale (no missing, no redundant keys)


danielweck:
This is an adjacent issue to https://github.com/readium/readium-desktop/issues/397

danielweck:
@panaC I have assigned this to you, as I remember you saying that you had started working on a solution based on regular expressions. Feel free to unassign if you prefer.

danielweck:
Potential tools:
https://www.i18next.com/overview/plugins-and-utils#extraction-tools

http://i18next.github.io/i18next-scanner/
https://github.com/i18next/i18next-parser
https://github.com/Cellule/i18next-static-analysis

danielweck:
I am thinking of the following implementation strategy:

1) scan the source code (AST or regexp), build `LIST` an exhaustive flat list of all the keys invoked by `i18next`, for example `translator.translate("app.edit.undo")` => `app.edit.undo`. Note: because of code like this: `const translator = container.get("translator") as any;` (note `any` instead of `Translator`) or simply `const translator = container.get("translator");` (`any` is default type), the TypeScript AST does not know that this is in fact the function signature `translate(message: string, options: any = {}): string`, which takes an optional `options` parameter ... so beware during code analysis! Also note that the function is often aliased, e.g. `const translate = translator.translate.bind(translator); this.props.__ = translate;`
2) remove duplicates in `LIST` (e.g. multiple `translate()` calls to locale key `app.edit.undo`)
3) create a JSON object based on each item in `LIST`, by expanding all keys. For example: `app.edit.undo` => `{ "app": { "edit": { "undo": "" } } }` (note the empty string "" for the `undo` JSON property)
4) "canonicalize" the JSON (key sorting)
5) create the JSON file `src/resource/locales/temp.json` (next to `en.json`, `fr.json`, etc.)
6) use the `i18next-json-sync` NPM package (see existing `i18n` script in `package.json`), and launch: `sync-i18n --files 'src/resources/locales/*.json' --primary temp --languages en fr de --space 4 --finalnewline --newkeysempty` => this will probably remove several unused JSON keys in `en`, `fr` and `de` (hopefully this will not add any missing ones!).
7) `git status && git --no-pager diff` to see the changes
8) remove `src/resource/locales/temp.json`
9) double-check manually that removed keys are *really* not used anywhere in the codebase (just in case the scanning algorithm missed something), and `git commit -a && git push`
10) celebrate.

[...]

PR #516, opened 2019-08-01
codebase scanning for used i18next locale keys

danielweck:
Fixes: https://github.com/readium/readium-desktop/issues/509


danielweck:
@panaC the key regular expression used in `scripts/translate-scan.js` is `(\.translate|__)\(['"]([^\]'"]+)['"]`.

I tested it in VisualStudioCode as well, to be sure.

=>
`161` matches (`138` once duplicates are removed), over `239` source code files (`*.ts` and `*.tsx`).

(note that the matches include code comments)

If you agree, I would like to merge these changes as soon as possible (so we can work with cleaned-up locale JSONs from now on).

danielweck:
Follow-up hotfix PR: https://github.com/readium/readium-desktop/pull/527

Issue #514, opened 2019-08-01
Replace shell scripts with more portable NodeJS script / NPM package


danielweck:
TODO: replace JSON key-sorting utility (native cross-platform CLI utility "jq") with more portable NodeJS script / NPM package:

`jq`:
https://stedolan.github.io/jq/manual/

...was introduced in this PR:
https://github.com/readium/readium-desktop/pull/506

Actual shell script:
https://github.com/readium/readium-desktop/blob/c6f6d0a3d3f839752109eca374cfd38db2884948/scripts/locales-utils.sh#L1-L26

danielweck:
TODO: replace JSON patcher with more portable NodeJS script / NPM package:

https://github.com/readium/readium-desktop/blob/1352b0170b7a622f01bc4bab6e8c727421c31dad/scripts/i18n-typed-patch.sh#L1-L5

danielweck:
Re-opening to address the second shell utility (prefix + suffix on JSON locale)

PR #517, opened 2019-08-01
[fix #509] add i18n typed check

### LINKED ISSUE #509: Localization JSONs: scan the source code to create a complete list of all locale keys, and verify master English locale (no missing, no redundant keys)

panaC:
I created a i18n typing with a tools https://github.com/kogai/typed_i18n/issues

It's allow to check if a key is typed. automatic typing with vscode on __("mykey");

To regenerate the typing file: `npm run i18n-typed`

It's doesn't allow to check redundant keys.

fix #509

be careful to generate the typing file with `npm run i18n-typed` at each update of json locale file


danielweck:
Sophisticated approach, I like it :)
(reviewing...)

src/common/services/translator.ts
```diff
@@ -15,20 +15,11 @@ import * as frCatalog from "readium-desktop/resources/locales/fr.json";
 import * as deLang from "readium-desktop/resources/locale-names/enLang.json";
 import * as enLang from "readium-desktop/resources/locale-names/enLang.json";
 import * as frLang from "readium-desktop/resources/locale-names/frLang.json";
+import { TFunction } from "readium-desktop/typings/en.translation";
 
 const initI18n = i18n.init({
+    resources: Object.assign(enCatalog, deCatalog, frCatalog),
```
danielweck: I have a request: by design, the various JSON locale files are currently (i.e. `develop` branch) comparable in their canonical form, for example key sorting + value removal/ignoring. This makes it trivial order to detect missing or redundant keys across multiple languages, using off the shelf tooling (as implemented in the `develop` branch). So, although moving the `LANG_CODE.translation` structure into each locale JSON is elegant, this also breaks existing tooling. Is it possible to restore this feature? Or is this change essential to the `I18nTyped` implementation?

danielweck:
I do like the proposed technique which introduces robust "typing" of locale JSONs.

Questions:

* Does the technique implemented in this PR help us identify missing keys in existing `locales/LANG.json` files?
* Same question for redundant/unused keys.

Either way, I would like to adopt the PR, which introduces robustness using a sophisticated technique.
However, this can be combined with the lighweight technique implemented in this PR (which solves both questions above): https://github.com/readium/readium-desktop/pull/516

panaC:
can you remove the *.d.ts linting rules ? or Why vscode isn't incrementing correctly ?

src/common/services/translator.ts
panaC: this change is essential to the I18nTyped implementation, sorry. It's the best practice https://www.i18next.com/principles/translation-resolution. https://github.com/kogai/typed_i18n/issues/66#issuecomment-456303425

danielweck: Ok, we will work around this breaking change, no problems.

danielweck:
So, this new "strong locale typing" feature is a fantastic helper, but in practice this has a prerequisite: that the `en` primary language JSON is clean (no missing keys, no redundant keys). Which is why ; I think ; we need the existing tooling in place, plus the pending PR. The JSON structure refactoring needs to be done first.

danielweck:
Follow-up issue: https://github.com/readium/readium-desktop/issues/514

Issue #492, opened 2019-07-29
Window dimensions not saved on "resize" event (only "move")


danielweck:
https://github.com/readium/readium-desktop/blob/bc3f1337bf50c857bf7a0a48f57e1c076056a697/src/main.ts#L210-L213

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/493

Issue #494, opened 2019-07-29
When detaching reader window with unchanged dimensions / position, library view restored "behind" without obvious visual feedback


danielweck:
... when the reader window dimensions or position is changed, then the library view is restored with the original saved dimensions, so it is usually visible (unless the reader window is larger / overlaps the library view)

So, how could we ensure better visual feedback for this feature?

danielweck:
Re-opening, as the problem persists.

Scarsniik:
Does someone has an idea to do that ? I really have no idea

panaC:
i believe that is fixed

Scarsniik:
Is this issue fixed ? @danielweck

danielweck:
I disagree about closing the issue, but I totally agree about not rushing to a design "fix", but instead think more carefully about the "window management" UX (comparing it with established reading systems on desktop platforms, like MacOS Books.app)

danielweck:
"meta issue" (more general UX discussion):  https://github.com/readium/readium-desktop/issues/271

danielweck:
bug triage: obsolete.

PR #524, opened 2019-08-05
WIP windows position

### LINKED ISSUE #493: Window dimensions for library and reader views are initially the same, but reader view dimension changes are ignored
References:

https://github.com/readium/readium-desktop/blob/bc3f1337bf50c857bf7a0a48f57e1c076056a697/src/main.ts#L88-L91

https://github.com/readium/readium-desktop/blob/bc3f1337bf50c857bf7a0a48f57e1c076056a697/src/main/redux/sagas/reader.ts#L57-L60

The above is okay, but only the library view handles changes in dimensions:

https://github.com/readium/readium-desktop/blob/bc3f1337bf50c857bf7a0a48f57e1c076056a697/src/main.ts#L210-L213

panaC:
- saved windows rectangle into db
 - open new reader windows detached in quincunx

fixes #493 #494 #492

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "start:dev:renderer-reader": "concurrently --kill-others \"npm run start:dev:renderer\" \"npm run start:dev:reader\"",
     "start:dev:renderer": "webpack-dev-server --config webpack.config.renderer.js",
     "start:dev:reader": "webpack-dev-server --config webpack.config.reader.js",
-    "start:dev:main:electron": "cross-env DEBUG=r2:*,readium-desktop:* NODE_ENV=development electron .",
+    "start:dev:main:electron": "cross-env DEBUG=readium-desktop:main:redux:sagas:reader,readium-desktop:common:rectangle NODE_ENV=development electron .",
     "start:dev:main": "npm run build:dev:main && npm run start:dev:main:electron",
     "start:devex": "cross-env WEBPACK=bundle-external npm run start:dev",
     "start:dev": "cross-env DEBUG_COLORS=true concurrently --kill-others \"npm run start:dev:renderer\" \"npm run start:dev:reader\" \"npm run start:dev:main\"",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "start:dev:renderer-reader": "concurrently --kill-others \"npm run start:dev:renderer\" \"npm run start:dev:reader\"",
     "start:dev:renderer": "webpack-dev-server --config webpack.config.renderer.js",
     "start:dev:reader": "webpack-dev-server --config webpack.config.reader.js",
-    "start:dev:main:electron": "cross-env DEBUG=readium-desktop:main:redux:sagas:reader,readium-desktop:common:rectangle NODE_ENV=development electron .",
+    "start:dev:main:electron": "cross-env DEBUG=r2:*,readium-desktop:* NODE_ENV=development electron .",
     "start:dev:main": "npm run build:dev:main && npm run start:dev:main:electron",
     "start:devex": "cross-env WEBPACK=bundle-external npm run start:dev",
     "start:dev": "cross-env DEBUG_COLORS=true concurrently --kill-others \"npm run start:dev:renderer\" \"npm run start:dev:reader\" \"npm run start:dev:main\"",
--- a/src/common/rectangle/window.ts
+++ b/src/common/rectangle/window.ts
@@ -5,32 +5,58 @@
 // that can be found in the LICENSE file exposed on Github (readium) in the project repository.
 // ==LICENSE-END==
 
-import { app, Rectangle, screen } from "electron";
-import * as fs from "fs";
-import * as path from "path";
-import { promisify } from "util";
+import * as debug_ from "debug";
+import { BrowserWindow, Rectangle, screen } from "electron";
+import { ConfigRepository } from "readium-desktop/main/db/repository/config";
+import { container } from "readium-desktop/main/di";
+import { debounce } from "readium-desktop/utils/debounce";
 
-const CONFIG_NAME = "windowRectangle.json";
-const PATH = path.join(app.getPath("userData"), CONFIG_NAME);
+// Logger
+const debug = debug_("readium-desktop:common:rectangle");
 
-export const savedWindowsRectangle = async (rectangle: Rectangle) => {
-    try {
-        await promisify(fs.writeFile)(PATH, JSON.stringify(rectangle));
-    } catch (e) {
-        console.error(e);
-    }
-};
-
-export const getWindowsRectangle = async (): Promise<Rectangle> => {
-    if (await promisify(fs.exists)(PATH)) {
-        return JSON.parse(await promisify(fs.readFile)(PATH, { encoding: "utf8" }));
-    }
-    const c: Rectangle = {
+const configIdKey = "rectangle";
+const defaultRectangle = (): Rectangle => (
+    {
         height: 600,
         width: 800,
         x: screen.getPrimaryDisplay().workAreaSize.width / 3,
         y: screen.getPrimaryDisplay().workAreaSize.height / 3,
-    };
-    savedWindowsRectangle(c);
-    return c;
+    });
+
+type t_savedWindowsRectangle = typeof savedWindowsRectangle;
+export const savedWindowsRectangle = async (rectangle: Rectangle) => {
+    try {
+        const configRepository: ConfigRepository = container.get("config-repository") as ConfigRepository;
+        configRepository.save({
+            identifier: configIdKey,
+            value: rectangle,
+        });
+        debug("new windows rectangle position :", rectangle);
+    } catch (e) {
+        debug("save error", e);
+    }
+    return rectangle;
+};
+
+export const getWindowsRectangle = async (): Promise<Rectangle> => {
+    try {
+        const configRepository: ConfigRepository = container.get("config-repository") as ConfigRepository;
+        const rectangle = await configRepository.get(configIdKey);
+        if (rectangle && rectangle.value) {
+            return rectangle.value;
+        }
+        return await savedWindowsRectangle(defaultRectangle());
+    } catch (e) {
+        debug("get error", e);
+        return defaultRectangle();
+    }
+};
+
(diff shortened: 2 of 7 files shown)


danielweck:
Does this PR address any of the following issues?
* https://github.com/readium/readium-desktop/issues/493
* https://github.com/readium/readium-desktop/issues/494

src/main/redux/sagas/reader.ts
```diff
@@ -147,6 +139,11 @@ async function openReader(publication: Publication, manifestUrl: string) {
         readerWindow.setMenu(null);
     }
 
+    /**
+     * watcher to record windows rectangle position in the db
+     */
+    initRectangleWatcher(readerWindow);
```
danielweck: Naming suggestion: `initWindowRectangleWatcher` ("rectangle" on its own is vague / generic)

danielweck:
This PR fixes https://github.com/readium/readium-desktop/issues/492 , right?

src/main/redux/sagas/reader.ts
panaC: well seen, more clear, thanks

src/common/rectangle/window.ts
```diff
@@ -5,32 +5,58 @@
...
+        x: screen.getPrimaryDisplay().workAreaSize.width / 3,
+        y: screen.getPrimaryDisplay().workAreaSize.height / 3,
+    });
+
+type t_savedWindowsRectangle = typeof savedWindowsRectangle;
 export const savedWindowsRectangle = async (rectangle: Rectangle) => {
     try {
-        await promisify(fs.writeFile)(PATH, JSON.stringify(rectangle));
+        const configRepository: ConfigRepository = container.get("config-repository") as ConfigRepository;
+        configRepository.save({
```
danielweck: missing `await` (`.save()` returns a promise)

src/common/rectangle/window.ts
```diff
@@ -5,32 +5,58 @@
...
+import { ConfigRepository } from "readium-desktop/main/db/repository/config";
+import { container } from "readium-desktop/main/di";
+import { debounce } from "readium-desktop/utils/debounce";
 
-const CONFIG_NAME = "windowRectangle.json";
-const PATH = path.join(app.getPath("userData"), CONFIG_NAME);
+// Logger
+const debug = debug_("readium-desktop:common:rectangle");
 
+const configIdKey = "rectangle";
```
danielweck: Naming suggestion: `windowRectangle` ("rectangle" on its own is vague / generic)

src/common/rectangle/window.ts
```diff
@@ -5,32 +5,58 @@
...
+import * as debug_ from "debug";
+import { BrowserWindow, Rectangle, screen } from "electron";
+import { ConfigRepository } from "readium-desktop/main/db/repository/config";
+import { container } from "readium-desktop/main/di";
+import { debounce } from "readium-desktop/utils/debounce";
 
-const CONFIG_NAME = "windowRectangle.json";
-const PATH = path.join(app.getPath("userData"), CONFIG_NAME);
+// Logger
+const debug = debug_("readium-desktop:common:rectangle");
```
danielweck: Naming suggestion: `readium-desktop:common:windowRectangle` ("rectangle" on its own is vague / generic)

src/common/rectangle/window.ts
```diff
@@ -5,32 +5,58 @@
...
+type t_savedWindowsRectangle = typeof savedWindowsRectangle;
 export const savedWindowsRectangle = async (rectangle: Rectangle) => {
     try {
-        await promisify(fs.writeFile)(PATH, JSON.stringify(rectangle));
+        const configRepository: ConfigRepository = container.get("config-repository") as ConfigRepository;
+        configRepository.save({
+            identifier: configIdKey,
+            value: rectangle,
+        });
+        debug("new windows rectangle position :", rectangle);
```
danielweck: Minor typos: 
1) `windows`=> `window`
2) `position :` => `position:`

package.json
```diff
@@ -23,7 +23,7 @@
     "start:dev:renderer-reader": "concurrently --kill-others \"npm run start:dev:renderer\" \"npm run start:dev:reader\"",
     "start:dev:renderer": "webpack-dev-server --config webpack.config.renderer.js",
     "start:dev:reader": "webpack-dev-server --config webpack.config.reader.js",
-    "start:dev:main:electron": "cross-env DEBUG=r2:*,readium-desktop:* NODE_ENV=development electron .",
+    "start:dev:main:electron": "cross-env DEBUG=readium-desktop:main:redux:sagas:reader,readium-desktop:common:rectangle NODE_ENV=development electron .",
```
danielweck: Is that change intentional? This looks very specific to this PR.

General note: I imagine the use-case for this (local) change was to decrease the default verbosity of `debug()` console output, and narrow it down to PR-specific logging. That's a pretty common use-case for any developer. I wonder how we could set the `DEBUG` environment variable **outside** of `npm run start:dev` (i.e. on the calling side) so that it takes precedence over the `DEBUG` environment variable set by `cross-env` (for example, to remove all `r2:*` console output by using the negative / minus sign `-`). Unfortunately I think this is not possible, because `cross-env` has a higher priority. https://github.com/kentcdodds/cross-env

src/main/services/win-registry.ts
```diff
@@ -50,10 +52,15 @@ export class WinRegistry {
      */
     public registerWindow(win: BrowserWindow, type: AppWindowType): AppWindow {
         const winId = win.id;
-        const appWindow = {
+        const appWindow: AppWindow = {
             identifier: uuid.v4(),
             type,
             win,
+            setBoundsHandler: () => {
```
danielweck: I think this function should include:
```
win.on("move", setBoundsHandler);
win.on("resize", setBoundsHandler);
```
...to avoid the repetitive calls to the event handler registrations.
`setBoundsHandler(win)` then just becomes a local function in `src/main/services/win-registry.ts` with the `win: BrowserWindow` parameter.

danielweck: ...and of course, `removeBoundsHandler()` can be added here too, so that `win.removeListener("move", setBoundsHandler);` and `win.removeListener("resize", readerWindow.setBoundsHandler);` can be called.

src/main/redux/sagas/reader.ts
```diff
@@ -83,7 +74,13 @@ async function openReader(publication: Publication, manifestUrl: string) {
     const readerAppWindow = winRegistry.registerWindow(
         readerWindow,
         AppWindowType.Reader,
-    );
+        );
+
+    // If there are 2 win, record window position in the db
+    if (Object.keys(appWindows).length === 2) {
```
danielweck: I am wondering about the logic here ... shouldn't this be `Object.keys(appWindows).length >= 2`? (sorry if I am missing something obvious :)

src/main/services/win-registry.ts
danielweck: I made a typo in my comments above, I have now updated them, please discard the email notification and read the code comments directly in GitHub. Thanks! :)

package.json
panaC: done

src/main/redux/sagas/reader.ts
panaC: done

src/main/services/win-registry.ts
panaC: `onWindowMoveResize(win)` how do you pass `win` in parameter of event ?

danielweck: `onWindowMoveResize()` can be a local closure where `win` is defined (sorry, I typed my comments in "pseudo code", I didn't go through all the details)

src/main/redux/sagas/reader.ts
danielweck: What do you mean by "done"? The code hasn't changed :)

panaC: it's logic

danielweck: You have to be more explicit, can you please explain the logic behind this code? Question: why not listen to `resize` and `move` events for reader windows when there are more than 2 windows in total (including the library window). Why restrict this to the first reader window only?

src/main/services/win-registry.ts
panaC: done

src/main/redux/sagas/reader.ts
panaC: call me on slack, I will show a demonstration on my computer with screen share

src/main/services/win-registry.ts
danielweck: Nice :)

src/main/redux/sagas/reader.ts
danielweck: Sorry, Slack screenshare only works with the Slack desktop app (I use their web app). Could you please articulate in a couple of sentences why we should only record the bounds of the first reader window, and not the subsequent ones?

panaC: https://slack.com/intl/en-fr/downloads/mac, sorry but it's work for me

panaC: i doesn't speak english very well on github , sorry

danielweck: You're doing fine, don't worry :)

danielweck: Resolving, because UX model / inconsistencies need to be resolved in separate PR (will file dedicated issue)

danielweck:
Note that I get this error in dev / non-prod mode:
```
[2]   readium-desktop:common:rectangle:window save error { [conflict: Document update conflict]
[2]   status: 409,
[2]   name: 'conflict',
[2]   message: 'Document update conflict',
[2]   error: true,
[2]   id: 'config_windowRectangle',
[2]   docId: 'config_windowRectangle' } +71ms
```

...which breaks prod mode completely (I think I have to delete the database from a previous state that was creating during testing the updated WindowsRectangle feature)

danielweck:
Hotfix: https://github.com/readium/readium-desktop/commit/376468474b056785fb620aea12db5cd836853d64

danielweck:
Bug to follow-up: https://github.com/readium/readium-desktop/issues/532

PR #533, opened 2019-08-07
[fix #501] add new i18n key deleteFeed

### LINKED ISSUE #501: OPDS / catalog view: delete confirmation message is wrong
ThoriumDeleteOPDS

panaC:
fix #501


panaC:
ok, but we need to add a complete script for this
"npm run i18n-sort && npm run i18n-typed && ..."

PR #539, opened 2019-08-08
[fix] pass reject boolean into react props when api return a rejection

panaC:
@danielweck as we discussed

Action ERROR handled like SUCCESS allow to pass ERROR string data like a SUCCESS string data and to flag an error boolean into resultIsRejectProp

### DIFF
--- a/src/renderer/redux/reducers/api.ts
+++ b/src/renderer/redux/reducers/api.ts
@@ -20,7 +20,7 @@ export function apiReducer(
     state: ApiState<any> = initialState,
     action: ApiAction,
 ) {
-    const resultIsReject = action.type === ActionType.Error ? true : false;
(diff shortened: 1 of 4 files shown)


src/renderer/redux/reducers/api.ts
```diff
@@ -8,23 +8,29 @@
...
     lastSuccess: null,
 };
 
 // The api reducer.
 export function apiReducer(
-    state: any = initialState,
+    state: ApiState<any> = initialState,
     action: ApiAction,
 ) {
+    const resultIsReject = action.type === ActionType.Error ? true : false;
```
danielweck: Suggestion: `const resultIsReject = action.type === ActionType.Error;`

danielweck:
Look good :)
Let me know if/when you are ready to merge.

Issue #540, opened 2019-08-09
HTTP GET/POST/PUT (LCP/LSD), support for "Problem Details JSON" (RFC7807)


danielweck:
Related issue:
https://github.com/readium/readium-desktop/issues/326

We need to better support schemes such as "Problem Details JSON" (RFC7807) for server "error" responses (to be precise: response bodies associated with non-`2xx` status codes).

The current code obfuscates the response by creating an `Error` object with a syntactically non-parsable string (at least, not reliably parsable).

https://github.com/readium/readium-desktop/blob/04deba11264bd82d9af90aeb58d3f0b2681e6fc1/src/common/utils/http.ts#L58-L61

Suggestion (1): `throw` an Error object with smarter heuristics to build the error payload, maybe modelled on the LCP/LSD `return`/`renew`/`register` principles (copied here):

// Both LSD "renew" and "return" interactions can return errors (i.e. Promise.reject => try/catch with async/await)
// when the server responds with HTTP statusCode < 200 || >= 300.
// The err object in the above code snippet can be a number (HTTP status code) when no response body is available.
// Otherwise, it can be an object with the httpStatusCode property (number) and httpResponseBody (string)
// when the response body cannot be parsed to JSON.
// Otherwise, it can be an object with the httpStatusCode property (number) and other arbitrary JSON properties,
// depending on the server response. Typically, compliant LCP/LSD servers are expected to return Problem Details JSON (RFC7807),
// which provides title type and details JSON properties.
// See https://readium.org/technical/readium-lsd-specification/#31-handling-errors

Note that there should be an externalized reusable helper function (to create the appropriate `throw Error()` payload), so that the same logic can be applied on the calling side of `httpGet()` (`try {} catch() {}` to capture promise rejection due to other errors not handled inside `httpGet()`, such as HTTP timeout).

Therefore, even better suggestion (2): `throw` the actual response body, no pre-processing. Just let the outside `try/catch` (or `promise.then.catch` callback chain) handle the error payload.

llemeurfr:
+1 to reuse the LSD error structure, which is based on RFC7807; from the LSD spec:

To provide a consistent behavior for both clients and end users, all servers must handle errors using the Problem Details JSON object as defined in [RFC7807] with the following additional requirements:
- The server MUST return a title and a type in the Problem Details JSON object
- The server SHOULD attempt to localize both title and detail based on the Accept-Language header sent by the client

danielweck:
[...]

PR #541, opened 2019-08-10
refactor httpGet

panaC:
new version of httpGet


src/common/utils/http.ts
```diff
@@ -53,14 +53,11 @@ export async function httpGet<T extends JsonMap | string = string>(url: string,
...
-    }
-    if (response.statusCode < 200 || response.statusCode >= 300) {
-        const b = response.body ? JSON.stringify(response.body) : "";
-        throw new Error(`HTTP response status code: ${url} => ${response.statusCode} => ${b}`);
-    }
-    if (!response.body) {
-        throw new Error(`HTTP no body?! ${url} => ${response.statusCode}`);
+        throw new Error(`HTTP no response for ${url}`);
+    } else if (response.statusCode < 200 || response.statusCode >= 300) {
+        throw new Error(`HTTP error ${response.statusCode} for ${url}`);
```
danielweck: The response body for HTTP status codes `<200` `>=300` can contain valuable information, so we should not discard it inside the `httpGet()` function. See https://github.com/readium/readium-desktop/issues/540

At a UX level, we can of course decide to display only user-friendly minimal information (e.g. maybe just the HTTP status code), whilst logging the full error message somewhere else (for example a "more info" expandable area in an popup modal dialog box, or some kind of textual application log that users can easily have access to, and copy-paste from, in order to report bugs to the development team, or the publication / content providers).

My suggestion is therefore to `throw` an `Error` with a text message corresponding to the `JSON.stringify()` of a well-defined TypeScript data structure, such as `export interface IHttpGetErrorPaylaod { url: string, httpStatus: number, responseBody?: JsonMap | string, responseType?: string }`. The `responseBody` JSON property is optional and is obtained from the `response.body` (`JsonMap` is useful for properly-formatted returns, like RFC7807 "Problem Details JSON", and `string` accommodates `text/html`, or just `text/plain` ... in other words, no support for `Buffer` responses, which I think a perfectly acceptable in our context). The optional `responseType` JSON field corresponds to `response.headers.get("content-type")`. I think this is all we would need to convey in the error message. Thoughts?

Consequently, on the receiving end / API consumer side (i.e. `try/catch(err)` of the `httpGet()` promise, or `.then().catch(err)`), the code can do: `const errJsonOrString = JSON.parse(err.message)` and check with `typeof errJsonOrString === "string" | "object"`. If `object`, then check for known `IHttpGetErrorPaylaod` data structure (i.e. by introspecting the presence of the JSON properties `errJsonOrString.httpStatus` + `errJsonOrString.responseBody?` + `errJsonOrString.responseType?`), and once again: `const responseJsonOrString = JSON.parse(errJsonOrString.responseBody)`, check with `typeof responseJsonOrString === "string" | "object"` (maybe also cross-check with `errJsonOrString.responseType`, for example `typeof responseJsonOrString === "object"` to match `application/json` prefix (could be `application/json+xxx`)). 

Finally, format the received data as a string suitable for displaying to the end-user (warning: in the case of `typeof responseJsonOrString === "string"` and `errJsonOrString.responseType == "text/html"`, makes sure to sanitize `responseJsonOrString` before inserting into the UI!).

danielweck: I added a copy of this message to https://github.com/readium/readium-desktop/issues/540#issuecomment-520147074

panaC: i know and i totally agree with you to pass error code when HTTP status codes <200 >=300. But be careful, for me stringify an object to pass to `Error` and then parse this object from catch is an awful practice : in first because we lost type inference and then nobody do this ? no ?

It's my opinion.

we have to handle status code in httpGet with a special message to print to user
or return the response object at the end of httpGet and handle the print of status code in api / react.

danielweck: > But be careful, for me stringify an object to pass to Error and then parse this object from catch is an awful practice : in first because we lost type inference and then nobody do this ? no ? It's my opinion.

Sure. What is you counter-proposal? I am very open to alternative suggestions :)

danielweck: PS: from a pragmatic perspective, marshalling / (de)serializing data structures from error/exception payloads is an acceptable practice in a context where promise rejections convey important additional information.
Also see:
https://github.com/readium/readium-desktop/blob/develop/src/common/errors.ts

danielweck: The lack of strong typing in promise rejections is a design limitation in TypeScript, but it also makes sense as a rejection captures expected cases (like in our case the HTTP status codes outside of the resolve range), as well as totally unexpected cases (actual thrown "exceptions").
A possible alternative is to *resolve* the HTTP `request` promise returned from `httpGet()` when HTTP status code is `<200` `>=300`, with a suitable TypeScript type that reflects the duality of the returned value (i.e. string/JSON as well as known error data structure). But this would add the same complexity as my proposal, and would actually also add confusion in the resolve/reject logic, on the client side.

danielweck:
Fixes #540

src/common/utils/http.ts
```diff
@@ -52,15 +71,19 @@ export async function httpGet<T extends JsonMap | string = string>(url: string,
...
-        throw new Error(`No HTTP response?! ${url}`);
-    }
-    if (response.statusCode < 200 || response.statusCode >= 300) {
-        const b = response.body ? JSON.stringify(response.body) : "";
-        throw new Error(`HTTP response status code: ${url} => ${response.statusCode} => ${b}`);
-    }
-    if (!response.body) {
-        throw new Error(`HTTP no body?! ${url} => ${response.statusCode}`);
+    const result = {
+        isFailure: response.statusCode < 200 && response.statusCode >= 300,
```
danielweck: `isFailure` is always false here. Should be `||`, not `&&` logical operator.

src/main/services/lcp.ts
```diff
@@ -164,25 +172,40 @@ export class LcpManager {
...
-        lsdStatus = await this.getLsdStatus(newPublicationDocument);
+        lsdStatus = await this.getLsdStatus(newPublicationDocument.data);
+        if (lsdStatus.isFailure) {
+            throw new Error(`Http getLsdStatus error with code
+                ${lsdStatus.statusCode} for ${lsdStatus.url}`);
+        }
         newPublicationDocument = await this.updateLsdStatus(
             publicationDocument,
-            lsdStatus,
+            lsdStatus.data,
```
danielweck: Due to the changed returned type of `updateLsdStatus()` and `getLsdStatus()`, the `.data` accessor should indeed be added to all instances. However in the current PR, there are missing ones. The program compiles fine because `lsdStatus` is `any`, so TypeScript does not see anything different between `lsdStatus` and `lsdStatus.data`. But the code will crash at runtime due to `lsdStatus` being a wrapper data structure instead of the actual `lsdStatus.data` object.

PR #542, opened 2019-08-10
[fix #419] display error to user when access to an opds feed

### LINKED ISSUE #418: Need notification for invalid OPDS feed
Steps to repeat:
OPDS feed loading error

1. Go to Catalogs
2. Type in a URL. Use an invalid URL or a non OPDS URL, for example www.google.com
3. Name the feed, for example "Google" click ADD
4. Feed name is added as a large Blue Box
5. Click on the feed blue box 
6. The spinning circle continues endlessly. See screen shot.

### LINKED ISSUE #419: Adding or Attempting to Access an OPDS feed while offline results in endless spinning circle
Related to https://github.com/readium/readium-desktop/issues/418

If user is offline attempting to access an OPDS feed results in the endless spinning circle (see screen shot, attached.) 

Similarly, you can add a URL, name the URL, create a blue box "catalog" before getting any feedback while offline. When the user clicks the blue box catalog, the endless spinning circle appears.

I recommend that an error message be added when the user clicks on "Catalogs" if the device is offline. For example:

The Catalogs feature requires an Internet connection. Please try again when your device is online. All books in My Books can be read offline. [Return to My Books] button

OPDS feed loading error

panaC:
Hello @danielweck and @llemeurfr,

Can you tell me what do you think

When an error has occurred, I print to user a title and an error message into BrowserResult

I will add a fix when the user add a new opds feed url : to check before push in db if the feed is good and works (replaced opds:// by http:// also (attramenta)). This fix disallow in most case the bad content-type error message

fix #419 
fix #418


src/main/api/opds.ts
```diff
@@ -66,31 +66,27 @@ export class OpdsApi {
...
-        // Convert to opds version 2
         const xmlDom = new xmldom.DOMParser().parseFromString(opdsFeedData);
-        if (!xmlDom || !xmlDom.documentElement) {
+        if (xmlDom && xmlDom.documentElement) {
+            // This is an opds feed in version 1
+            // Convert to opds version 2
+            const opds1Feed = XML.deserialize<OPDS>(xmlDom, OPDS);
+            opds2Feed = convertOpds1ToOpds2(opds1Feed);
+        } else {
+            const isEntry = xmlDom.documentElement.localName === "entry";
```
danielweck: This `else` branch is `!xmlDom || !documentElement`, which means this line of code is guaranteed to crash. Typo during code copy/paste?

src/main/api/opds.ts
```diff
@@ -66,31 +66,27 @@ export class OpdsApi {
 
     public async browse(data: any): Promise<OpdsResultView> {
         const { url } = data;
-        const opdsFeedData = await httpGet(url);
-        // let opds2Publication: OPDSPublication = null;
+        const opdsFeedData = await httpGet(url, {
```
danielweck: Side question about `const { url } = data;` and `browse(data: any)` ... do you know what the TypeScript type is for `data`?

src/main/api/opds.ts
danielweck: Fixed in https://github.com/readium/readium-desktop/pull/542/files/f57697dda2185e6f99e6e1f71e5badd8649ed2f6..c96d9649800af35454daf776cb15213f6182fc27

src/main/api/opds.ts
danielweck: I am merging this PR without a code change for `any`. This can be resolved at a later stage.

danielweck:
With this branch `panac/fix/display-http-opds-error-to-user`, I am getting an infinite spinner when trying to access `https://www.feedbooks.com/catalog.xml` (the correct URL is `https://www.feedbooks.com/catalog.atom`).

ThoriumOPDSSpinnerInfinite

On the plus side, there is a message in the console:

```
readium-desktop:main#redux/sagas/api Error: HTTP response status code: https://www.feedbooks.com/catalog.xml => 406 => " "
[2]   readium-desktop:main#redux/sagas/api     at Object.<anonymous> (/Volumes/500GB/Code/readium-desktop/dist/main.js:1406:19)
[2]   readium-desktop:main#redux/sagas/api     at Generator.next (<anonymous>)
[2]   readium-desktop:main#redux/sagas/api     at fulfilled (/Volumes/500GB/Code/readium-desktop/node_modules/tslib/tslib.js:107:62)
[2]   readium-desktop:main#redux/sagas/api     at process._tickCallback (internal/process/next_tick.js:68:7) +0ms
```

danielweck:
By the way, I think this PR will fix both:
* https://github.com/readium/readium-desktop/issues/418
* https://github.com/readium/readium-desktop/issues/419
Right?

src/common/utils/http.ts
```diff
@@ -66,21 +67,36 @@ export async function httpGet<TBody extends JsonMap | string = string , TData =
...
-        isFailure: response.statusCode < 200 || response.statusCode >= 300,
-        isSuccess: response.statusCode >= 200 && response.statusCode < 300,
-        url,
-        responseUrl: response.url,
-        statusCode: response.statusCode,
-        body: response.body,
-        data: callback ? undefined : response.body,
-        contentType: response.caseless.get("Content-Type"),
-    };
+    const result = await (async (opt): Promise<IHttpGetResult<TBody, TData>> =>
```
danielweck: Just out of interest: why remove `promisify`? Why not just `try {} catch(err) {}` the `err` (Promise rejection) which is exactly the same as the `err` parameter of the `request` callback? (then, from inside the `catch(err)` body, treat `ETIMEDOUT` in a specific manner, and `throw` any other `err`)

Note that I am "playing devil's advocate" here, because personally I actually prefer to **avoid** using `util.promisify` which; in my opinion; doesn't add value to the perfectly-fine callback-based mechanism of the original `request` API (in this particular case, not generally-speaking).

So I am not asking because I want `util.promisify` to be re-instated, but because I want to better understand the technical rationale for its removal.

Thank you.

src/common/utils/http.ts
```diff
@@ -66,21 +67,36 @@ export async function httpGet<TBody extends JsonMap | string = string , TData =
...
+        new Promise((resolve, reject) => {
+            request(opt, (err, response) => {
+                if (err) {
+                    if (err.code === "ETIMEDOUT") {
+                        resolve({
+                            isTimeout: true,
```
danielweck: Now that the `ETIMEDOUT` error is captured "internally" and transformed into a promise.resolve (it was a promise.reject before, only optionally captured by the caller of `httpGet()`), there is a loss of information, namely the `err.connect` boolean. I therefore suggest that this information be encoded as `timeoutConnect: boolean` in the `IHttpGetResult<TBody, TData>`.
https://github.com/request/request#timeouts

src/common/utils/http.ts
```diff
@@ -66,21 +67,36 @@ export async function httpGet<TBody extends JsonMap | string = string , TData =
...
-        url,
-        responseUrl: response.url,
-        statusCode: response.statusCode,
-        body: response.body,
-        data: callback ? undefined : response.body,
-        contentType: response.caseless.get("Content-Type"),
-    };
+    const result = await (async (opt): Promise<IHttpGetResult<TBody, TData>> =>
+        new Promise((resolve, reject) => {
+            request(opt, (err, response) => {
```
danielweck: Just out of interest: why remove `promisify`? Why not just `try {} catch(err) {}` the `err` (Promise rejection) which is exactly the same as the `err` parameter of the `request` callback? (then, from inside the `catch(err)` body, treat `ETIMEDOUT` in a specific manner, and `throw` any other `err`)

Note that I am "playing devil's advocate" here, because personally I actually prefer to **avoid** using `util.promisify` which; in my opinion; doesn't add value to the perfectly-fine callback-based mechanism of the original `request` API (in this particular case, not generally-speaking).

So I am not asking because I want `util.promisify` to be re-instated, but because I want to better understand the technical rationale for its removal.

Thank you.

src/common/utils/http.ts
danielweck: This "one-liner" code compression makes the source hard to read.
I would suggest preserving the previous coding style: `const promisifiedRequest: Promise<IHttpGetResult<TBody, TData>> = new Promise((resolve, reject) => { ... }); const response = await promisifiedRequest(requestOptions);`, and to call the original `request`  API directly with `requestOptions` (no intermediary `async` function to pass `opt` parameter).

What do you think?

src/common/utils/http.ts
panaC: done

src/common/utils/http.ts
panaC: It's more beautiful for me like this when the err callback have to test before throwing through promisify. I prefer promisify to promise a callback function with (err, data) pattern only.

But in fact, you're right from the beginning

src/common/utils/http.ts
panaC: Is it very useful, this arrow function will not be never called

danielweck: > Is it very useful, this arrow function will never be called

I am not sure I fully understand your reply, so let me try to explain what I mean by "hard to read" code:

Current:
```
const result = await(async (opt): Promise<IHttpGetResult<TBody, TData>> =>
    new Promise((resolve, reject) => {
        request(opt, (err, response) => {
            // ...
        });
    }))(requestOptions);
```

Proposed alternative:
```
const result: IHttpGetResult<TBody, TData> = await new Promise((resolve, reject) => {
    request(requestOptions, (err, response) => {
        // ...
    });
});
```

What do you think?

panaC: ok

src/main/api/opds.ts
```diff
@@ -66,14 +60,22 @@ export class OpdsApi {
...
+        }, async (opdsFeedData) => {
             // let opds2Publication: OPDSPublication = null;
             let opds2Feed: OPDSFeed = null;
 
             if (opdsFeedData.isFailure) {
                 return opdsFeedData;
             }
 
+            debug("opdsFeed content-type", opdsFeedData.contentType);
+            if (!opdsFeedData.contentType.startsWith("application/json") &&
```
danielweck: This string comparison will fail `application/opds+json` (OPDS 2.0)
https://github.com/opds-community/drafts/blob/master/opds-2.0.md

src/main/api/opds.ts
```diff
@@ -66,14 +60,22 @@ export class OpdsApi {
...
+                !opdsFeedData.contentType.startsWith("application/atom+xml")) {
+                throw new Error("bad content-type");
```
danielweck: This will be hard to debug at runtime (especially when troubleshooting remotely, when people send bug reports), so I suggest including at least the URL and the HTTP Content-Type, such as:
`throw new Error(`Not a valid OPDS HTTP Content-Type: ${opdsFeedData.contentType.url} => ${opdsFeedData.contentType}`);`

src/main/api/opds.ts
panaC: I add it.
We need a constant array of all contentType available in opds.

src/main/api/opds.ts
panaC: done

panaC: https://github.com/readium/readium-desktop/pull/542/commits/a5d70e3c7e09d8d5a0e9810224444aed7083095c

src/main/api/opds.ts
panaC: https://github.com/readium/readium-desktop/pull/542/commits/8bbaf722c134b50626397a828dec198bdfb6d6c3

src/resources/locales/en.json
```diff
@@ -111,14 +111,19 @@
...
             "aboutBook": "About this book",
             "addExtract": "Import the extract",
             "addTeaser": "Add the extract to the bookshelf",
             "goBuyBook": "Buy this book",
             "goLoanBook": "Borrow this book",
             "goSubBook": "Subscribe"
+        },
+        "network": {
+            "error": "Unable to access to opds feed",
+            "reject": "An error has occurred",
```
danielweck: Given the catch-all nature of this error message, I think it should be more polite, for example:
"Sorry, there was an unexpected error when trying to process the OPDS feed"

src/resources/locales/en.json
```diff
@@ -111,14 +111,19 @@
...
+            "timeout": "The opds feed takes time to get back, reload the page"
```
danielweck: Capitalize "opds" => `OPDS`

src/renderer/components/opds/BrowserResult.tsx
```diff
@@ -16,82 +16,98 @@ import {
...
+            } else if (result.isTimeout) {
+                content = (
+                    <MessageOpdBrowserResult title={__("opds.network.timeout")} />
+                );
+            } else {
+                content = (
+                    <MessageOpdBrowserResult
+                        title={`${__("opds.network.error")}${result.statusCode} ${result.statusMessage}`}
+                    />
+                );
```
danielweck: `typeof result === "object" && resultIsReject` is not captured by these conditional statements. Is that intentional? (I assume that a promise rejection / unexpected error signaled with `resultIsReject` can originate from a situation where `result` is not a string ... but I must admit not being totally aware of the details under the hood)

src/renderer/components/opds/BrowserResult.tsx
```diff
@@ -16,82 +16,98 @@ import {
...
+                        break;
+                }
+            } else if (result.isTimeout) {
+                content = (
+                    <MessageOpdBrowserResult title={__("opds.network.timeout")} />
+                );
+            } else {
+                content = (
+                    <MessageOpdBrowserResult
+                        title={`${__("opds.network.error")}${result.statusCode} ${result.statusMessage}`}
```
danielweck: There is no space character after the translated `opds.network.error` message (before the HTTP status code).
That being said, I think this is a prime candidate for `i18next` interpolation, such as:

https://github.com/readium/readium-desktop/blob/5a235e99460f162c887c90840473dbc8558f1272/src/resources/locales/en.json#L93-L100

https://www.i18next.com/translation-function/interpolation

src/main/api/opds.ts
```diff
@@ -66,14 +60,24 @@ export class OpdsApi {
...
             let opds2Feed: OPDSFeed = null;
 
             if (opdsFeedData.isFailure) {
                 return opdsFeedData;
             }
 
+            debug("opdsFeed content-type", opdsFeedData.contentType);
+            if (!opdsFeedData.contentType.startsWith("application/json") &&
+                !opdsFeedData.contentType.startsWith("application/opds+json") &&
+                !opdsFeedData.contentType.startsWith("application/atom+xml")) {
```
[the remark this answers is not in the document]
panaC: https://github.com/readium/readium-desktop/pull/542/commits/1d5f0a81451ef73e367b291e3984a6b5ad173c64

src/main/api/opds.ts
```diff
@@ -66,14 +60,24 @@ export class OpdsApi {
...
+        }, async (opdsFeedData) => {
             // let opds2Publication: OPDSPublication = null;
             let opds2Feed: OPDSFeed = null;
 
             if (opdsFeedData.isFailure) {
                 return opdsFeedData;
             }
 
+            debug("opdsFeed content-type", opdsFeedData.contentType);
+            if (!opdsFeedData.contentType.startsWith("application/json") &&
```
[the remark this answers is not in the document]
panaC: I saw that too, normally fixed in 1d5f0a8

src/resources/locales/en.json
panaC: done

src/resources/locales/en.json
panaC: ok thanks

src/resources/locales/en.json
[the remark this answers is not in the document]
panaC: ok thanks

src/resources/locales/fr.json
```diff
@@ -111,14 +111,19 @@
         "addMenu": "Ajouter un flux OPDS",
         "back": "Retour",
         "breadcrumbRoot": "Catalogues",
-        "empty": "",
+        "empty": "Aucune publication trouvées",
```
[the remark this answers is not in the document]
panaC: thanks

src/renderer/components/opds/BrowserResult.tsx
panaC: i handle it now,

src/renderer/components/opds/BrowserResult.tsx
panaC: i changed the text print
I added a message field

src/main/api/opds.ts
```diff
@@ -5,31 +5,41 @@
...
 
 @injectable()
 export class OpdsApi {
+
+    /**
+     * test all possible content-type for both xml and json
+     * @param contentType content-type headers
+     * @returns if content-Type is missing accept
+     */
+    public static contentTypeisAccepted(contentType: string) {
```
danielweck: `contentType` can be undefined, so the parameter's TypeScript type should be `string | undefined`.
Digging deeper: `contentType` is an optional field in `IHttpGetResult<TBody, TData>`, typically `undefined` when `response.caseless.get(...)` does not match anything.
Test case: `node -e "const url = 'https://flaschenpost.piratenpartei.de/catalog/'; require('request')(url, (err, res) => { console.log(res.headers); console.log(res.caseless.get('Content-Type')); });"`

src/resources/locales/en.json
danielweck: You changed `error` instead of `reject`. My comment was attached to the `reject` message, which is not accompanied with additional information (as it is a "catch-all" error situation of unknown origin / reason), unlike `error` which has a well-defined `statusCode` and `statusMessage`.

src/renderer/components/opds/BrowserResult.tsx
danielweck: So just to be clear, based on the current code:

1) if `resultIsReject` is `true` (equivalent to `ApiAction.type === ActionType.Error`, see `apiReducer()` function), and:
2) `result` is non-falsy (i.e. non-null/undefined `ApiAction.payload`), and:
3) the type of `result` is `"object"` (i.e. not `string`), then:
4) `MessageOpdBrowserResult` is invoked with `result.statusCode` / `result.statusMessage` (which can be `undefined` => bug)

If I am not mistaken, the only condition that is protecting us from state (4) is the fact that we can assume that step (3) never yields true in practice, because rejections when `ActionType.Error` usually mean that `ApiAction.payload` has type `string`. But I don't like assumptions like this, especially for the sake of ` future code maintenance.

So, I would like to suggest that `if (resultIsReject)` should be moved **before** `if (typeof result === "object" && result)` to ensure the rejection is handled first, irrespective of the type of `result` (which is why we stringify it, because it could hypothetically be `any`-thing).

(please let me know if you think my concern is unfounded, maybe I am missing something)

src/resources/locales/en.json
panaC: done https://github.com/readium/readium-desktop/pull/542/commits/29a4bbccaacd9759bab1f4b1f18913a8a77e29bf

src/renderer/components/opds/BrowserResult.tsx
panaC: When resultIsReject is true the result props must be a string.

I applied your suggest https://github.com/readium/readium-desktop/pull/542/commits/2b7e93baa8484816e1d8cd024a6b5fdb215b8d3f

src/main/api/opds.ts
panaC: yes i fixed the type https://github.com/readium/readium-desktop/pull/542/commits/eb40109765dec0a3dc4092d277e3c7c42e4770e8

panaC:
> Let me know when this PR is not a WIP anymore, and I will merge. Thanks!

Sorry i haven't surrounded [WIP], so I did not see it

danielweck:
Tested OPDS feed access when offline (`https://www.feedbooks.com/catalog.atom`):

ThoriumOPDSOffline

panaC:
> Tested OPDS feed access when offline (`https://www.feedbooks.com/catalog.atom`):
> 
> ThoriumOPDSOffline

it's the expected behaviour
We can handle this error particularly

danielweck:
* I am getting an infinite spinner when using a bogus URL like `https://www.feedbookzzz.com/catalog.atom`

danielweck:
`https://www.feedbooks.com/catalogzz.atom` =>

ThoriumOPDS500

danielweck:
`https://www.feedbooks.com/catalog.atomzz` =>

ThoriumOPDS406

panaC:
> * https://www.feedbookzzz.com/catalog.atom

the timeout is 10 seconds

panaC:
all of this is expecting behaviour

danielweck:
`http://www.feedbooks.com/publicdomain` =>

ThoriumOPDSfail

danielweck:
The bogus URL (non-existing domain) took an eternity to "respond" (seems more than 10s), but it finally arrived :)

`https://www.feedbookzzz.com/catalog.atom` =>

ThoriumOPDStimeout

panaC:
> https://www.feedbookzzz.com/catalog.atom

Screenshot 2019-08-29 at 19 31 17

src/main/api/opds.ts
```diff
@@ -77,6 +87,12 @@ export class OpdsApi {
                 return opdsFeedData;
             }
 
+            debug("opdsFeed content-type", opdsFeedData.contentType);
+            if (!OpdsApi.contentTypeisAccepted(opdsFeedData.contentType)) {
+                throw new Error(`Not a valid OPDS HTTP Content-Type:
+                    ${opdsFeedData.url} => ${opdsFeedData.contentType}`);
```
danielweck: The line break and other whitespaces are literal (backticks string), resulting in `\n`. See screenshot:

ThoriumOPDSfail

panaC: i saw

danielweck:
Regarding `https://www.feedbookzzz.com/catalog.atom`, my internet provider / gateway handles domain names differently, thus the timeout.

danielweck:
Well-spotted! ( https://github.com/readium/readium-desktop/pull/542/commits/e59954bc70b2c508453ec3f880683b79f43e54a1 )

I was confused by this commit:
https://github.com/readium/readium-desktop/pull/639/commits/bbcd212c89016b8ef2c5b8b30a3baaec61a2cc65

...from this PR:
https://github.com/readium/readium-desktop/pull/639/commits

! :)

src/main/api/opds.ts
panaC: done

I added the timeout, it's was went out

Issue #345, opened 2019-06-20
catalog - add feed text input (URL): should have submit button to accept typed-in tag (currently only press enter)


danielweck:
Similar UX issue to https://github.com/readium/readium-desktop/issues/344

llemeurfr:
This issue is about the addition of an OPDS feed in the Catalog view (not a tag in the Book Info view). 

I agree that the current UX is not great: the user enters a URL, has to select enter, this opens a dialog box, he enters a catalog name and can modify the URL, he selects "Add" (or "Back". 

Let's fully use the dialog box. the new behavior is:
"Add a feed" becomes a button named "Add an OPDS feed" (I think it's good to see the term OPDS somewhere). The input field is suppressed. Selecting this button opens the dialog box. The name and URL are entered in this form. 

Be careful to update the labels accordingly please.

Scarsniik:
Fixede in #550

PR #550, opened 2019-08-12
add a button instead of a form on the opds feed page to add a feed

Scarsniik:
fix issue #345

### DIFF
--- a/src/renderer/assets/styles/opds.css
+++ b/src/renderer/assets/styles/opds.css
@@ -1,31 +1,21 @@
 /* FORM */
 
 .opds_form {
-    & p {
-        display: block;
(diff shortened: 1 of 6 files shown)


danielweck:
The popup modal dialog currently does not return keyboard focus on the calling site (button), when it closes (top-right cross button, escape key, etc.).

However, let's merge this PR anyway, because this is a more global problem that exists in several UX scenarios. This has in fact been partially fixed (on an ad-hoc basis), but we need a more scalable approach by injecting this behaviour on the dialog component itself (note that the "calling site" of some functionality can be an app-global keyboard shortcut, so not necessarily a physical keyboard-focusable button in the UI).

PR #606, opened 2019-08-16
[fix #547] lock-instance

### LINKED ISSUE #547: Ensure single Electron app instance (especially CLI, "open with / double click in file explorer", "drag-drop on app icon", etc.)
`app.requestSingleInstanceLock()`:
https://github.com/electron/electron/blob/master/docs/api/app.md#apprequestsingleinstancelock

`second-instance`:
https://github.com/electron/electron/blob/master/docs/api/app.md#event-second-instance

For example in main process (where the library `BrowserWindow` is created):

https://github.com/readium/readium-desktop/blob/de31c7be16460edffefc10ef71fc857e7dddac4f/src/main.ts#L69-L90

...add something like (code must be adapted for Thorium, this is just a conceptual layout):

```javascript
const singleInstanceLock = app.requestSingleInstanceLock();
if (!singleInstanceLock) {
  app.quit(); // see the `second-instance` event handler below
} else {

function handleFilePath(filepath) {
  app.whenReady().then(() => {

    function openFilePath() {
        // TODO: import or read filepath (as in the current Thorium CLI code)
    }

    function checkWin() { // ensures library BrowserWindow exists (this may be optional in Thorium?)
      if (browserWindow) {
        setTimeout(() => {
          openFilePath();
        }, 200);
      } else {
        setTimeout(() => {
          checkWin();
        }, 600);
      }
    }
    checkWin();
  });
}

function handleArgv(argv) {
  console.log(JSON.stringify(argv));
    if (argv) {
      const args = argv.slice(isDev ? 2 : 1); // TODO: isDev should really be isPackaged (installed app)
      for (let i = 0; i < args.length; i++) {
          // we need to filter out additional `--xxxx` params injected by the Electron / Chromium runtime (notably, on Windows)
          if (args[i] && !args[i].startsWith("--") && fs.existsSync(args[i])) {
            handleFilePath(args[i]);
            break; // just open the first existing file
          }
      }
    }
}
if (process.platform === 'darwin') {
  app.on('will-finish-launching', () => {
    app.on('open-file', (ev, filepath) => {
      ev.preventDefault();
      if (browserWindow) { // this would be the library view
          if (browserWindow.isMinimized()) {
            browserWindow.restore();
          }
          browserWindow.focus();
      }
      handleFilePath(filepath);
    });
  });
}

// Note that the "open-file" event is not raised on MacOS when launching from the CLI, instead the regular Linux/Windows method applies (based on "single instance lock").
  app.on('second-instance', (event, argv, workingDirectory) => {
      if (browserWindow) { // this would be the library view
          if (browserWindow.isMinimized()) {
            browserWindow.restore();
          }
          browserWindow.focus();
      }
    handleArgv(argv);
  });

handleArgv(process.argv);
} // singleInstanceLock
```

panaC:
fix #547 

tested on ubuntu, mac, win


src/main/lock.ts
```diff
@@ -0,0 +1,33 @@
...
+import { WinRegistry } from "readium-desktop/main/services/win-registry";
+
+export function lockInstance() {
+    const gotTheLock = app.requestSingleInstanceLock();
+
+    if (!gotTheLock) {
+        app.quit();
+        return true;
+    } else {
+        app.on("second-instance", () => {
```
danielweck: The `argv` handling is missing.
See my full example in the issue tracker to visualize the darwin / linux+windows flow of actions:
https://github.com/readium/readium-desktop/issues/547

panaC: I thinks it's done, test on your side

src/main/cli/process.ts
```diff
@@ -135,10 +156,23 @@ export function cli(mainFct: () => void) {
...
+        },
+    )
+    .help();
+
+/**
+ * main entry of thorium
+ * @param main main function to exec
+ * @param processArgv process.argv
+ */
+export function cli(_main: () => void, _processArgv = process.argv) {
```
danielweck: Normally, underscore-prefixed function parameters are for explicitly declaring unused local variables, so that the TypeScript linter ignores them.

src/main/cli/process.ts
```diff
@@ -135,10 +156,23 @@ export function cli(mainFct: () => void) {
...
+    mainFct = _main;
+    yargs.parse((_PACKAGING === "0") ? _processArgv.slice(2) : _processArgv.slice(1));
```
danielweck: This will be impacted by https://github.com/readium/readium-desktop/pull/607/files

src/main/lock.ts
```diff
@@ -0,0 +1,44 @@
...
+                }
+                win.win.focus();
+            }
+
+            // execute command line from second instance
+            // when the command line doesn't used electron: execute and exit in second instance process
+            // when the command has needed to open win electron: execute with below cli function
+            // the mainFct is disallow to avoid to generate new mainWindow
+            // remove all -- option arg because isn't handle in ready state app
+            cli(() => ({}), commandLine.filter((arg) => !arg.startsWith("--")));
```
danielweck: This will be impacted by https://github.com/readium/readium-desktop/pull/607/files

src/main/lock.ts
```diff
@@ -0,0 +1,44 @@
...
+import { container } from "readium-desktop/main/di";
+import { WinRegistry } from "readium-desktop/main/services/win-registry";
+
+import { cli } from "./cli/process";
+
+// Logger
+const debug = debug_("readium-desktop:main:lock");
+
+export function lockInstance() {
+    const gotTheLock = app.requestSingleInstanceLock();
```
danielweck: Any attempt to launch the Thorium app more than once must result in the process being "terminated". In other words, an app that does not obtain the "Single Instance Lock" must call `app.quit()` **immediately**. When that happens, the first / original application process receives the `"second-instance"` event (or, in the case of MacOS, the `"open-file" event`).

So, I think we should have code like:
```
const gotTheLock = app.requestSingleInstanceLock();
if (!gotTheLock) {
    app.quit();
} else {
    ...
}

```

src/main/lock.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (gotTheLock) {
+        app.on("second-instance", (_e, commandLine, _workingDir) => {
```
danielweck: The `second-instance` event handling should be side-by-side with the MacOS `open-file` event handling (within `will-finish-launching`), because they are logically-related.

You can see in this comment how I illustrated the different code paths, using `handleFilePath()` and `handleArgv()` (which itself can ultimately invoke `handleFilePath()`, depending on the CLI parameters):
https://github.com/readium/readium-desktop/issues/547#issue-479589133

danielweck: Note that on MacOS, `open-file` is mutually-exclusive with `second-instance`, in the sense that the former is called when:
* double-clicking / activating a file in the Finder with an extension or content-type that is declaratively handled / registered by the app
* drag-and-drop of a file onto the application's icon
* AppleScript launch
...whereas the latter is called only when the application is launched via the command line (not with the `open` shell utility, but directly from the executable)

I tried to illustrate this in my example:
https://github.com/readium/readium-desktop/issues/547#issue-479589133

src/main/cli/process.ts
panaC: done https://github.com/readium/readium-desktop/pull/606/commits/9beff77e9daf58afef8dc2cf8e953e1e1b627b92

panaC:
What I did seems good,

I don't see

I handle the cli on the second-instance, open-file works, and the simple execution of a new instance focus on the main app.

Can you call me

src/main/lock.ts
danielweck: @panaC any comments on the "single instance" design pattern? Is there anything blocking or unclear about it?

Electron documentation:

`app.requestSingleInstanceLock()`:
https://github.com/electron/electron/blob/master/docs/api/app.md#apprequestsingleinstancelock

`second-instance`:
https://github.com/electron/electron/blob/master/docs/api/app.md#event-second-instance

panaC: thank you Daniel,

It's seems good to me

I have tested on mac and win, both of this platforms work

I showed to Laurent: he is agree with that.

I don't understand, why you blocking this PR.

danielweck: Maybe I am missing something obvious, but I hope my code review is correct :)

I already explained the rationale in my previous comment, let me try again:

The Electron design pattern dictates that `app.quit()` (or `app.exit()`) must be called immediately when the "single instance lock" is not obtained, because Electron gives control to the existing/original instance via the `second-instance` event (or `open-file`, in a mutually-exclusive manner).

In the current PR, the code after the call to `const gotTheLock = lockInstance()` continues to execute, initializes Yargs, and inside the Yargs handlers the boolean condition `!gotTheLock` is checked which is nonsensical because an Electron app that does not have a self instance lock  should have aborted immediately.

The solution is quite simple (copy-paste from my original comment above):
```
const gotTheLock = app.requestSingleInstanceLock();
if (!gotTheLock) {
    app.quit();
} else {
    ...
}
```

src/main/lock.ts
danielweck: Just to re-iterate, the `will-finish-launching` + `open-file` MacOS event handler must reside side-by-side (at the code level) with the `second-instance` event handler, because they are logically-related in the Electron app design pattern, which boils down to:

The latter event handler (`second-instance`) is invoked on Windows and Linux for app launches (i.e. CLI, or other types such as "open with" or "double-click on registered/supported file extension from explorer") that fail to acquire a "single instance lock", and only called on MacOS for direct CLI executions (which exclude the `open APP_NAME` or AppleScript app launches). The former (`open-file`) is therefore invoked for all other MacOS app launches that fail to acquire the "single instance lock" (which include the file drag-drop onto the app icon, etc.).

The solution is quite simple, move the event handler from `src/main.ts` to `src/main/lock.ts`:

https://github.com/readium/readium-desktop/blob/dc23718ed6e9e7ccbe69f7790e63b7b21669f63c/src/main.ts#L88-L100

src/main/lock.ts
danielweck: When the code is fixed as per the Electron's documentation, please also remove unnecessary code such as:
```
            // if it's the main instance
            if (gotTheLock) {
```

or:

```
            // if the app is not started or it's the second instance
            if (!app.isReady || !gotTheLock) {
```

etc.

(there is no such thing as a running app that has been refused the "single instance lock", as by definition it must be terminated immediately)

PR #607, opened 2019-08-16
[fix] ignore chromium arg

panaC:
https://github.com/readium/readium-desktop/issues/547#issuecomment-520785137

https://electronjs.org/docs/api/chrome-command-line-switches

### DIFF
--- a/src/main/cli/process.ts
+++ b/src/main/cli/process.ts
@@ -138,7 +138,15 @@ export function cli(mainFct: () => void) {
                     });
                 }
             },
-        )
+    )
         .help()
-        .parse((_PACKAGING === "0") ? process.argv.slice(2) : process.argv.slice(1));
+        .parse(process.argv
+            .filter((arg) => knownOption(arg) || !arg.startsWith("--"))
+            .slice((_PACKAGING === "0") ? 2 : 1));
 }
+
+// arrow function to filter declared option in yargs
+const knownOption = (str: string) => [
+    "--help",
+    "--version",
+].includes(str);
--- a/src/main/cli/process.ts
+++ b/src/main/cli/process.ts
@@ -31,10 +31,10 @@ export function cli(mainFct: () => void) {
                     describe: "title opds feed",
                     type: "string",
                 })
-                    .positional("url", {
-                        describe: "url opds feed",
-                        type: "string",
-                    })
+                .positional("url", {
+                    describe: "url opds feed",
+                    type: "string",
+                })
             ,
             (argv) => {
                 const promise = cliOpds(argv.title, argv.url);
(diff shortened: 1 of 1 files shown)


src/main/cli/process.ts
```diff
@@ -140,5 +140,19 @@ export function cli(mainFct: () => void) {
...
-        .parse((_PACKAGING === "0") ? process.argv.slice(2) : process.argv.slice(1));
+        .parse(cliParse());
+}
+
+function cliParse() {
+    process.argv.shift();
+
+    let findIdx = 0;
+    while (process.argv[findIdx++].startsWith("--")) {
+        process.argv.shift();
```
danielweck: I do not think this code achieves the desired result. Test case:
```
node -e "const arr = ['--first', '--second', 'param1', 'param2', '--third', '--fourth']; let findIdx = 0; while (arr[findIdx++].startsWith('--')) { arr.shift(); } console.log(JSON.stringify(arr));"
```
=>
```
["--second","param1","param2","--third","--fourth"]
```

To make things simple, I think that the Electron/Chromium-injected `argv` parameters (which can occur anywhere in the CLI sequence of parameters, not just at the start) should be preserved, and we should just ignore them in the CLI processing (unrecognised by the Yargs schema => ignore).

panaC: I see, But how to do that ? 

yargs parse : --help and --version
Should we filter only help and version option and ignore all '--' option ?

danielweck: If I am not mistaken, Yargs is just a convenient wrapper around 'argv' which effectively allows us to define (in a declarative manner) a whitelist of parameters, as well as a catch-all fallback procedure for unrecognised params (in our case, we filter-in EPUB paths). So I think we should be able to filter-out any non-recognized params (i.e. outside of the declared whitelist) in Yargs' fallback / catch-all callback?

panaC: I just saw your message, I propose a new fix
What do you think ?

panaC: I tested my code on linux

src/main/cli/process.ts
```diff
@@ -140,5 +140,24 @@ export function cli(mainFct: () => void) {
...
+
+// arrow function to filter known option in CLI
+const knownOption = (str: string) => [
+    "--help",
+    "--version",
+].includes(str);
+
+function cliParse() {
+    // ./electron main.js --my-argument
+    process.argv.shift();
```
danielweck: I would strongly suggest **not** mutating the `process.argv` array.
For example:
```
return process.argv.slice((_PACKAGING === "0") ? 2 : 1).filter((arg) => arg === "--help" || arg === "--version" || !arg.startsWith("--"));
```

danielweck: Actually, I had a look at my old comment ( https://github.com/readium/readium-desktop/issues/547#issuecomment-520785137 ) and it reminded me that parameters can be injected first rather than last!
So, we need to filter first, then slice. Example code:
```
return process.argv.filter((arg) => arg === "--help" || arg === "--version" || !arg.startsWith("--")).slice((_PACKAGING === "0") ? 2 : 1);
```

panaC: done https://github.com/readium/readium-desktop/pull/607/commits/a8f78c2ad87bacdd075de43bee237a8635598ee5

Issue #491, opened 2019-07-29
library view, publication without cover image => long title overflow


danielweck:
![ThoriumCoverImageTitleOverflow](https://user-images.githubusercontent.com/625201/62068041-2412e000-b22d-11e9-9557-847ce59c4f1f.png)

Scarsniik:
fixed by #619

Issue #409, opened 2019-07-11
migrate styles.ts and inline styles to CSS stylesheets ?


danielweck:
https://github.com/readium/readium-desktop/blob/develop/src/renderer/components/styles.ts

vs.

https://github.com/readium/readium-desktop/tree/develop/src/renderer/assets/styles

danielweck:
Related: https://github.com/readium/readium-desktop/issues/410

danielweck:
There is also inline CSS, for example `backgroundColor` here:
https://github.com/readium/readium-desktop/blob/3bab0fd6d84ce1006e155a2885f1d93259586240/src/renderer/components/reader/ReaderHeader.tsx#L84-L91
https://github.com/readium/readium-desktop/blob/3bab0fd6d84ce1006e155a2885f1d93259586240/src/renderer/components/reader/ReaderHeader.tsx#L134-L137

PR #619, opened 2019-08-19
fix the fake cover overflow for long title or author list

Scarsniik:


Scarsniik:
fix #491

src/renderer/components/publication/Cover.tsx
```diff
@@ -27,15 +27,6 @@ export default class Cover extends React.Component<ICoverProps, null> {
 
         if (this.props.publication.cover == null) {
             let authors = "";
-            const bodyCSS = Object.assign({}, Styles.BookCover.body);
-            let colors = this.props.publication.customCover;
```
danielweck: If I am not mistaken, in this PR `publication.customCover` becomes an unused feature.
I suggest to:
1) Either - reinstate support for `publication.customCover`.
2) Or - eliminate dead code, such as (but not limited to):
https://github.com/readium/readium-desktop/blob/cd3997f902be46d4299dc97e0f3ccbdd09fbe69c/src/common/models/custom-cover.ts#L8-L26
https://github.com/readium/readium-desktop/blob/cd3997f902be46d4299dc97e0f3ccbdd09fbe69c/src/main/services/catalog.ts#L332-L339

The latter option has repercussions on the database schema. I do not know the technical details well-enough to assess whether or not this would introduce risks of regression bugs (i.e. handling existing database storage with different schema / parsing rules).

src/renderer/components/styles.ts
```diff
@@ -1,298 +0,0 @@
...
-import { relative } from "path";
-
-const auto: "auto" = "auto";
-const fixed: "fixed" = "fixed";
-const hidden: "hidden" = "hidden";
-const absolute: "absolute" = "absolute";
-const column: "column" = "column";
-const wrap: "wrap" = "wrap";
-
-export const Styles = {
```
danielweck: Note that the removal of `src/renderer/components/styles.ts` partially addresses issue: https://github.com/readium/readium-desktop/issues/409
(we might want to file a separate issue to address "inline styles")

src/renderer/assets/styles/publication.css
```diff
@@ -164,3 +164,39 @@
...
     }
 }
+
+.cover {
+    height: 100%;
+    width: 100%;
+    text-align: center;
+    box-sizing: border-box;
+    padding: 5%;
+    background-image: linear-gradient(#d18e4b, #7c4c1c);
```
danielweck: Suggestion: use CSS variables in `background-image: linear-gradient(#d18e4b, #7c4c1c);`, so that randomly-generated colour schemes can be used (see my comment about `publication.customCover` deprecation).

src/renderer/components/publication/Cover.tsx
```diff
@@ -44,16 +35,24 @@ export default class Cover extends React.Component<ICoverProps, null> {
                 }
                 authors += this.translator.translateContentField(newAuthor);
             }
+            let colors = this.props.publication.customCover;
+            if (colors === undefined) {
+                colors = {
+                    topColor: "#d18e4b",
+                    bottomColor: "#7c4c1c",
+                };
```
danielweck: Instead of hard-coding the CSS colour strings here, would it be possible to reference them from the existing `common/models/custom-cover.ts` file?
https://github.com/readium/readium-desktop/blob/cd3997f902be46d4299dc97e0f3ccbdd09fbe69c/src/common/models/custom-cover.ts#L13-L26

Once centralized, these hard-coded CSS styles will be easier to refactor / extract into proper CSS variables (or any other applicable technique), see related issue https://github.com/readium/readium-desktop/issues/409

llemeurfr:
Just a comment: as we need to publish our V1.0 very soon, it is important to separate enhancements from fixes. If making the CSS more flexible take more than 5 mn, I prefer having a new issue open in the backlog.

danielweck:
Noted, Laurent.

src/renderer/components/publication/Cover.tsx
```diff
@@ -44,16 +37,21 @@ export default class Cover extends React.Component<ICoverProps, null> {
                 }
                 authors += this.translator.translateContentField(newAuthor);
             }
+            let colors = this.props.publication.customCover;
+            if (colors === undefined) {
```
danielweck: I will merge this PR as-is, because at the moment the following bug does not exist: but, there is a potential fatal bug here (crash) when `this.props.publication.customCover` is `null` (because `colors === undefined` will not match), resulting in `colors` remaining `null`, and the rest of the code crashing. This could be detected if `strictNullChecks` was enabled in the TypeScript compiler settings (see my WIP PR), but unfortunately this would currently break a huge amount of Thorium code, so we have to accept some degree of uncertainty (no static checks), and/or rely on unit-tests, and programmer's awareness (thus why I am bringing this up here).

danielweck: I created a separate issue to track this project-wide problem. https://github.com/readium/readium-desktop/issues/634

PR #620, opened 2019-08-19
fix opds search breadcrumb

### LINKED ISSUE #616: OPDS search, simplify breadcrumbs
At each level (section) of a feed we are likely to find a search link. In Feedbooks and for most other providers, this is always a "general" search (its context is not the section in which the search URL is found). 

The results lists are currently presented in Thorium with breadcrumbs like "Catalogs/<feed&gt;/Best Sellers/<term searched&gt;"  as if the search was contextual.

Please remove the context, so that breadcrumbs are "Catalogs/<feed&gt;/<term searched&gt;".

Scarsniik:
fix #616


danielweck:
Tested, seems to work fine. Thanks Leo! :)

PR #633, opened 2019-08-20
remove redundant description

### LINKED ISSUE #586: Library view / bookshelf, cover images: not alt="" role="presentation" (or some other text description, like title? ... maybe redundant with other labels though)

Scarsniik:
Fixes #586

### DIFF
--- a/src/resources/locales/en.json
+++ b/src/resources/locales/en.json
@@ -1,13 +1,13 @@
 {
     "accessibility": {
-        "closeDialog": "Close",
         "bookMenu": "Menu",
+        "closeDialog": "Close",
         "homeMenu": "Reading view menu",
         "importFile": "Import a publication",
-        "searchBook": "Search by title",
-        "skipLink": "Skip to content",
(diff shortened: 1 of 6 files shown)


src/resources/locales/en.json
```diff
@@ -1,6 +1,7 @@
 {
     "accessibility": {
         "closeDialog": "Close",
+        "bookMenu": "Menu",
```
danielweck: Could you please run the command `npm run i18n-sort` followed by `npm run i18n-typed` so that the locale JSON is in "canonical" form (i.e. all language translations sorted by key alphabetical order, same indentation + trailing line break, etc.)
PS: you only need `npm run i18n-scan` if you need to check that the `i18next` keys used in the source code are all present in the JSONs, and that there are no superfluous / unused keys.
You only need `npm run i18n-check` if you need to compare the "primary" english locale with the "secondary" ones (missing keys are automatically added with an empty string value, redundant keys are removed).

Scarsniik: It's fixed

PR #637, opened 2019-08-21
add padding in some button for better focus display

### LINKED ISSUE #573: Get some padding in focused labels (in focus rings)
On some labels, the first and last letter is “cut’ by the focus ring (i.e. blue line around the focused item). This is especially the case for “skip to context” (before it is moved, which is another open issue). It is NOT the case for labels already in “boxes” (e.g. All Books”).  

Outline-offset seems to be the proper CSS trick to use when needed. 
https://css-tricks.com/almanac/properties/o/outline-offset/

Scarsniik:
fix #573

### DIFF
--- a/src/renderer/components/reader/ReaderMenu.tsx
+++ b/src/renderer/components/reader/ReaderMenu.tsx
@@ -103,7 +103,6 @@ export class ReaderMenu extends React.Component<Props, State> {
     }
 
     private createTOCRenderList(TOC: any[]): JSX.Element {
-        console.log(styles);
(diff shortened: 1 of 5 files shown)


src/renderer/components/reader/ReaderMenu.tsx
```diff
@@ -103,6 +103,7 @@ export class ReaderMenu extends React.Component<Props, State> {
     }
 
     private createTOCRenderList(TOC: any[]): JSX.Element {
+        console.log(styles);
```
danielweck: Please remove this :)

PR #639, opened 2019-08-26
[fix] relative url to absolute url in converter opds

panaC:
replace opds protocol in url with http

add small typing change

tested on mac

### DIFF
--- a/src/main/api/opds.ts
+++ b/src/main/api/opds.ts
@@ -69,9 +69,7 @@ export class OpdsApi {
         if (new URL(url).protocol === "opds:") {
             url = url.replace("opds://", "http://");
         }
-        return await httpGet(url, {
-            timeout: 10000,
-        }, async (opdsFeedData) => {
+        return await httpGet(url, {}, async (opdsFeedData) => {
             // let opds2Publication: OPDSPublication = null;
             let opds2Feed: OPDSFeed = null;
 
--- a/src/main/api/opds.ts
+++ b/src/main/api/opds.ts
@@ -66,7 +66,7 @@ export class OpdsApi {
 
     public async browse(data: any): Promise<THttpGetOpdsResultView> {
         let url: string = data.url;
-        if (new URL(url).protocol === "opds") {
+        if (new URL(url).protocol === "opds:") {
             url = url.replace("opds://", "http://");
         }
         return await httpGet(url, {
--- a/src/main/api/opds.ts
+++ b/src/main/api/opds.ts
@@ -67,7 +67,7 @@ export class OpdsApi {
     public async browse(data: any): Promise<THttpGetOpdsResultView> {
         let url: string = data.url;
         if (new URL(url).protocol === "opds") {
-            url = url.replace(/[a-zA-Z]*)\:\/\//, "http");
+            url = url.replace(/([a-zA-Z]*)\:\/\//, "http");
         }
         return await httpGet(url, {
             timeout: 10000,
--- a/src/main/api/opds.ts
+++ b/src/main/api/opds.ts
@@ -65,8 +65,13 @@ export class OpdsApi {
     }
 
     public async browse(data: any): Promise<THttpGetOpdsResultView> {
-        const { url } = data;
-        return await httpGet(url, {}, async (opdsFeedData) => {
+        let url: string = data.url;
+        if (new URL(url).protocol === "opds") {
+            url = url.replace(/[a-zA-Z]*)\:\/\//, "http");
+        }
+        return await httpGet(url, {
+            timeout: 10000,
+        }, async (opdsFeedData) => {
             // let opds2Publication: OPDSPublication = null;
             let opds2Feed: OPDSFeed = null;
 
@@ -92,7 +97,7 @@ export class OpdsApi {
                     OPDSFeed,
                 );
             }
-            opdsFeedData.data = await this.opdsFeedViewConverter.convertOpdsFeedToView(opds2Feed);
+            opdsFeedData.data = await this.opdsFeedViewConverter.convertOpdsFeedToView(opds2Feed, url);
             return opdsFeedData;
         });
     }
--- a/src/main/api/opds.ts
+++ b/src/main/api/opds.ts
@@ -67,7 +67,7 @@ export class OpdsApi {
     public async browse(data: any): Promise<THttpGetOpdsResultView> {
         let url: string = data.url;
         if (new URL(url).protocol === "opds") {
-            url = url.replace(/([a-zA-Z]*)\:\/\//, "http");
+            url = url.replace("opds://", "http://");
         }
         return await httpGet(url, {
             timeout: 10000,
--- a/src/main/converter/opds.ts
+++ b/src/main/converter/opds.ts
@@ -179,7 +179,7 @@ export class OpdsFeedViewConverter {
 
             // concatenate all relative path to an absolute URL path
             navigation = navigation.map((nav) => {
-                if (nav.url && !/[a-zA-Z]+\:\/\//.exec(nav.url)) {
+                if (nav.url && !/ˆ[a-zA-Z]+:\/\//.exec(nav.url)) {
                     nav.url = `${url}${nav.url}`;
                 }
                 return nav;
--- a/src/main/converter/opds.ts
+++ b/src/main/converter/opds.ts
@@ -158,11 +158,11 @@ export class OpdsFeedViewConverter {
         };
     }
 
-    public async convertOpdsFeedToView(feed: OPDSFeed): Promise<OpdsResultView> {
+    public async convertOpdsFeedToView(feed: OPDSFeed, url: string): Promise<OpdsResultView> {
         const title = convertMultiLangStringToString(feed.Metadata.Title);
         let type = OpdsResultType.Empty;
-        let navigation = null;
-        let publications = null;
+        let navigation: OpdsLinkView[] | undefined;
+        let publications: OpdsPublicationView[] | undefined;
 
         if (feed.Publications) {
             // result page containing publications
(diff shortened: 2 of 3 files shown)


src/main/api/opds.ts
```diff
@@ -65,8 +65,13 @@ export class OpdsApi {
     }
 
     public async browse(data: any): Promise<THttpGetOpdsResultView> {
-        const { url } = data;
-        return await httpGet(url, {}, async (opdsFeedData) => {
+        let url: string = data.url;
+        if (new URL(url).protocol === "opds") {
+            url = url.replace(/([a-zA-Z]*)\:\/\//, "http");
```
danielweck: This regular expression does not seem to work. What URL did you successfully test it with?

To reproduce the bug, copy-paste this Javascript code snippet in your web browser console:
`'opds://127.0.0.1:8080/test.atom'.replace(/([a-zA-Z]*)\:\/\//, "http")`
=>
`"http127.0.0.1:8080/test.atom"`

Several issues:
1) The capturing group `([a-zA-Z]*)` is not consumed anywhere in the replacement pattern (i.e. `$1`).
2) The sequence `\:` escapes a non-reserved character. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
3) The `[a-zA-Z]*` quantifier should be `+`, not `*`.
4) The `[a-zA-Z]` character definition does not match the URL scheme syntax https://tools.ietf.org/html/rfc3986?#section-3.1 (`ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`)
5) To be precise, the boundary "beginning" indicator `^` should be specified in the `RegExp`, before the URL scheme matcher (caveat: `url` *must* be trimmed from whitespace characters on the leading edge).

To conclude, because the condition `new URL(url).protocol === "opds"` is `true`, personally I would avoid using regular expressions, and I would simply replace `opds://` with `http://`, e.g. `'opds://127.0.0.1:8080/test.atom'.replace("opds://", "http://")` => `http://127.0.0.1:8080/test.atom`

panaC: many thanks Daniel for your explanation

I fixed it

src/main/converter/opds.ts
```diff
@@ -176,6 +176,14 @@ export class OpdsFeedViewConverter {
             navigation = feed.Navigation.map((item) => {
                 return this.convertOpdsLinkToView(item);
             });
+
+            // concatenate all relative path to an absolute URL path
+            navigation = navigation.map((nav) => {
+                if (nav.url && !/[a-zA-Z]+\:\/\//.exec(nav.url)) {
+                    nav.url = `${url}${nav.url}`;
```
danielweck: This string concatenation will fail in cases where the "base" URL of the OPDS feed contains query parameters, e.g.:
* `url` = "https://domain.com/catalog?format=opds&lang=fr"
* `nav.url` = "category/fiction/latest"
* `${url}${nav.url}` => "https://domain.com/catalog?format=opds&lang=frcategory/fiction/latest"

danielweck: This string concatenation technique will fail in cases where the "relative" URL (i.e. `nav.url`) is in fact an absolute path reference, for example:
* `url` = `https://domain.com/catalog.opds`
* `nav.url` = `/catalog/best-sellers.atom`
* `${url}${nav.url}` => `https://domain.com/catalog.opds/catalog/best-sellers.atom`

src/main/api/opds.ts
```diff
@@ -65,8 +65,13 @@ export class OpdsApi {
     }
 
     public async browse(data: any): Promise<THttpGetOpdsResultView> {
-        const { url } = data;
-        return await httpGet(url, {}, async (opdsFeedData) => {
+        let url: string = data.url;
+        if (new URL(url).protocol === "opds") {
```
danielweck: This code will always fail, because the `URL` `.protocol` accessor returns the `:` suffix. 

NodeJS API documentation: https://nodejs.org/api/url.html#url_url_protocol

Reduced repro test case: `node -e "console.log(new URL('https://domain.com').protocol)"` => `https:`

danielweck: Note that this is true for the legacy `URL` API as well:
`node -e "console.log(require('url').parse('https://domain.com').protocol)"`

src/main/converter/opds.ts
danielweck: The current URL normalization / absolutization routine does not correctly handle "protocol-relative URLs", for example:
* `url` = `https://domain.com/catalog.opds`
* `nav.url` = `//domain.com/catalog/best-sellers.atom`
* `${url}${nav.url}` => `https://domain.com/catalog.opds//domain.com/catalog/best-sellers.atom`

(this is probably not a common case, but I do remember coming across an OPDS feed that contained such "relative" URLs)

danielweck: The proposed string concatenation technique will fail in the following case:

* `url` = `https://domain.com/catalog/main.opds`
* `nav.url` = `category/fiction/latest.atom` (same problem with `./category/fiction/latest.atom`)
* `${url}${nav.url}` => `https://domain.com/catalog/main.opdscategory/fiction/latest.atom`

(the resulting absolutized URL should be `https://domain.com/catalog/category/fiction/latest.atom`)

danielweck: The proposed string concatenation technique will also fail in the following case:

* `url` = `https://domain.com/main/catalog.opds`
* `nav.url` = `../category/fiction/latest.atom`
* `${url}${nav.url}` => `https://domain.com/main/catalog.opds../category/fiction/latest.atom`

(the resulting absolutized URL should be `https://domain.com/category/fiction/latest.atom`)

src/main/converter/opds.ts
```diff
@@ -176,6 +176,14 @@ export class OpdsFeedViewConverter {
             navigation = feed.Navigation.map((item) => {
                 return this.convertOpdsLinkToView(item);
             });
+
+            // concatenate all relative path to an absolute URL path
+            navigation = navigation.map((nav) => {
+                if (nav.url && !/[a-zA-Z]+\:\/\//.exec(nav.url)) {
```
danielweck: The sequence `\:` escapes a non-reserved character. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions

danielweck: The lack of boundary "beginning" indicator `^` in the `RegExp` (i.e. before the pattern matcher for the URL scheme / protocol) means that there is a possibility of false positives, for instance:

`node -e "const url = './catalog/search://test'; console.log(/[a-zA-Z]+:\/\//.exec(url));"`
=>
```
[ 'search://',
  index: 10,
  input: './catalog/search://test',
  groups: undefined ]
```

versus (just added `^`):

`node -e "const url = './catalog/search://test'; console.log(/^[a-zA-Z]+:\/\//.exec(url));"`
=>
`null` (correct expected result)

(note: `^` means that the input `url` string *must* be trimmed from whitespace characters on the leading edge!)

danielweck: The `[a-zA-Z]+` RegExp pattern does not match the URL scheme syntax https://tools.ietf.org/html/rfc3986?#section-3.1 (`ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`)

In our use-case we only expect `http://` or `https://` (in my tests I have personally never come across anything else, not even `opds://` inside OPDS feeds), so instead of implementing the correct URL protocol syntax / RFC parsing rules, we could just match with `!/^https?:\/\//.exec(nav.url)`.

What do you think?

src/main/api/opds.ts
panaC: done https://github.com/readium/readium-desktop/pull/639/commits/0872fd8d706098f3a11553ef9c4f8184e8daa027

src/main/converter/opds.ts
panaC: thanks Daniel
Done https://github.com/readium/readium-desktop/pull/639/commits/bd5fb1ef779b422b850121d3a44ad2600d7d347f

panaC: sure

The opds:// is for attramenta opds feed (opds://www.atramenta.net/opds/catalog.atom)

but it doesn't appear inside opds feed

done https://github.com/readium/readium-desktop/pull/639/commits/cac1e6acdd7568b7a506c1aab8ffff29d8301e51

src/main/converter/opds.ts
panaC: thanks daniel
I used url.resolve instead string concatenation

fixed https://github.com/readium/readium-desktop/pull/639/commits/8072856b3b320d8b3b3de70c1696580ee5eabaa2

danielweck: Passes the tests okay.

`node -e "const url = require('url'); const a = 'https://domain.com/catalog?format=opds&lang=fr'; const b = 'category/fiction/latest?param=test'; console.log(url.resolve(a, b));"`
=>
`https://domain.com/category/fiction/latest?param=test`

`node -e "const url = require('url'); const a = 'https://domain.com/catalog.opds'; const b = '/catalog/best-sellers.atom'; console.log(url.resolve(a, b));"`
=>
`https://domain.com/catalog/best-sellers.atom`

`node -e "const url = require('url'); const a = 'https://domain.com/catalog.opds'; const b = '//domain.com/catalog/best-sellers.atom'; console.log(url.resolve(a, b));"`
=>
`https://domain.com/catalog/best-sellers.atom`

`node -e "const url = require('url'); const a = 'https://domain.com/catalog/main.opds'; const b = 'category/fiction/latest.atom'; console.log(url.resolve(a, b));"`
=>
`https://domain.com/catalog/category/fiction/latest.atom`

`node -e "const url = require('url'); const a = 'https://domain.com/main/catalog.opds'; const b = '../category/fiction/latest.atom'; console.log(url.resolve(a, b));"`
=>
`https://domain.com/category/fiction/latest.atom`

PR #644, opened 2019-08-28
MathML support via MathJax (for advanced users, disabled by default)

### LINKED ISSUE #643: Support MathML via MathJax

danielweck:
Fixes #643

Sample EPUB3 publication "linear algebra":
https://idpf.github.io/epub3-samples/30/samples.html#linear-algebra

"Before" screenshot:

ThoriumMathML

"After" screenshots:

ThoriumMathJaxMenu
ThoriumMathJaxAbout


danielweck:
Current notable caveats:
* By default, MathJax loads incrementally / in an async manner (which is good, because otherwise there would be a long / blocking rendering operation), so the "resize sensor" in `r2-navigator-js` is triggered many times, and the "reading location" event is emitted many times as well due to detected changes.
* MathJax's popup menu is useful (and keyboard-accessible too), but it also interferes with document rendering (changes the viewport overflow, for example)
* All MathJax-injected DOM markup **must** be skipped during selection/annotations (not a problem right now in Thorium, as not implemented yet), TTS / read-aloud with text-to-speech (again, not an immediate concern in Thorium), and more crucially in the "reading location" computations => we need a blacklisting mechanism for CSS classes and/or element IDs.

danielweck:
Note that the chosen MathJax rendering method in this PR is the default CommonHTML one (not SVG, which was used in Readium1 due to technical limitations, although this was also a conscious design decision due to the opportunity to serve a single giant file `MathJax.js` which contained the SVG glyphs / paths instead of relying on external fonts)

danielweck:
Overall, I think that MathJax should be an opt-in feature, which users could easily deactivate in case problems arise from it.

Issue #538, opened 2019-08-08
Disable the possibility to download a book multiple times from an OPDS feed


llemeurfr:
Steps: 
- open an OPDS feed and select a free book. 
- Select "Add book to library" ->  a toast message appears
- Immediately select "Add" again -> a new toast message appears. 

Once the action has been taken into account, a second action should be blocked until the end of the import. 

After the first import has been achieved, if the user selects a new import of the same book (and this is a generic rule) a dialog box should appear: "This book has already been imported. Do you want to download it again? Yes/No".

If it is complex to record that an import is active and not finished (transient state), the second rule should at least apply (consider that the book is in the DB as soon as its import has begun). 

See also #289 where this issue arised first.

Scarsniik:
There is already the second rule
![image](https://user-images.githubusercontent.com/15014810/62944131-564f4080-bddc-11e9-8748-8390dc8f8c23.png)

Or do you think about something else ?

llemeurfr:
Pas pour la première partie de la demande : pour l'instant rien n'empêche un second téléchargement PENDANT le premier (via un second clic dans la seconde). 

Mais en effet une fois le 1er chargement achevé, la modale actuelle est parfaite, je ne l'avais pas repéré.

Scarsniik:
fixed by #646

PR #646, opened 2019-08-29
add a check for duplicate import in current downloads

Scarsniik:
Fixes issue #538

### DIFF
--- a/src/common/redux/states/import.ts
+++ b/src/common/redux/states/import.ts
@@ -6,6 +6,13 @@
 // ==LICENSE-END==
 
 export interface ImportState {
-    publication: any;
+    publication: ImportOpdsPublication;
     downloadSample: boolean;
+}
+
+export interface ImportOpdsPublication {
+    url: string;
+    base64OpdsPublication: string;
+    downloadSample: boolean;
+    title: string;
 }
--- a/src/common/redux/states/import.ts
+++ b/src/common/redux/states/import.ts
@@ -13,6 +13,5 @@ export interface ImportState {
 export interface ImportOpdsPublication {
     url: string;
     base64OpdsPublication: string;
-    downloadSample: boolean;
     title: string;
 }
--- a/src/renderer/components/utils/SameFileImportManager.tsx
+++ b/src/renderer/components/utils/SameFileImportManager.tsx
@@ -25,34 +25,50 @@ interface Props  {
     search?: any;
     searchResult?: any;
     importOpdsEntry?: any;
+    downloads?: any[];
 }
 
 class SameFileImportManager extends React.Component<Props> {
     public componentDidUpdate(oldProps: Props) {
-        const { searchResult, lastImport } = this.props;
+        const { searchResult, lastImport, downloads } = this.props;
 
         if (searchResult !== oldProps.searchResult) {
             if (searchResult.length === 0) {
-                this.props.importOpdsEntry(
-                    {
-                        url: lastImport.publication.url,
-                        base64OpdsPublication: lastImport.publication.base64OpdsPublication,
-                        downloadSample: lastImport.downloadSample,
-                        title: lastImport.publication.title,
-                    },
-                );
+                this.importOpds();
             } else {
                 this.props.displayImportDialog(lastImport.publication, lastImport.downloadSample );
             }
         }
 
         if (lastImport !== oldProps.lastImport) {
-            this.props.search({text: lastImport.publication.title});
+            const foundInCurrentDownload = downloads.findIndex(
+                (value) => {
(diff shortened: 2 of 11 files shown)


src/renderer/components/utils/SameFileImportManager.tsx
```diff
@@ -25,35 +25,51 @@ interface Props  {
...
             } else {
                 this.props.displayImportDialog(lastImport.publication, lastImport.downloadSample );
             }
         }
 
         if (lastImport !== oldProps.lastImport) {
-            this.props.search({text: lastImport.publication.title});
+            const foundInCurrentDownload = downloads.findIndex(
+                (value) => {
+                    return value.publicationTitle === lastImport.publication.title;
```
danielweck: The type of `lastImport` is `ImportState`:
https://github.com/readium/readium-desktop/blob/develop/src/common/redux/states/import.ts
```
export interface ImportState {
    publication: any;
    downloadSample: boolean;
}
```

...but the type of `publication` is `any`...so by reverse-engineering the type system I was able to find this:
```
{
url: string;
base64OpdsPublication: string;
title: string;
}
```
(there may be more properties, I am not sure)

Could you please help me (and other contributors) by replacing `any` with the actual type of `publication`? Many thanks!

src/renderer/components/utils/SameFileImportManager.tsx
```diff
@@ -18,42 +18,58 @@ import { OpdsPublicationView } from "readium-desktop/common/views/opds";
...
 import { ImportState } from "readium-desktop/common/redux/states/import";
+import { Download } from "readium-desktop/renderer/redux/states/download";
 
 interface Props  {
     lastImport?: ImportState;
     displayImportDialog?: any;
     search?: any;
     searchResult?: any;
     importOpdsEntry?: any;
+    downloads?: Download[];
```
danielweck: Is the type of the `importOpdsEntry` Prop:
`(data: ImportOpdsPublication) => Promise<PublicationView>`
?
If so, would it be possible to replace `any`?

src/common/redux/states/import.ts
```diff
@@ -6,6 +6,13 @@
...
 export interface ImportState {
-    publication: any;
+    publication: ImportOpdsPublication;
     downloadSample: boolean;
 }
+
+export interface ImportOpdsPublication {
+    url: string;
+    base64OpdsPublication: string;
+    downloadSample: boolean;
```
danielweck: The `downloadSample` boolean member is defined twice (in the `ImportState` and `ImportOpdsPublication` interfaces). It is indeed perfectly logical for `lastImport: ImportState` to convey the `downloadSample` true/false property. However I think we should remove the `downloadSample` member from the `ImportOpdsPublication` data structure, and change the signature of the `importOpdsEntry()` function accordingly:
`importOpdsEntry(data: ImportOpdsPublication, downloadSample: boolean): Promise<PublicationView>`

The alternative (which I personally do not like) would be to remove the `downloadSample` member from the `ImportOpdsPublication` data structure, and to create an extended type (interface) specifically for the parameter of the `importOpdsEntry()` function:
```
export interface ImportOpdsPublicationWithDownloadSample extends ImportOpdsPublication {
    downloadSample: boolean;
}
```

src/renderer/components/utils/SameFileImportManager.tsx
```diff
@@ -17,43 +17,52 @@ import { OpdsPublicationView } from "readium-desktop/common/views/opds";
...
             } else {
                 this.props.displayImportDialog(lastImport.publication, lastImport.downloadSample );
             }
         }
 
         if (lastImport !== oldProps.lastImport) {
-            this.props.search({text: lastImport.publication.title});
+            const foundInCurrentDownload = downloads.findIndex(
+                (value) => {
+                    return value.title === lastImport.publication.title;
```
danielweck: `publication.title` does not uniquely identifies a publication download. Instead, `publication.url` should used.

src/renderer/redux/reducers/download.ts
```diff
@@ -0,0 +1,37 @@
...
+    action: Action<DownloadPayload>,
+): DownloadState {
+    const downloads = state.downloads;
+    const title = oc(action).payload.title(undefined);
+    switch (action.type) {
+        case downloadActions.ActionType.DownloadRequest:
+            downloads.push({title});
+            return Object.assign({}, state, { downloads });
+        case downloadActions.ActionType.DownloadSuccess:
+            const index = downloads.findIndex((value) => value.title === title);
```
danielweck: Here also: `publication.title` does not uniquely identifies a publication download. Instead, `publication.url` should used.

Issue #430, opened 2019-07-15
`deLang.json` is missing (`enLang.json` and `frLang.json` ok)


danielweck:
https://github.com/readium/readium-desktop/blob/541e48bbe45407a8918018e0076d4ca5ebecd68b/src/common/services/translator.ts#L15

https://github.com/readium/readium-desktop/tree/develop/src/resources/locales

danielweck:
Possible source:
https://github.com/haliaeetus/iso-639

panaC:
Where is the source for 'enLang' and 'frLang' ?

danielweck:
No idea. @Scarsniik ?

Scarsniik:
I made a python script to parse csv from wikipedia pages and used it with the fench and english pages

llemeurfr:
It would be good to have the script saved in the 'scripts' folder of the project.

panaC:
We can use this package : https://github.com/michaelwittig/node-i18n-iso-countries
It seems great with a last commit yesterday and 287stars
Or we can just extracts the json files from this package.

danielweck:
Yep, or https://github.com/haliaeetus/iso-639
(there are a few reliable sources)

danielweck:
Note that this PR moves the locale names into a different folder: https://github.com/readium/readium-desktop/pull/506

danielweck:
PR https://github.com/readium/readium-desktop/pull/506 has been merged, see new folder location for locale names: https://github.com/readium/readium-desktop/tree/develop/src/resources/locale-names

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/519

panaC:
https://github.com/haliaeetus/iso-639 and https://github.com/michaelwittig/node-i18n-iso-countries doesn't allow to get back the language translation from locale code.

I try to use this package https://www.npmjs.com/package/cldr, but it need access to CLDR XML files. This file can't be opened from renderer

panaC:
@Scarsniik  Can you make the same things with your parser for DE ? 
I think you have the best solution

https://de.wikipedia.org/wiki/Liste_der_ISO-639-1-Codes

danielweck:
@Scarsniik
> I made a python script to parse csv from wikipedia pages and used it with the fench and english pages

@llemeurfr 
> It would be good to have the script saved in the 'scripts' folder of the project.

+1 :)

https://github.com/readium/readium-desktop/tree/develop/scripts

[...]

PR #647, opened 2019-08-29
add de languages names + converter from csv to json files

Scarsniik:
Fixes issue #430

I added my python script to parse csv into json but I don't know where to explain how it works (it's easy but it must be explain anyway)


llemeurfr:
please add a readme.md in the same folder, with a section relative to this script. Other sections about other scripts may be added later.

Scarsniik:
I added the readme.md

PR #648, opened 2019-08-29
[fix] display name in HOC withApi

panaC:
https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging


danielweck:
I do not have the expertise to make an informed comment, other than "the code looks okay to me" :)
So, if you are happy with your code, I am happy to merge this PR. Let me know please.

PR #485, opened 2019-07-29
[fix #464] disable console in package mode

panaC:
What do you think @danielweck ?

PR #649, opened 2019-08-30
[fix] restore console for yargs

panaC:
yargs used console and doesn't used process.stdout


danielweck:
Note, this is a revert of PR https://github.com/readium/readium-desktop/pull/485

PR #659, opened 2019-09-02
[TYPING] API, DI

panaC:
I haven't typed withApi, because I manage a new PR with a new APi access entire typed

Di renderer and main has 100% typed
Api in main process has 100% typed
Api in renderer process hasn't correctly typed, I reserved this for a next typing PR

I had to disable no-floating-promise : it's temporary. indeed the actual typing in main API is correct, but the handle with the hoc withApi hasn't the same behaviour. In this case the actual typing will be used in the next PR renderer api typing.


src/renderer/components/dialog/SameFileImportConfirm.tsx
```diff
@@ -53,19 +50,17 @@ class SameFileImportConfirm extends React.Component<Props> {
 
     private addToCatalog() {
         this.props.importOpdsEntry(
-            {
-                url: this.props.publication.url,
-                base64OpdsPublication: this.props.publication.base64OpdsPublication,
-                downloadSample: this.props.downloadSample,
-                title: this.props.publication.title,
-            },
```
panaC: this actual call to this function doesn't works, this PR may fix it

src/renderer/components/dialog/SameFileImportConfirm.tsx
```diff
@@ -53,19 +50,17 @@ class SameFileImportConfirm extends React.Component<Props> {
 
     private addToCatalog() {
         this.props.importOpdsEntry(
```
panaC: this actual call to this function doesn't works, this PR may fix it

src/main/redux/sagas/api.ts
```diff
@@ -5,29 +5,37 @@
...
+        const apiModule = diMainGet(getSymbolName(api.moduleId));
+        const apiMethod = apiModule[api.methodId].bind(apiModule);
+
+        debug(api.moduleId, api.methodId, requestAction.payload);
+
         const result = yield call(
             apiMethod,
-            requestAction.payload,
+            ...(requestAction.payload || []),
         );
```
danielweck: Question: what if the action `payload` is an `array` *meant* to be a single function parameter to the API method?

src/renderer/components/publication/publicationInfos/PublicationInfo.tsx
```diff
@@ -5,39 +5,31 @@
...
-import * as styles from "readium-desktop/renderer/assets/styles/bookDetailsDialog.css";
-
 interface Props extends TranslatorProps {
     publicationIdentifier: string;
     isOpds?: boolean;
-    publication?: PublicationView;
+    publication?: TPublicationApiGet_result;
     closeDialog?: any;
-    getPublicationFromId?: any;
+    getPublicationFromId?: () => void;
```
danielweck: Shouldn't `() => void` be `TPublicationApiGet`? ...but actual usage of `getPublicationFromId()` does not correspond to:

```
        operations: [
            {
                moduleId: "publication",
                methodId: "get",
                resultProp: "publication",
                callProp: "getPublicationFromId",
                buildRequestData,
            },
        ],
```

?!

src/main/redux/sagas/api.ts
danielweck: Ah, I think that you modified all occurrences in the code so that `buildRequestData()` always returns an array, which means that a single array parameter would in fact be array-of-array, right? (so that the spread operator generates the original encapsulated array)

tslint.json
```diff
@@ -1,8 +1,9 @@
 {
   "extends": "tslint:recommended",
   "rules": {
+    "no-string-literal": true,
     "no-reference": false,
-    "no-floating-promises": true,
+    "no-floating-promises": false,
```
danielweck: Oh, interesting change! :)

If `true`:

```
ERROR: - Promises must be handled appropriately
src/renderer/components/catalog/PublicationAddButton.tsx:54:9
src/renderer/components/dialog/DeleteOpdsFeedConfirm.tsx:52:9
src/renderer/components/dialog/DeletePublicationConfirm.tsx:52:9
src/renderer/components/dialog/FileImport.tsx:40:13
src/renderer/components/dialog/LcpAuthentication.tsx:76:9
src/renderer/components/dialog/OpdsFeedAddForm.tsx:88:9
src/renderer/components/dialog/RenewLsdConfirm.tsx:51:9
src/renderer/components/dialog/ReturnLsdConfirm.tsx:52:9
src/renderer/components/dialog/SameFileImportConfirm.tsx:52:9
src/renderer/components/opds/BrowserResult.tsx:110:9
src/renderer/components/publication/TagManager.tsx:115:9
src/renderer/components/publication/menu/PublicationExportButton.tsx:61:9
src/renderer/components/reader/Reader.tsx:545:9
src/renderer/components/reader/Reader.tsx:651:17
src/renderer/components/reader/Reader.tsx:655:13
src/renderer/components/reader/UpdateBookmarkForm.tsx:66:9
src/renderer/components/searchResult/TextSearchResult.tsx:35:13
src/renderer/components/utils/SameFileImportManager.tsx:35:17
src/renderer/components/utils/SameFileImportManager.tsx:52:17
```

I think we should implement this correctly, by adding `async` and `await` where necessary (otherwise we are just obfuscating a coding mistake). For example:

```
    private async importFile(event: any) {
        const files = event.target.files;
        const paths: string[] = [];

        for (const f of files) {
            paths.push(f.path);
        }
        await this.props.importFiles(paths);
    }
```

...which in this case means that the `importFile()` function called by the `input@onChange` event handler will return a "dangling" / unhandled Promise, immediately discarded in the call stack (local variable in the function body), just like when `this.props.importFiles(paths)` is called without `await`.

```
onChange={ this.importFile }
```

src/main/api/publication.ts
```diff
@@ -84,18 +135,22 @@ export class PublicationApi {
...
         let returnView: PublicationView;
-        let title: string;
+        let titleView: string;
         // if url exist import new entry by download
-        if (data.publication.url) {
-            const httpPub = await this.catalogService.importOpdsEntry(data.publication.url, data.downloadSample);
+        if (url) {
+            const httpPub = await this.catalogService.importOpdsEntry(url, downloadSample);
             if (httpPub.isSuccess) {
                 title = httpPub.data.title;
```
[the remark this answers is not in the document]
panaC: title was assigned but not used in future
it's titleView assigned,
I'm right ?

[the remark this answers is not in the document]
panaC: https://github.com/readium/readium-desktop/pull/659/commits/6d679b73537f831c80daf23577837fd6ddf7e448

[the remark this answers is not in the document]
danielweck: So if the `if` conditions `url` and `httpPub.isSuccess` are true, `titleView` remains `undefined` in the call to `this.translator.translate("message.download.success", { title: titleView })`.

Is that a bug in the PR code, or a feature?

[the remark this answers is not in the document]
panaC: yes, you're right, i have to replace title with titleview

panaC:
i fixed titleview.
Your comment has disappeared in my side.
Bref..
You can merge it ?

src/renderer/components/reader/Reader.tsx
```diff
@@ -164,27 +170,26 @@ interface ReaderProps {
...
 
 export class Reader extends React.Component<ReaderProps, ReaderState> {
     private fastLinkRef: any;
 
-    @lazyInject("store")
+    @lazyInject(diRendererSymbolTable.store)
     private store: Store<RootState>;
 
-    @lazyInject("translator")
+    @lazyInject(diRendererSymbolTable.history)
```
danielweck: Damn, I missed this fatal coding error (probably copy/paste mistake) during my PR review!
This line should be `@lazyInject(diRendererSymbolTable.translator)`
I am creating a quickfix into the `develop` branch.

danielweck: @panaC can the static TypeScript compiler checks verify that `diRendererSymbolTable.history` is not compatible with the type `Translator`? Right now the fatal crash is only detectable at runtime :(

danielweck: Hotfix: https://github.com/readium/readium-desktop/commit/12fcce9de927a37cc8b0c17731d78b3cc46546bb

danielweck:
Post-merge hotfix: https://github.com/readium/readium-desktop/commit/12fcce9de927a37cc8b0c17731d78b3cc46546bb

danielweck:
Regression bug: https://github.com/readium/readium-desktop/issues/709

src/main/redux/sagas/api.ts
```diff
@@ -5,29 +5,37 @@
...
     try {
+        const apiModule = diMainGet(getSymbolName(api.moduleId));
+        const apiMethod = apiModule[api.methodId].bind(apiModule);
+
+        debug(api.moduleId, api.methodId, requestAction.payload);
+
         const result = yield call(
             apiMethod,
-            requestAction.payload,
+            ...(requestAction.payload || []),
```
danielweck: I think this is responsible for this regression bug:
https://github.com/readium/readium-desktop/issues/709

danielweck: Hotfix: https://github.com/readium/readium-desktop/commit/baa17321404985e5d721f8894203a745091df73b

danielweck:
Another regression bug: https://github.com/readium/readium-desktop/issues/711

PR #661, opened 2019-09-02
Fixes #660 reader paginated mode button

### LINKED ISSUE #660: Regression, impossible to change disposition / scrollable to paginated
On a reflow ebook, I open the ebook, open the tab controlling parameters / Display.
I'm in scrollable mode at this point. I select Paginated  (several times) -> no action.

Now I select Scrollable (which is the current mode) then Paginated -> it works.

Scarsniik:


src/renderer/components/reader/Reader.tsx
```diff
@@ -670,6 +670,7 @@ export class Reader extends React.Component<ReaderProps, ReaderState> {
         settingsValues[name] =  value;
 
         this.setState({settingsValues});
+        console.log(name, givenValue, settingsValues);
```
panaC: console.log :)

src/renderer/components/reader/ReaderOptions.tsx
```diff
@@ -198,6 +198,7 @@ export class ReaderOptions extends React.Component<Props> {
 
     private displayContent() {
         const {__, settings} = this.props;
+        console.log(settings.paged);
```
panaC: console.log :)

PR #662, opened 2019-09-03
remove lcp import

panaC:
@llemeurfr


llemeurfr:
Note: as there will be no LCP support in v1.0, we have to remove temporarily a message (displayed on an empty bookshelf) and the .lcpl import green light.

danielweck:
I created a new issue so we can easily revert this PR:
https://github.com/readium/readium-desktop/issues/663

PR #667, opened 2019-09-04
Keyboard-accessible and screen-reader friendly navigation (TOC, landmarks), skip-to-content main area quick-access

### LINKED ISSUE #156: [a11y] Navigate to chapters through the Table of Contents
OS: Windows 10
Assistive technology: NVDA 2018.1.1

When on the TOC you click on a entry, the text of the chapter appears but the focus remains on the topbar: the hyperlink anchor should be focused

### LINKED ISSUE #565: Move from chapter to chapter
In the Reading view, for a blind person, the left and right keys and arrows (which move from page to page or from screen to screen in scroll view) don’t mean anything (the screen reader reads the whole resource). 
They would prefer a direct action for moving to the next/previous chapter (i.e. resource). And sighed people prefer the arrow keys are currently designed and are ready to use the timeline or ToC to move to another chapter. 

Therefore we need keyboard shortcuts for moving to next/previous chapters. These will be known from blind people, but there will be no equivalent visual button in the Reading View. 

There is not such key defined in https://github.com/readium/readium-js-viewer/wiki/Keyboard.
Ctrl+Left/right Arrows are already taken by JAWS/NVDA for previous/net Word.
Up/Down Arrows for prev/next line.
(source http://www.daisy.org/daisypedia/adobe-digital-editions-overview-and-accessibility-tips)

Proposal: like AZARDI, use Shift Right Arrow or Shift PgUp and Shift Left Arrow or Shift PgDn
source: http://www.daisy.org/daisypedia/azardi-epub-reader-windows

See also #155

danielweck:
Fixes #565
Fixes #156

(TOC, landmarks, etc.) and spine-level previous/next nav in addition to page-level (requires latest r2-navigator-js package)


danielweck:
Hotfix (no PR):  https://github.com/readium/readium-desktop/commit/72111aeb336dcd2a76e51f4ff9206e92f7ab8523

Rationale: screen readers have their own shortcut to activate hyperlinks (e.g. VoiceOver CTRL+OPT+SPACE), so we must not limit the focus behaviour to app-defined keyboard interaction (i.e. ENTER key) - note that this means the focus is moved even when TOC items clicked with mouse, which is fine.

PR #668, opened 2019-09-04
Fixes #82 add go to page in the reader

### LINKED ISSUE #82: EPUB page-list UI (for page-breaks)
This is implemented in `r2-testapp-js`, but is currently missing in `readium-desktop`.

Scarsniik:


src/resources/locales/de.json
```diff
@@ -151,6 +151,10 @@
             "bookmarkTitle": "bookmark",
             "detachWindowTitle": "das Fenster lösen",
             "fullscreenTitle": "fullscreen mode",
+            "goTo": "Go to",
+            "goToError": "This page does not exist",
+            "goToPlaceHolder": "Go to page",
+            "goToTitle": "Change page",
```
danielweck: Anybody with German skills? :)
@panaC ? @llemeurfr ?

panaC: I have my friend https://www.deepl.com/translator , if you likes, :)

src/renderer/components/reader/ReaderMenu.tsx
```diff
@@ -195,9 +219,55 @@ export class ReaderMenu extends React.Component<Props, State> {
...
+    }
+
     private closeBookarkEditForm() {
         this.setState({ bookmarkToUpdate: undefined });
     }
+
+    private handleSubmitPage(e: any) {
+        e.preventDefault();
+        const pageNbr = this.goToRef.value;
+        const foundPage = this.props.publication.PageList.find((page) => page.Title === pageNbr.toString());
```
danielweck: `page` is in fact a `Link` object. `page.Title` is a "cleaned-up" string extracted from the text descendants of the corresponding `<a>` HTML element in the EPUB3 Navigation Document (`<nav epub:type="pagelist"`). The "cleanup" method is: `pageLink.Title = aElement.textContent.trim().replace(/\s\s+/g, " ")`.

`pageNbr` is an arbitrary string entered by users (it should not be limited to a number/integer, because of roman numerals such as `ii` - but I am filing a separate issue for this). So `pageNbr` must be cleaned up as well, so that it can be compared with `page.Title`.

src/renderer/components/reader/ReaderMenu.tsx
```diff
@@ -195,9 +219,55 @@ export class ReaderMenu extends React.Component<Props, State> {
...
+                    aria-relevant="all"
+                    role="alert"
+                >
+                    { __("reader.navigation.goToError") }
+                </p>
+            }
+            <form onSubmit={this.handleSubmitPage}>
+                <input
+                    ref={(ref) => this.goToRef = ref}
+                    type="number"
```
danielweck: This should be a text input field so that users can enter any arbitrary string, such as Roman numerals (e.g. `ii`).
Also, `type="number"` allows negative values which is weird :)

src/renderer/components/reader/sideMenu/SideMenu.tsx
```diff
@@ -57,15 +57,17 @@ export class SideMenu extends React.Component<Props, State> {
...
+                            <SideMenuSection
+                                open={ openedSection === index }
+                                id={index}
+                                key={index}
+                                title={section.title}
+                                content={section.content}
+                                onClick={this.handleClickSection}
+                                disabled={section.disabled}
+                            />
+                        : section.content,
```
danielweck: Missing `key={index}` on iterated React structure.

src/renderer/components/reader/ReaderMenu.tsx
```diff
@@ -92,7 +116,7 @@ export class ReaderMenu extends React.Component<Props, State> {
                 toggleMenu={toggleMenu}
                 focusMenuButton={this.props.focusNaviguationMenu}
             />
-        );
+        </> );
```
danielweck: Could you explain the rationale for the addition of the wrapping React `<>` `</>` fragment?
(sorry if this is a stupid question, I just want to make sure I understand the reason, to educate myself)

src/renderer/components/reader/ReaderMenu.tsx
```diff
@@ -39,23 +40,41 @@ interface Props extends TranslatorProps {
...
+            pageError: false,
+            refreshError: false,
         };
 
         this.closeBookarkEditForm = this.closeBookarkEditForm.bind(this);
+        this.handleSubmitPage = this.handleSubmitPage.bind(this);
+    }
+
+    public componentDidUpdate() {
+        if (this.state.refreshError) {
```
danielweck: The error message is too "sticky", it should disappear as soon as the user starts typing a new value in the input text field for "goto page".

ThoriumGotoPageStickyError

src/renderer/components/reader/ReaderMenu.tsx
danielweck: Furthermore, the built-in HTML validation feature does not seem to take into account the document language (HTML `lang` attribute), resulting in locale mismatch (see screenshot below with exponent notation).

danielweck: ThoriumGotoPageInputValidationNumber

src/resources/locales/de.json
```diff
@@ -151,6 +151,10 @@
             "bookmarkTitle": "bookmark",
             "detachWindowTitle": "das Fenster lösen",
             "fullscreenTitle": "fullscreen mode",
+            "goTo": "Go to",
+            "goToError": "This page does not exist",
+            "goToPlaceHolder": "Go to page",
```
danielweck: I think that the text input field "placeholder" should be something like "enter a page number", not a description of the associated action.

src/resources/locales/en.json
```diff
@@ -151,6 +151,10 @@
             "bookmarkTitle": "bookmark",
             "detachWindowTitle": "detach the window",
             "fullscreenTitle": "open fullscreen mode",
+            "goTo": "Go to",
+            "goToError": "This page does not exist",
+            "goToPlaceHolder": "Go to page",
```
danielweck: I think that the text input field "placeholder" should be something like "enter a page number", not a description of the associated action.

src/resources/locales/en.json
```diff
@@ -151,6 +151,10 @@
...
+            "goToTitle": "Change page",
```
danielweck: I think that "Change page" is not an adequate description of the action here, what about "Go to page break" instead?

src/resources/locales/en.json
```diff
@@ -151,6 +151,10 @@
             "bookmarkTitle": "bookmark",
             "detachWindowTitle": "detach the window",
             "fullscreenTitle": "open fullscreen mode",
+            "goTo": "Go to",
+            "goToError": "This page does not exist",
```
danielweck: I would be more specific here: "This page break does not exist"

src/resources/locales/en.json
danielweck: ...or "Go to page" which is less technical that "page break" :)

src/renderer/components/reader/ReaderMenu.tsx
```diff
@@ -195,9 +219,55 @@ export class ReaderMenu extends React.Component<Props, State> {
...
+            {error &&
+                <p
+                    className={styles.goToErrorMessage}
+                    aria-live="assertive"
+                    aria-relevant="all"
+                    role="alert"
+                >
+                    { __("reader.navigation.goToError") }
+                </p>
+            }
```
danielweck: I think that the error should be located below the text input field, because right now the error "pushes down" the interactive input control when it appears.

danielweck:
Could you please remove the `package-lock.json` changes from this PR, just so that we can track explicit updates of NPM package dependencies dedicated PRs? Many thanks!

src/renderer/components/reader/ReaderMenu.tsx
danielweck: Fixed via https://github.com/readium/readium-desktop/commit/d38073a801fbf57a96525287a026e9b0da06f196

src/resources/locales/en.json
danielweck: Fixed via https://github.com/readium/readium-desktop/commit/d38073a801fbf57a96525287a026e9b0da06f196

src/resources/locales/en.json
danielweck: Fixed https://github.com/readium/readium-desktop/commit/d38073a801fbf57a96525287a026e9b0da06f196

src/renderer/components/reader/ReaderMenu.tsx
Scarsniik: I juste forgot to remove it, I wanted to add something but I found an other way to do it

src/renderer/components/reader/ReaderMenu.tsx
Scarsniik: I changed the pageNbr. It's now a string

src/resources/locales/de.json
Scarsniik: I translated it with google translate

src/renderer/components/reader/ReaderMenu.tsx
danielweck: Yes, but it is not normalized (leading+trailing trim of whitespace characters, and inner whitespace collapsing).
Once it is normalized, than it can be compared reliably with the base `pageLink.Title`

src/renderer/components/reader/ReaderMenu.tsx
danielweck: Do you intend to remove it, or leave it?

Scarsniik: I removed it

src/renderer/components/reader/ReaderMenu.tsx
Scarsniik: I added the same normalization

PR #680, opened 2019-09-05
Fixes #626 : multiple import don't stop if an import fail

### LINKED ISSUE #626: Multiple EPUB files drag-drop (bulk import), first fail aborts rest of the import queue, should only skip the failing import
In the screenshot example, the second EPUB fails to import (because LCP support currently broken, or erroneous LCP data inside publication), causing the entire rest of the import queue to be aborted. Instead, the culprit publication should be skipped, and announced as failing.

ThoriumDragDropBulkImportFail

Scarsniik:


src/main/services/catalog.ts
```diff
@@ -270,60 +272,66 @@ export class CatalogService {
...
-        debug("[END] Store publication in database", filePath);
+            debug("Publication imported", filePath);
+            return newPubDocument;
+        } catch (error) {
+            debug("Import EPUB File - ERROR", error);
+        }
 
-        debug("Publication imported", filePath);
-        return newPubDocument;
+        return undefined;
```
danielweck: The current return type for the `importEpubFile()` function is `Promise<PublicationDocument>`, so in principle `return undefined` (equivalent to `return Promise.resolve(undefined)`) is incompatible. This can cause problems on the client side, as a valid `PublicationDocument` object is always expected, as per the contract expressed in the function signature. Instead, any `null` or `undefined` value should effectively be a Promise rejection (i.e. `throw error` or `return Promise.reject(error)`, which are semantically equivalent).

I think that any uncaught exception or Promise.rejection in this function block should be allowed to cascade outside of the `importFile()` function. In order words I think that there should be a rejection handler on the calling side of `importFile()` (I assume there is a loop to iterate multiple candidate import files).

PS: in fact, `return null` in `importFile()` is also incompatible with the return type `Promise<PublicationDocument>`, so I guess the TypeScript compiler is currently too relaxed, which would explain why the TypeScript code contains such inconsistencies / typing errors (see the `strictNullChecks` PR https://github.com/readium/readium-desktop/pull/475 )

danielweck: Yes, I confirm that adding `"strictNullChecks": true,` in `tsconfig.json` triggers the TypeScript compiler error:
`Type 'null' is not assignable to type 'PublicationDocument'.ts(2322)`
for `return null` in `importFile()`, and `return undefined` in `importEpubFile()`.

Anyway, this is a more general problem related to the entire Thorium codebase, which contains many oversights like this (including the `any` proliferation). We will need to fix at a later date, as this is a huge refactoring task.

danielweck: To summarize my PR review:

* `importFile()` should be allowed to reject/throw instead of returning/resolving a valid promise value (`PublicationDocument`). We should not have a giant `try/catch`, as this only obfuscates the problem, and it creates a secondary problem of `undefined` or `null` returned value instead of a correct promise type.
*  So instead, the caller(s) of `importFile()` should correctly handle promise rejection (typically, using the `try/catch` `async/await` notation, or the legacy `promise.then().catch()` chaining syntax). Rejections can be the result of explicit `return Promise.reject()` or a unexpected exception `throw` from deep inside the code (e.g. the LCP error that made me discover the bug, see https://github.com/readium/readium-desktop/issues/626 ).

Generally-speaking, the Thorium codebase lacks proper handling of rejected promises (i.e. edge / error cases). We started cleaning-up / improving the code flow a few weeks ago, but there is still much to do :)

danielweck: Here is a concrete code adaptation resulting from the `strictNullCheck` static compiler verification of `null` / `undefined` in returned promise types:
https://github.com/readium/readium-desktop/pull/475/files#diff-968212bd5d4ddfe1c6bab01fa6bf4e16
(`if (newDoc)`)

This is just a naive code fix in order to attempt to pass the build, not a proper fix.
A proper fix would consist in capturing possible promise rejections inside the iteration.
(`for (const path of paths)`)
...and `continue` after dumping a friendly error message in the console :)
(ideally, there should be an error handling strategy, for example keeping track of every failure, and reporting them to the end-user)

src/main/api/publication.ts
```diff
@@ -122,8 +126,14 @@ export class PublicationApi {
         const newDocs = [];
 
         for (const path of paths) {
-            const newDoc = await this.catalogService.importFile(path);
-            newDocs.push(newDoc);
+            try {
+                const newDoc = await this.catalogService.importFile(path);
+                newDocs.push(newDoc);
```
danielweck: The signature of the `importFile()` function indicates `Promise<PublicationDocument>` return type, but the function body can in fact return `null` (when `strictNullChecks` is enabled in `tsconfig.json`, the TypeScript compiler will detect this coding error).

So, I would suggest not pushing `newDoc` into the array because otherwise this edge-case will crash `convertDocumentToView()`.

Issue #410, opened 2019-07-11
CSS is currently centralized, should be componentized?


danielweck:
https://github.com/readium/readium-desktop/tree/develop/src/renderer/assets/styles

=> all in one folder, but should be tied to specific JSX component?

danielweck:
Related: https://github.com/readium/readium-desktop/issues/409

danielweck:
A concrete problematic situation related to this issue:
https://github.com/readium/readium-desktop/pull/681#issuecomment-528363734

danielweck:
Also note that some SVG icons contain non-parameterized inline CSS definitions:
https://github.com/readium/readium-desktop/tree/develop/src/renderer/assets/icons

For example:
```
<style type="text/css">
	.st0{fill:#C1C1C1;}
</style>
```

This could cause problems when implementing support for OS high-contrast / dark mode.
See https://github.com/readium/readium-desktop/issues/577

The solution we will likely implement is CSS "variables" (custom properties), just like ReadiumCSS, at which point the SVG icons might need tweaking too.

PR #681, opened 2019-09-05
[PARKED] (DO NOT MERGE) attempt to fix #678 by migrating the reader's sidemenu into an overlay portal

### LINKED ISSUE #678: reader view: when side menu is open, screen reader can access "hidden" UI below the overlay
This was solved for Dialog and MenuContent by using `"app-overlay"` and `ReactDOM.createPortal()`.

The problem with SideMenu is that the top horizontal button bar remains accessible, and it is inside the `"app"` viewport which the portal + `aria-hidden` technique of Dialog and MenuContent would conflict with.

danielweck:
Same React technique as modal dialog and popup menu, in order to set aria-hidden on the principal app viewport (screen reader), and ensure consistency between keyboard focus lock (cycle within the sidemenu) and mouse interaction (top bar button icons should not be interacted with when the menu is open, if keyboard cannot access them without closing the menu first)

This PR currently fails, because of:
1) CSS containment hierarchy is broken, so the menu styles are screwed (notably, the sliders)
2) click-to-close outside of the sidemenu needs to be re-implemented following the same design pattern as modal dialog and popup menu (in addition to visual blurred / translucent background overlay)
3) sidemenu needs to stretch the whole height


danielweck:
Point of interest: if the CSS styles were componentized instead of centralized (see `reader-app.css`), it would be a lot easier to scope changes to a particular React component (in this case: to migrate a whole set of related styles to the modal React portal). Right now the hard-scoping with `#ID` is a brittle design approach, and the monolithic CSS file is very hard to maintain / refactor.
So, related issue: https://github.com/readium/readium-desktop/issues/410

PR #696, opened 2019-09-11
[READY TO MERGE][TYPING] replace withApi and typing react component

panaC:
@danielweck Can you pass the first layer review
I added an all checks of react component (except Reader : only API change)


src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export async function apiFetch<T extends TApiMethodName>(path: T, ...requestData: Parameters<TApiMethod[T]>) {
+    return new Promise<ReturnPromiseType<TApiMethod[T]>>((resolve, reject) => {
+        const store = diRendererGet("store");
+        const requestId = uuid.v4();
+        const moduleId = path.split("/")[0] as TModuleApi;
+        const methodId = path.split("/")[1] as TMethodApi;
+        let lastSuccess: ApiLastSuccess | undefined;
+        let storeUnsubscribe: Unsubscribe| undefined;
+        let timeout: number | undefined;
```
danielweck: 1) I think that the variable `timeout` should be renamed to `timeoutId` (just like in https://github.com/readium/readium-desktop/blob/develop/src/utils/debounce.ts ) because otherwise this gives the reader the impression that `timeout` is a `number` of seconds (when in fact it is a unique "identifier").
2) To avoid hard-coding the TypeScript type `number` vs. `NodeJS.Timer`, would it be better to use `ReturnType<typeof setTimeout>` instead?

src/renderer/apiRefresh.ts
```diff
@@ -0,0 +1,45 @@
...
+import { TMethodApi, TModuleApi } from "readium-desktop/main/di";
+import { diRendererGet } from "readium-desktop/renderer/di";
+import { ApiLastSuccess } from "readium-desktop/renderer/redux/states/api";
+
+/**
+ * subscribe to redux to automaticaly execute callback when any path in parameter is called on IPC
+ *
+ * don't forget to unsubscribe with the return value of function
+ */
+export function apiRefresh(pathArrayToRefresh: TApiMethodName[], cb: () => void | Promise<void>) {
```
danielweck: Just to improve code legibility / readability, I would add the return type `Unsubscribe` explicitly (rather than relying on implicit type inference in IDEs like Visual Studio Code)

danielweck:
TODO: remove or comment-out the file https://github.com/readium/readium-desktop/blob/6cbd90399ae4c9bcefed0880c8ced776bf57fa5b/src/renderer/components/utils/hoc/api.tsx

src/typings/pierre.d.ts
```diff
@@ -0,0 +1,11 @@
...
+// Copyright 2017 European Digital Reading Lab. All rights reserved.
+// Licensed to the Readium Foundation under one or more contributor license agreements.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END==
+
+/**
+ * Obtain the promise return type of a function type
+ */
+export type ReturnPromiseType<T extends (...args: any) => any> = T extends (...args: any) => Promise<infer R> ? R : any;
```
danielweck: The file is currently named `pierre.d.ts`. What about  `promise.d.ts` instead?

src/renderer/components/utils/menu/AccessibleMenu.tsx
```diff
@@ -82,6 +80,8 @@ export default class AccessibleMenu extends React.Component<Props, State> {
             && this.state.triggerElem
         ) {
             if (this.state.triggerElem != null) {
+                // WHY ?
+                // @ts-ignore
```
danielweck: Simply change the line `triggerElem: document.activeElement` to `triggerElem: document.activeElement as HTMLElement`, and `triggerElem: Element | undefined;` to `triggerElem: HTMLElement | undefined;`, then remove the `@ts-ignore` comment.

src/renderer/components/utils/menu/AccessibleMenu.tsx
```diff
@@ -134,11 +134,13 @@ export default class AccessibleMenu extends React.Component<Props, State> {
...
     }
 
-    private handleFocus(event: any) {
+    private handleFocus(event: Event) {
         const focusedNode = event.target;
 
         if (this.containerRef
             && this.containerRef.current
+        // WHY ?
+        // @ts-ignore
```
danielweck: Simply replace the line `const focusedNode = event.target;` with `const focusedNode = event.target as HTMLElement;` and remove the `@ts-ignore` comment.

src/renderer/components/opds/Header.tsx
```diff
@@ -23,17 +23,21 @@ export enum DisplayType {
...
     public render(): React.ReactElement<{}> {
         const { __ } = this.props;
         const search = parseQueryString(this.props.location.search.replace("?", ""));
         const displayType = search.displayType || DisplayType.Grid;
         delete(search.displayType);
 
+        /**
+         * Why css style is apply in code and not imported from css ressource ?
+         * code:cssincode
+         */
```
danielweck: Is the `code:cssincode` marker a known syntax? (i.e. recognized by Visual Studio Code, or a plugin?)
If not, may I suggest using `TODO` or `FIXME`, with a comment like `extract inline CSS into https://github.com/readium/readium-desktop/blob/develop/src/renderer/assets/styles/opds.css` (for example)

src/renderer/components/catalog/PublicationAddButton.tsx
```diff
@@ -44,18 +40,23 @@ export class PublicationAddButton extends React.Component<Props> {
...
 
-    private importFile(event: any) {
+    private importFile(event: TChangeEvent) {
         const files = event.target.files;
         const paths: string[] = [];
 
-        for (const f of files) {
-            paths.push(f.path);
+        // files: FileList doesn't have [symbol.interator()]
+        // tslint:disable-next-line: prefer-for-of
```
danielweck: Suggestion: edit `tsconfig.json`, change `"lib": ["es2017", "es6", "dom"],` to `"lib": ["es2017", "es6", "dom", "dom.iterable"],` (added `"dom.iterable"`), and restore the `for (const f of files)` code.

src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,79 @@
...
+import { apiActions } from "readium-desktop/common/redux/actions";
+import { TApiMethod, TApiMethodName } from "readium-desktop/main/api/api.type";
+import { TMethodApi, TModuleApi } from "readium-desktop/main/di";
+import { diRendererGet } from "readium-desktop/renderer/di";
+import { ApiLastSuccess } from "readium-desktop/renderer/redux/states/api";
+import { ReturnPromiseType } from "readium-desktop/typings/pierre";
+import { Unsubscribe } from "redux";
+import * as uuid from "uuid";
+
+export async function apiFetch<T extends TApiMethodName>(path: T, ...requestData: Parameters<TApiMethod[T]>) {
```
danielweck: The term `fetch` is overloaded ( https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API ), and I think this leads to confusion when reading the code (i.e. usages of the `apiFetch()` function).

Given that both `apiFetch()` and `apiRefresh()` are really about "store" ("dispatch" / "subscribe") and "state", can we find more suitable names that represent what the functions actually do?

src/renderer/components/dialog/DeleteOpdsFeedConfirm.tsx
```diff
@@ -28,33 +32,37 @@ export class DeleteOpdsFeedConfirm extends React.Component<DeleteOpdsFeedConfirm
     }
 
     public render(): React.ReactElement<{}> {
-        const {__} = this.props;
-        if (!this.props.feed) {
-            return <></>;
+        if (!this.props.open) {
```
danielweck: Shouldn't this conditional statement also check for `!this.props.feed`?

src/renderer/components/opds/Entry.tsx
```diff
@@ -7,21 +7,23 @@
...
+    level?: number;
+    entry: OpdsLinkView;
 }
 
-export class Entry extends React.Component<EntryProps, undefined> {
+class Entry extends React.Component<EntryProps, undefined> {
     public render(): React.ReactElement<{}>  {
         const { entry } = this.props;
 
+        // why ? entry is required
```
danielweck: I agree that this conditional statement should be removed.
Rationale: in principle, better to see a "crash" because of `undefined` `entry` in the following `entry.url` line of code, rather than silently ignoring the (impossible) occurrence of undefined `entry: OpdsLinkView`.

src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,79 @@
...
+                    return;
+                }
+
+                // New api success
+                lastSuccess = apiLastSuccess;
+
+                const meta = apiLastSuccess.action.meta.api;
+                if (moduleId === meta.moduleId && methodId === meta.methodId && state.api.data[requestId]) {
+                    const request = Object.assign({}, state.api.data[requestId]);
+                    store.dispatch(apiActions.clean(requestId));
```
danielweck: In the old `withAPI` HigherOrderComponent (HOC), `apiActions.clean(ID)` was called for each "operationRequest" in the `componentWillUnmount` React lifecycle function. Furthermore, it was possible to dispatch the "clean" action via the component's props, for example in `src/renderer/components/opds/BrowserResult.tsx` immediately before the `apiFetch("opds/browse")`.
Strictly-speaking, the (re)implementation / refactoring in this PR breaks the previous contract, because "clean" is only invoked in the very specific case of `moduleId === meta.moduleId && methodId === meta.methodId && state.api.data[requestId]`. Is that an oversight, or is it intentional? (in which case, I imagine that the "clean" in other scenarios was deemed useless, which justified its removal?)

src/renderer/components/opds/BrowserResult.tsx
```diff
@@ -76,38 +88,46 @@ export class BrowserResult extends React.Component<BrowserResultProps, null> {
...
-            newUrl = result.data.searchUrl;
+
+        if (search && browserResult && browserResult.isSuccess && browserResult.data.searchUrl) {
+            newUrl = browserResult.data.searchUrl;
             newUrl = this.addSearchTerms(newUrl, search) +
                 Object.keys(oldQs).map((id) => `&${id}=${oldQs[id]}`).join("");
         }
+
         this.currentUrl = newUrl;
-        this.props.cleanData();
```
danielweck: See my comment about "clean":
https://github.com/readium/readium-desktop/pull/696/files#r327787727

src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,79 @@
...
+import { ReturnPromiseType } from "readium-desktop/typings/pierre";
+import { Unsubscribe } from "redux";
+import * as uuid from "uuid";
+
+export async function apiFetch<T extends TApiMethodName>(path: T, ...requestData: Parameters<TApiMethod[T]>) {
+    return new Promise<ReturnPromiseType<TApiMethod[T]>>((resolve, reject) => {
+        const store = diRendererGet("store");
+        const requestId = uuid.v4();
+        const moduleId = path.split("/")[0] as TModuleApi;
+        const methodId = path.split("/")[1] as TMethodApi;
```
danielweck: For the avoidance of doubt / confusion, I would suggest not using the name `path` for this variable, as this is a commonly used NodeJS package `const path = require("path");`. Perhaps instead: `apiPath`?

src/renderer/apiRefresh.ts
```diff
@@ -0,0 +1,45 @@
...
+            return;
+        }
+
+        // New api success
+        lastSuccess = apiLastSuccess;
+
+        const meta = apiLastSuccess.action.meta.api;
+        if (pathArrayToRefresh.find((path) => {
+            const moduleId = path.split("/")[0] as TModuleApi;
+            const methodId = path.split("/")[1] as TMethodApi;
```
danielweck: For the avoidance of doubt / confusion, I would suggest not using the name `path` for this variable, as this is a commonly used NodeJS package `const path = require("path");`. Perhaps instead: `apiPath`?

src/renderer/apiRefresh.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (!apiLastSuccess || apiLastSuccess.date <= lastSuccessDate) {
+            return;
+        }
+
+        // New api success
+        lastSuccess = apiLastSuccess;
+
+        const meta = apiLastSuccess.action.meta.api;
+        if (pathArrayToRefresh.find((path) => {
+            const moduleId = path.split("/")[0] as TModuleApi;
```
danielweck: Suggestion: `const splitPath = path.split("/");` and then reuse the variable `splitPath` (marginal performance improvement, but more importantly: cleaner code ;)

src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,79 @@
...
+import { ApiLastSuccess } from "readium-desktop/renderer/redux/states/api";
+import { ReturnPromiseType } from "readium-desktop/typings/pierre";
+import { Unsubscribe } from "redux";
+import * as uuid from "uuid";
+
+export async function apiFetch<T extends TApiMethodName>(path: T, ...requestData: Parameters<TApiMethod[T]>) {
+    return new Promise<ReturnPromiseType<TApiMethod[T]>>((resolve, reject) => {
+        const store = diRendererGet("store");
+        const requestId = uuid.v4();
+        const moduleId = path.split("/")[0] as TModuleApi;
```
danielweck: Suggestion: `const splitPath = path.split("/");` and then reuse the variable `splitPath` (marginal performance improvement, but more importantly: cleaner code ;)

src/renderer/apiFetch.ts
panaC: https://github.com/readium/readium-desktop/pull/696/commits/9c17d0bf9740edb8a7a58d28befef85ef0d33cc8

src/renderer/apiRefresh.ts
panaC: 3553a8a

src/typings/pierre.d.ts
panaC: great idea e974e95

src/renderer/components/utils/menu/AccessibleMenu.tsx
panaC: but why force cast Element to HTMLElement, is the typing broken ?

danielweck: `document.activeElement` is a DOM property (HTML, SVG, XML, etc.)
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/activeElement
https://developer.mozilla.org/en-US/docs/Web/API/Document

But in our case, we know that the Thorium UI is built with HTML markup (+ a few SVG images), so we can cast the correct type for better compile-time checking.

src/renderer/components/utils/menu/AccessibleMenu.tsx
panaC: same above

src/renderer/components/opds/Header.tsx
panaC: it's comment code to easily find all css in code with simple `ctrl` `f`.
I added a `noany` also.

Is it no conventional ?
I can replace it with a simple TODO instead

src/renderer/apiFetch.ts
panaC: 626b2d8

returnType doesn't works on compilation

src/renderer/components/catalog/PublicationAddButton.tsx
panaC: aae238c

src/renderer/apiFetch.ts
panaC: I like this both names !

what is your proposal ?

src/renderer/components/dialog/DeleteOpdsFeedConfirm.tsx
panaC: 9688474

src/renderer/components/opds/Entry.tsx
panaC: 9f61dd0

src/renderer/apiFetch.ts
panaC: The `clean` redux api data is called before each resolve or reject. In this case the data api in redux state is all the time "clean". When the data is getting back by the promise apiFetch the redux api data state will not useful.

It's false : `because "clean" is only invoked in the very specific case of moduleId === meta.moduleId && methodId === meta.methodId && state.api.data[requestId].`

Sorry if i'm not answering your question, or if i skip anything else

src/renderer/components/opds/BrowserResult.tsx
panaC: It's automatically handle with apiFetch

src/renderer/apiFetch.ts
panaC: c4f51c0

src/renderer/apiRefresh.ts
panaC: c4f51c0

src/renderer/apiRefresh.ts
panaC: 3552ed2

src/renderer/apiFetch.ts
panaC: ee0471c

src/renderer/components/utils/menu/AccessibleMenu.tsx
danielweck: `Event.target` is a generic DOM property, which does not only apply to HTML (also SVG, XML, etc.). However, in Thorium we know that the UI is implemented in HTML markup (with the odd SVG image here and there), so we can safely cast the appropriate type for better static / compile-time checking.
https://developer.mozilla.org/en-US/docs/Web/API/Event/target
https://developer.mozilla.org/en-US/docs/Web/API/Event

src/renderer/components/opds/Header.tsx
danielweck: `TODO` and `FIXME` are commonly-used comment tokens to mark code locations that need addressing.
https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight
https://marketplace.visualstudio.com/items?itemName=TzachOvadia.todo-list
https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree
https://www.jetbrains.com/help/idea/using-todo.html
https://docs.microsoft.com/en-us/visualstudio/ide/using-the-task-list?view=vs-2019
etc.

src/renderer/components/utils/menu/AccessibleMenu.tsx
panaC: bfea739

thank you Daniel 👍

src/renderer/apiFetch.ts
danielweck: I do not have a concrete proposal, because my understanding of the scope/responsibility of the two "api" functions is only superficial. I am merely highlighting the fact that the term "fetch" is overloaded. I can see that this is really about "store" ("dispatch" / "subscribe") and "state" ... but I am not sure how to formulate / abbreviate the function contract into a concise meaningful name.

src/renderer/components/utils/menu/AccessibleMenu.tsx
panaC: d7e13e2

👍

src/renderer/components/opds/Header.tsx
panaC: thank you daniel

src/renderer/apiFetch.ts
panaC: It's more over ipc communication with the usage of redux state

src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,80 @@
...
+                    resolveSubscribe(request.result);
+                }
+
+                // handle promise<void>
+                timeoutId = window.setTimeout(() => reject("API Timeout"), 5000);
+            });
+        });
+
+        // The linter doesn't accept .finaly(). Why ? Is it a Bug ?
+        // tslint:disable-next-line: no-floating-promises
```
danielweck: When I delete the comments above, `npm run lint` and `npm run build` complete successfully. Could you please double-check at your end? (prerequisite: make sure to use the latest `package-lock.json` from the develop branch: `rm -rf node_modules && npm install && npm run __postinstall`)

Longer analysis:

Thorium's current Chromium version is '66.0.3359.181', and NodeJS is version `10.2.0` (see `process.versions` debug console output when the app launches).

* In my system shell: `node --version` => `v10.16.3`, and `node -e "Promise.resolve().finally(() => {console.log('OK')})"` => `OK` (see https://node.green/#ES2018-features-Promise-prototype-finally )

* In Thorium's web inspector console (`renderer` process, Chromium V8 runtime): `Promise.resolve().finally(() => {console.log('OK')})` => `OK` (see https://caniuse.com/#feat=promise-finally )

* In Thorium's `main` process, NodeJS V8 runtime: `Promise.resolve().finally(() => { console.log("OK"); });` => `OK` (see https://node.green/#ES2018-features-Promise-prototype-finally )


However, `Promise.finally` is *normally* not included in Thorium's current TypeScript type definitions (ES6/2015 target + ES2017 lib):
https://github.com/readium/readium-desktop/blob/develop/tsconfig.json#L57-L58
Typically, `es2018.promise` needs to be *explicitly* added to the `lib` configuration: `"lib": ["es2017", "es6", "dom", "dom.iterable", "es2018.promise"]`.

So, I do not understand why Thorium compiles successfully given the current `target` and `lib` configuration. I would expect failure because `es2018.promise` is normally not included by default (!?)

Side note - the NodeJS v10 typings also limit a number of features to match the specific NodeJS runtime API of Electron v3:
https://github.com/readium/readium-desktop/blob/54685000999c2802fa5f3a068f2a1f535eff5fd8/package.json#L226

Other references:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

src/renderer/components/opds/Header.tsx
danielweck: Sorry, I do not understand your last message. Does this mean you want to replace the `code:cssincode` (etc.) notation with `TODO` / `FIXME` in *this PR*, or that you wish to implement the correction in *another PR*?

src/renderer/apiRefresh.ts
```diff
@@ -0,0 +1,47 @@
...
+        if (!apiLastSuccess || apiLastSuccess.date <= lastSuccessDate) {
+            return;
+        }
+
+        // New api success
+        lastSuccess = apiLastSuccess;
+
+        const meta = apiLastSuccess.action.meta.api;
+        if (pathArrayToRefresh.find((path) => {
+            const splitPath = path.split("/");
```
danielweck: For consistency with the other code occurrence (to avoid confusion with `const path = require("path");`), replace `path` with `apiPath`?

src/renderer/apiFetch.ts
danielweck: Behind the scenes, Electron's IPC asynchronous events are used, but this is an "implementation detail" inside the underlying plumbing / system guts (in other words, we should not use "ipc" in the publicly-exposed function names).

From the perspective of the function callers, a Redux action corresponding to a well-defined "API" is dispatched, and a promised result is awaited. In that light, would it make sense to replace `apiFetch()` with `apiDispatch`? (I think so, because any Redux programmer would understand the notion of "dispatching" actions)

`apiRefresh()` does not actually "refresh" anything (imperatively-speaking). Instead, it is really about triggering calls to `apiDispatch()` based on the last successful API action, by listening / subscribing to store changes. So would it make sense to replace `apiRefresh()` with `apiSubscribe()`? (I think so, because the function returns the opposite `unsubscribe` function)

src/renderer/apiFetch.ts
danielweck: I am not sure about exactly what the role of "clean" is (generally-speaking, in the "api" IPC / Redux functionality).

But when I look at the pre-refactoring usage of `cleanData()`:

https://github.com/readium/readium-desktop/blob/6cbd90399ae4c9bcefed0880c8ced776bf57fa5b/src/renderer/components/utils/hoc/api.tsx#L116-L122

...I see that `cleanData()` is automatically called in `componentWillUnmount` of the HOC (which matches the lifecycle of the actual `withApi()` component), and also explicitly called in `browseOpds()`:

https://github.com/readium/readium-desktop/blob/54685000999c2802fa5f3a068f2a1f535eff5fd8/src/renderer/components/opds/BrowserResult.tsx#L105-L118

So, I am trying to understand the post-refactoring lifecycle, to see if this PR introduces breaking changes because of how/when the "clean" functionality is invoked, compared with the old HOC model.

danielweck: > It's false : `because "clean" is only invoked in the very specific case of moduleId === meta.moduleId && methodId === meta.methodId && state.api.data[requestId].`

Well, I am just quoting the exact code from this PR (conditional statement) :)

https://github.com/readium/readium-desktop/blob/6cbd90399ae4c9bcefed0880c8ced776bf57fa5b/src/renderer/apiFetch.ts#L50-L52

src/renderer/apiFetch.ts
panaC: the apiFetch function doesn't works like this. apiFetch ask a request with a path and get back the data, whereas dispatch, update the state with actual data. It's not the same thing. I don't think apiDispatch is a great name, at all. But in fact apiRefresh may be called apiSubscribe. The name refresh allow to perfect understand what this function execute the callback only when any api request match with api name array in first parameter. It's my opinion.

src/renderer/apiFetch.ts
panaC: what is false is `very specific case`in `because "clean" is only invoked in the very specific case`. sorry,

panaC: before, `cleanData` was called when the component was unmounted or when we called it with `this.props.cleanData` .. now the cleanData is automatically called after each request (apiFetch).
see the redux debugger in web-dev window with `npm run start:dev`

panaC: we have no side effect. now apiFetch handle the data like an window.fetch method

src/renderer/apiFetch.ts
panaC: i don't know, strange behaviour

b65f342

src/renderer/apiRefresh.ts
panaC: 10cee11

src/renderer/apiFetch.ts
danielweck: Thank you for the explanation. I trust your judgement, so if you are satisfied with the refactored lifecycle based on your analysis and your tests, then I am happy to merge the proposed change.

src/renderer/apiFetch.ts
danielweck: > the apiFetch function doesn't works like this
Once again, I am just quoting the code, verbatim:
https://github.com/readium/readium-desktop/blob/6cbd90399ae4c9bcefed0880c8ced776bf57fa5b/src/renderer/apiFetch.ts#L27-L34
Anyway, I am happy to follow your guidance, this is why I asked you in the first place: to get your expert opinion.
Let's close this matter.

src/renderer/apiFetch.ts
panaC: I'm satisfied of this change..
Thank you Daniel

But before merge I will continue to test all application. To prevent regression bug

Have you make some specific tests ? may be on reader.

src/renderer/components/opds/Header.tsx
panaC: https://github.com/readium/readium-desktop/pull/696/commits/25ec2d25ee457dcf766a51a0cd0e42a5be7ca17b

panaC:
d379c42 fix https://github.com/readium/readium-desktop/pull/696#issuecomment-534678709

src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,78 @@
...
+
+                // New api success
+                lastSuccess = apiLastSuccess;
+
+                const meta = apiLastSuccess.action.meta.api;
+                if (moduleId === meta.moduleId && methodId === meta.methodId && state.api.data[requestId]) {
+                    const request = Object.assign({}, state.api.data[requestId]);
+                    store.dispatch(apiActions.clean(requestId));
+                    if (request.resultIsReject) {
+                        rejectSubscribe(request.result);
```
danielweck: Missing `return`?

src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,78 @@
...
+                    }
+                    resolveSubscribe(request.result);
```
danielweck: Return to guarantee avoiding the timeout?

src/renderer/apiFetch.ts
```diff
@@ -0,0 +1,78 @@
...
+                }
+
+                // handle promise<void>
+                timeoutId = window.setTimeout(() => reject("API Timeout"), 5000);
```
danielweck: `rejectSubscribe()` otherwise finally is not called!

danielweck: Also nullify `timeoutId`

src/renderer/apiFetch.ts
danielweck: Just for tracking purposes:
`apiRefresh()` becomes `apiSubscribe()` (because `Unsubscribe` is returned)

panaC:
I have tested this PR. For me it is okay for merge.. What do you think @danielweck, let's go for a merge into develop !

panaC:
I have a problem on browser

panaC:
Ready to merge @danielweck

danielweck:
Ok, let's merge!

danielweck:
A couple more code comment removal in this follow-up PR:
https://github.com/readium/readium-desktop/pull/747

PR #716, opened 2019-09-23
Fixes #685 : fix default values of reader settings

### LINKED ISSUE #685: app first start (default settings) => reader view options are not initialized correctly (store sync problem?)
ThoriumDefaults1
ThoriumDefaults2

I noticed that changing the display setting from scroll to paginated results in the font size slider being set to the correct default value (see screenshots below), but the colour theme radio buttons are not updated (the tick is missing) ... in fact, nothing happens when I click on the default neutral option, only when I select the other ones.

ThoriumDefaults3
ThoriumDefaults4

Scarsniik:


danielweck:
Hello, would you mind removing the changes to `package-lock.json`? Thanks! :)

src/renderer/components/reader/ReaderOptions.tsx
```diff
@@ -100,58 +98,57 @@ export class ReaderOptions extends React.Component<Props> {
...
+                    </label>
                 </div>
-            </>
+            </div>
         );
     }
 
     private textContent() {
         const {__, settings} = this.props;
+        console.log(this.props.indexes.fontSize);
```
danielweck: Please use `debug()` or remove the `console.log()` statement.

src/renderer/components/reader/ReaderOptions.tsx
```diff
@@ -245,7 +242,7 @@ export class ReaderOptions extends React.Component<Props> {
                             name="alignment"
                             type="radio"
                             onChange={(e) => this.props.handleSettingChange(e, "align", "left")}
-                            checked={this.props.settings.align === "left"}
+                            checked={settings.align === "left"}
```
danielweck: This data structure is actually strongly-typed, via `ReaderConfig` (for example, "left" is recognized as a valid enum value for `align`). So, whilst you are editing code in `ReaderOptions.tsx`, could you please change `settings: any;` to `settings: ReaderConfig;` in `interface Props extends TranslatorProps` at the top of the file? Thank you! (`import { ReaderConfig } from "readium-desktop/common/models/reader";`)

PS: Pierre is working in separate PRs for further removal of `any` and better typing across the board.

src/renderer/components/reader/ReaderOptions.tsx
```diff
@@ -261,7 +258,7 @@ export class ReaderOptions extends React.Component<Props> {
                             name="alignment"
                             type="radio"
                             onChange={(e) => this.props.handleSettingChange(e, "align", "justify")}
-                            checked={this.props.settings.align === "justify"}
+                            checked={settings.align === "justify"}
```
danielweck: Replace `"justify"` with `textAlignEnum.justify`
(`import { textAlignEnum } from "@r2-navigator-js/electron/common/readium-css-settings";`)

src/renderer/components/reader/ReaderOptions.tsx
danielweck: Replace `"left"` with `textAlignEnum.left`
(`import { textAlignEnum } from "@r2-navigator-js/electron/common/readium-css-settings";`)

src/renderer/components/reader/ReaderOptions.tsx
```diff
@@ -100,58 +98,57 @@ export class ReaderOptions extends React.Component<Props> {
...
+                    <input
                         id={"radio-" + themeType.Without}
                         type="radio"
                         name="theme"
                         onChange={() => this.handleChooseTheme(themeType.Without)}
-                        {...(!this.props.settings.sepia && !this.props.settings.night
-                        && {checked: true})}
-                        />
-                        <label htmlFor={"radio-" + themeType.Without}>
-                            {this.props.settings.predefined && <SVG svg={DoneIcon} ariaHidden/>}
```
danielweck: I think the line of code `values.predefined = neutral;` must be removed too, in the `handleChooseTheme()` function. Is that correct?

danielweck: Any thoughts about `predefined`? Isn't this completely unused now? (I could not find any references using text search in the Thorium codebase)

Scarsniik: I already removed it

danielweck: My apologies. Because of the Git "forced pushes" I missed the commit. Thank you! :)

PR #733, opened 2019-09-25
Fixes #462 : add pagination in opds feed result

### LINKED ISSUE #462: OPDS feeds, implement pagination
Many OPDS feeds are paginated. We have the choice between two implementations:
1- simple Next / Previous links
2- infinite scroll.

The choice will be based on the difficulty of implementation and difficulty of making it accessible.

Scarsniik:


src/main/converter/opds.ts
```diff
@@ -153,13 +155,20 @@ export class OpdsFeedViewConverter {
...
         if (feed.Publications) {
             // result page containing publications
             type = OpdsResultType.PublicationFeed;
             publications = feed.Publications.map((item) => {
                 return this.convertOpdsPublicationToView(item);
             });
+
+            if (feed.Links) {
+                nextPageUrl = oc(feed.Links.find((link) => link.Rel[0] === "next")).Href(undefined);
+                previousPageUrl = oc(feed.Links.find((link) => link.Rel[0] === "previous")).Href(undefined);
```
danielweck: Strictly-speaking in Atom XML `link@rel` is optional (even though in practice OPDS links always seem to provide the link relation), so I would check `link.Rel` for non-undefined before accessing the array `link.Rel[0]`.
https://tools.ietf.org/html/rfc4287#section-4.2.7

danielweck: An OPDS2 link can carry multiple relations, such as `previous` and `first` (or `last` and `next`) in no particular order. So I would not only check for `link.Rel[0]`, but I would instead implement a lookup routine such as:
`=> link.Rel && link.Rel.includes("previous")`
https://drafts.opds.io/opds-2.0#3-pagination

src/renderer/components/opds/BrowserResult.tsx
```diff
@@ -24,14 +24,20 @@ import MessageOpdBrowserResult from "./MessageOpdBrowserResult";
 
 interface BrowserResultProps extends RouteComponentProps, ApiProps, TranslatorProps {
     url: string;
-    result?: THttpGetOpdsResultView | string;
+    result?: THttpGetOpdsResultView;
```
danielweck: @panaC what do you think of this code change?

src/renderer/components/opds/EntryPublicationList.tsx
```diff
@@ -16,12 +16,19 @@ import ListView from "readium-desktop/renderer/components/utils/ListView";
 import Loader from "readium-desktop/renderer/components/utils/Loader";
 import { RootState } from "readium-desktop/renderer/redux/states";
 
+import PageNavigation from "./PageNavigation";
+
 interface IProps extends RouteComponentProps {
     publications: OpdsPublicationView[];
+    goto: (url?: string) => void;
+    nextUrl: string;
+    previousUrl: string;
```
danielweck: If I am not mistaken, `previousUrl` and `previousUrl` can be `undefined`.

src/renderer/components/opds/PageNavigation.tsx
```diff
@@ -0,0 +1,47 @@
...
+import * as ArrowRightIcon from "readium-desktop/renderer/assets/icons/baseline-arrow_forward_ios-24px.svg";
+import * as ArrowLeftIcon from "readium-desktop/renderer/assets/icons/baseline-arrow_left_ios-24px.svg";
+import SVG from "readium-desktop/renderer/components/utils/SVG";
+
+import * as styles from "readium-desktop/renderer/assets/styles/opds.css";
+
+interface Props extends TranslatorProps {
+    goto: (url?: string) => void;
+    nextUrl: string;
+    previousUrl: string;
```
danielweck: If I am not mistaken, `previousUrl` and `nextUrl` can be `undefined`.

src/renderer/components/opds/BrowserResult.tsx
```diff
@@ -140,6 +151,10 @@ export class BrowserResult extends React.Component<BrowserResultProps, null> {
             return newUrl;
         }
     }
+
+    private goto(url?: string) {
```
danielweck: I don't think `url` should be an optional (i.e. possibly `undefined`) function argument. In fact, callers of the `goto()` function ensure it is never called with a falsy value.

src/renderer/components/opds/EntryPublicationList.tsx
```diff
@@ -16,12 +16,19 @@ import ListView from "readium-desktop/renderer/components/utils/ListView";
 import Loader from "readium-desktop/renderer/components/utils/Loader";
 import { RootState } from "readium-desktop/renderer/redux/states";
 
+import PageNavigation from "./PageNavigation";
+
 interface IProps extends RouteComponentProps {
     publications: OpdsPublicationView[];
+    goto: (url?: string) => void;
```
danielweck: I don't think `url` should be an optional (i.e. possibly `undefined`) function argument. In fact, callers of the `goto()` function ensure it is never called with a falsy value.

src/renderer/components/opds/PageNavigation.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { TranslatorProps, withTranslator } from "readium-desktop/renderer/components/utils/hoc/translator";
+
+import * as ArrowRightIcon from "readium-desktop/renderer/assets/icons/baseline-arrow_forward_ios-24px.svg";
+import * as ArrowLeftIcon from "readium-desktop/renderer/assets/icons/baseline-arrow_left_ios-24px.svg";
+import SVG from "readium-desktop/renderer/components/utils/SVG";
+
+import * as styles from "readium-desktop/renderer/assets/styles/opds.css";
+
+interface Props extends TranslatorProps {
+    goto: (url?: string) => void;
```
danielweck: I don't think `url` should be an optional (i.e. possibly `undefined`) function argument. In fact, callers of the `goto()` function ensure it is never called with a falsy value.

src/common/views/opds.ts
```diff
@@ -52,6 +52,8 @@ export interface OpdsResultView {
     navigation?: OpdsLinkView[];
     publications?: OpdsPublicationView[];
     searchUrl?: string;
+    nextPageUrl?: string;
+    previousPageUrl?: string;
```
danielweck: TODO: implement support for `first` and `last` link relations as well?

src/renderer/components/utils/GridView.tsx
```diff
@@ -46,4 +54,8 @@ export default class GridView extends React.Component<GridViewProps, undefined>
             </div>
         );
     }
+
+    private goOnTop() {
```
danielweck: Minor verbiage correction: `goToTop()` or `scrollToTop()`.

src/renderer/components/opds/BrowserResult.tsx
panaC: result may technically contains an string error message.

src/common/views/opds.ts
Scarsniik: Do you have opds feed with first and last link ?

src/renderer/components/opds/BrowserResult.tsx
Scarsniik: I deleted it because result was never used as a string but maybe I missed something

src/common/views/opds.ts
danielweck: No.
But the `first` and `last` link relations are in the OPDS/Atom specifications:
https://tools.ietf.org/html/rfc5005#section-3
https://drafts.opds.io/opds-2.0#3-pagination
https://tools.ietf.org/html/rfc5988#section-6.2.2

danielweck: The Feedbooks OPDS feed has `previous` / `next` pagination, but not `first` / `last`, as far as I can tell:
http://readium2.herokuapp.com/opds-v1-v2-convert/https%3A%2F%2Fcatalog.feedbooks.com%2Fpublicdomain%2Fbrowse%2Fen%2Ftop.atom%3Fcat%3DFBFIC009000%26page%3D2

src/renderer/components/opds/PageNavigation.tsx
```diff
@@ -0,0 +1,73 @@
...
+                        {__("opds.previous")}
+                    </button>
+                : <div/> }
+                { nextUrl ?
+                    <button onClick={() => goto(nextUrl)}>
+                        {__("opds.next")}
+                        <SVG svg={ArrowRightIcon} />
+                    </button>
+                : <div/>}
+            </div>
```
danielweck: Should this also display OPDS `metadata` `numberOfItems` and `itemsPerPage`, to give the user more pertinent pagination information?

danielweck: Example: http://readium2.herokuapp.com/opds-v1-v2-convert/https%3A%2F%2Fcatalog.feedbooks.com%2Fpublicdomain%2Fbrowse%2Fen%2Ftop.atom%3Fcat%3DFBFIC009000%26page%3D2

src/renderer/components/opds/PageNavigation.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { TranslatorProps, withTranslator } from "readium-desktop/renderer/components/utils/hoc/translator";
+
+import * as ArrowRightIcon from "readium-desktop/renderer/assets/icons/baseline-arrow_forward_ios-24px.svg";
+import * as ArrowLeftIcon from "readium-desktop/renderer/assets/icons/baseline-arrow_left_ios-24px.svg";
+import SVG from "readium-desktop/renderer/components/utils/SVG";
+
+import * as styles from "readium-desktop/renderer/assets/styles/opds.css";
+
+interface Props extends TranslatorProps {
+    goto: (urld: string) => void;
```
danielweck: `urld` typo.

src/renderer/components/opds/PageNavigation.tsx
```diff
@@ -0,0 +1,73 @@
...
+                : <div/>}
+            </div>
+        );
+    }
+
+    private handleKeyDown(e: KeyboardEvent) {
+        const { goto, nextUrl, previousUrl } = this.props;
+        const withModifierKeys = e.shiftKey && e.ctrlKey;
+        if (withModifierKeys) {
+            if (e.key === "ArrowLeft") {
```
danielweck: I would suggest filtering out falsy values with `if (previousUrl && e.key === "ArrowLeft")` (because `goto()` takes a non-undefined `url` argument).

PS: this coding "error" would normally be automatically detected at static compile time with TypeScript `strictNullChecks` ... but we cannot activate this yet (we will soon! :)

src/renderer/components/opds/PageNavigation.tsx
```diff
@@ -0,0 +1,73 @@
...
+                goto(previousUrl);
+            } else if (e.key === "ArrowRight") {
```
danielweck: Same recommendation here: filter-out falsy values with `if (nextUrl && e.key === "ArrowRight")` (because `goto()` takes a non-undefined `url` argument).

src/renderer/components/opds/BrowserResult.tsx
```diff
@@ -140,6 +151,12 @@ export class BrowserResult extends React.Component<BrowserResultProps, null> {
             return newUrl;
         }
     }
+
+    private goto(url: string) {
+        if (url) {
```
danielweck: `url` is "guaranteed" to be non-null/non-undefined (well, assuming TypeScript `strictNullChecks` compiler settings), so I would remove the `if (url) {}` conditional statement here.

src/renderer/components/opds/BrowserResult.tsx
danielweck: When `resultIsReject` is `true` (promise.catch() or async try/catch rejection), then `result` is a `string` corresponding to the error message. Otherwise `result` is a proper `THttpGetOpdsResultView` object (promise.then() resolve).

src/renderer/components/opds/BrowserResult.tsx
Scarsniik: No it's not guaranteed. When the user use shortkeys, the function is called even if the url is undefined

Scarsniik: Oh ok, I just read this one before reply, I'll do as you say for the shortkeys

src/renderer/components/opds/PageNavigation.tsx
```diff
@@ -0,0 +1,85 @@
...
+                { lastPage ?
+                    <button onClick={() => goto(lastPage)}>
+                        {__("opds.lastPage")}
+                    </button>
+                : <div/> }
+            </div>
+        );
+    }
+
+    private handleKeyDown(e: KeyboardEvent) {
```
danielweck: Because `keydown` events can fired in rapid succession, this can be problematic with the async/promise calls underlying to the implementation of `goto()` (`props.browse()`). Therefore, I think calls to `handleKeyDown()` should be throttled (not debounced).

src/renderer/components/opds/PageNavigation.tsx
Scarsniik: I just added it in my last commit

src/renderer/components/opds/PageNavigation.tsx
Scarsniik: I did it in my last commit : navigation shortkeys can be used every 2 seconds or when the page is refreshed

danielweck:
Hello Leo, time to sync your PR branch with `develop`! :)
(code refactoring ... probably a few conflicts to solve)

danielweck:
Should the "page" number be located between the previous/next arrow buttons?

ThoriumOPDSPrevNext

llemeurfr:
> Should the pagination indicator be located between the previous/next arrow buttons?

@danielweck I don't think so. I'd like more "air" btw the buttons and the "page indication" -> have it justified on the right of the screen, with a proper margin. I requested that to Léo.

Scarsniik:
Yes I have to do that but I wanted to finish what I'm doing before that. It juste take me more time than I thought

src/renderer/components/opds/PageNavigation.tsx
```diff
@@ -0,0 +1,131 @@
...
+import * as styles from "readium-desktop/renderer/assets/styles/opds.css";
+
+interface Props extends TranslatorProps {
+    goto: (url: string) => void;
+    urls: OpdsResultUrls;
+    page: OpdsResultPageInfos;
+}
+
+class PageNavigation extends React.Component<Props> {
+    private currentPage: number = 1;
```
danielweck: Ah, *strictly-speaking* starting at `1` is not correct. *In practice* though, this "works" (read below for explanation).

OPDS feed / Atom "pagination" metadata only provides `numberOfItems` and `itemsPerPage`, but not "currentPage". In other words, the pagination context is incomplete, so we have to infer it in order to display meaningful information to the end-user ... but this is not 100% reliable.

PS: also note that the URLs of link `rel` = `previous` or `next` (and also Atom's `first` or `last`) are not *templated* URLs (at least not in OPDS < v2). So, for example the Feedbooks `?page=2` query parameter is totally arbitrary, it could be `/page/2`, etc.)

Consequently, it is possible (right now, in Thorium) to import an OPDS feed that is *not* at "page" `1`. For example:
https://catalog.feedbooks.com/publicdomain/browse/en/top.atom?cat=FBFIC009000&page=2
=>
http://readium2.herokuapp.com/opds-v1-v2-convert/https%3A%2F%2Fcatalog.feedbooks.com%2Fpublicdomain%2Fbrowse%2Fen%2Ftop.atom%3Fcat%3DFBFIC009000%26page%3D2

...resulting in incorrect "currentPage" information displayed to the user (see screenshot).

So, although *strictly-speaking* the current implementation of "curentPage" (i.e. always starts at `1`) is incorrect, like I said before, *in practice* this is an acceptable hack for a boundary condition that is unlikely to occur, because we can expect users *not* to import OPDS feeds part-way through a pagination sequence.

Thoughts?

ThoriumOPDSPagination

danielweck:
Bug with "nextPage" not being disabled/hidden, when it should:

ThoriumNextOPDS1

ThoriumNextOPDS2

danielweck:
Addendum to the above screenshots:
http://readium2.herokuapp.com/opds-v1-v2-convert/https%3A%2F%2Fcatalog.feedbooks.com%2Ffeatured%2Fen.atom
* `numberOfItems` = `43`
* `itemsPerPage` = `50`

...yet there is a link `rel` = `next`, with `type` = `application/atom+xml;profile=opds-catalog;kind=acquisition`

http://readium2.herokuapp.com/opds-v1-v2-convert/https%3A%2F%2Fcatalog.feedbooks.com%2Fstore%2Fbrowse%2Fen%2Fhomepage_selection.atom%3Fpage%3D2

...so is this a bug in the Feedbooks OPDS?

To compare with a legit paginated OPDS feed:

http://readium2.herokuapp.com/opds-v1-v2-convert/https%3A%2F%2Fcatalog.feedbooks.com%2Fpublicdomain%2Fbrowse%2Fen%2Ftop.atom%3Fpage%3D2

danielweck:
TODO: there is currently no progress indicator ("spinner") when clicking on the `previous`/`next` or `first`/`last` arrows, so for a few milliseconds it looks like nothing is happening when the user activates a button.

Scarsniik:
For the "Next" button, Laurent said me that it was a prblem from the feed and I did'nt have to fix it.

I don't really know how to handle the page counter. Is "page = 2" in the url a standard?

src/renderer/components/opds/BrowserResult.tsx
```diff
@@ -114,15 +124,17 @@ export class BrowserResult extends React.Component<BrowserResultProps, IState> {
         </div>;
     }
 
-    private browseOpds() {
-        const { url, location } = this.props;
+    private browseOpds(url: string) {
+        this.setState({browserResult: undefined});
```
panaC: sorry to take the place to Daniel for reviewing. But is it compulsory to duplicate this line with https://github.com/readium/readium-desktop/pull/733/files#diff-e2c9e19293210a76adb1046d74cdbfd5R145

src/renderer/components/utils/GridView.tsx
```diff
@@ -6,23 +6,32 @@
...
 import * as React from "react";
+import { RouteComponentProps } from "react-router-dom";
 import { PublicationView } from "readium-desktop/common/views/publication";
 import * as styles from "readium-desktop/renderer/assets/styles/publicationView.css";
 import CatalogMenu from "readium-desktop/renderer/components/publication/menu/CatalogMenu";
 import OpdsMenu from "readium-desktop/renderer/components/publication/menu/OpdsMenu";
 import PublicationCard from "readium-desktop/renderer/components/publication/PublicationCard";
 
-interface IProps {
+interface Props extends RouteComponentProps {
```
panaC: Sorry if i don't understand, but why you extends RouteComponentProps ?

src/renderer/components/opds/EntryPublicationList.tsx
```diff
@@ -9,21 +9,28 @@ import * as qs from "query-string";
...
 import { DisplayType } from "readium-desktop/renderer/components/opds/Header";
 import GridView from "readium-desktop/renderer/components/utils/GridView";
 import ListView from "readium-desktop/renderer/components/utils/ListView";
 import Loader from "readium-desktop/renderer/components/utils/Loader";
 import { RootState } from "readium-desktop/renderer/redux/states";
 
+import PageNavigation from "./PageNavigation";
+
 interface IProps extends RouteComponentProps {
-    publications: OpdsPublicationView[] | undefined;
```
panaC: if you test publications above in render, then, the publications can be typed in undefined, no ?

src/renderer/components/opds/EntryPublicationList.tsx
```diff
@@ -9,21 +9,28 @@ import * as qs from "query-string";
...
+    publications: OpdsPublicationView[];
+    goto: (url: string, page: number) => void;
+    urls: OpdsResultUrls;
+    page: OpdsResultPageInfos;
```
panaC: page can be undefined, see src/common/views/opds.ts

src/renderer/components/opds/BrowserResult.tsx
```diff
@@ -114,15 +124,17 @@ export class BrowserResult extends React.Component<BrowserResultProps, IState> {
...
+        const { location } = this.props;
         const { browserResult } = this.state;
         const oldQs = parseQueryString(url.split("?")[1]);
         const search = qs.parse(location.search.replace("?", "")).search;
         let newUrl = url;
-
-        if (search && browserResult && browserResult.isSuccess && browserResult.data.searchUrl) {
-            newUrl = browserResult.data.searchUrl;
+        if (search && browserResult && typeof browserResult === "object"
```
panaC: i think that is a bad merge conflict : `typeof browserResult === "object"` has been remove it

src/renderer/components/utils/GridView.tsx
Scarsniik: I use the location to know if it is a grid or a list display

src/renderer/components/opds/EntryPublicationList.tsx
Scarsniik: I didn't do that on purpose, it's an mistake when I resolved the conflicts.
But is it really usefull to type it with undefined ? A type takes undefined by default, no ?

src/renderer/components/utils/GridView.tsx
panaC: really sorry, i don't see other props utilisation and no hoc injection. How you used it ?

image

src/renderer/components/opds/EntryPublicationList.tsx
panaC: not really, it's more clear to strictly typing our code.
You can see the rule strictNullChecks
https://www.tsmean.com/articles/learn-typescript/strict-null-checks-best-practice/

I'm not a specialist, it's my opinion

src/renderer/components/opds/BrowserResult.tsx
Scarsniik: I don't really understand, your link is pointing to the file differences but not a specific line

panaC: look just above apiAction.

src/renderer/components/utils/GridView.tsx
Scarsniik: Oh, sorry, I checked the wrong file...
It's true, the router props are useless here

src/renderer/components/opds/EntryPublicationList.tsx
danielweck: As you know, I am very much in support of removing `any` (@panaC refactorings are incrementally addressing this problem). Also, I am in favour of eventually introducing proper static / compile-time verification of `undefined` versus `null` usages (i.e. `strictNullChecks` TypeScript option).

Reminder: in Javascript `typeof null === "object"` vs. `typeof undefined === "undefined"`, and in TypeScript both `null` and `undefined` are assignable to `any` (and `undefined` is assignable to `void`).

These distinctions are important because the `readium-desktop` codebase (including the R2 "js" components / NPM package dependencies) must be capable of differentiating `null` object properties, vs. `undefined` / non-existing ones, notably when parsing arbitrary / varied JSON formats. Both 'null' and `undefined` are "falsy", but the our codebase sometimes need to check for a specific type.

I had started to explore the impact of adding `strictNullChecks` support in the `readium-desktop` codebase (see this WIP PR: https://github.com/readium/readium-desktop/pull/475/files ), but it quickly became obvious that this would need to be a gradual refactoring effort.

We started this effort a few weeks ago, and now at every PR we aim to implement stricter typing. Unfortunately the `strictNullChecks` compiler option is not activated yet (otherwise the entire codebase breaks) ... so we must rely on "common-sense", rather than TypeScript error messages.

Here is how the Visual Studio Code development team iterated over a much larger codebase than `readium-desktop`:
https://code.visualstudio.com/blogs/2019/05/23/strict-null

Documentation:
https://www.typescriptlang.org/docs/handbook/compiler-options.html

src/renderer/components/utils/GridView.tsx
danielweck: This comment has been addressed: https://github.com/readium/readium-desktop/pull/733/commits/cbf9576d938d2f91c9a2aef9b9f8305909f6cdc6

src/renderer/components/opds/EntryPublicationList.tsx
danielweck: This comment has been addressed: https://github.com/readium/readium-desktop/pull/733/commits/2fcf478637b8b6d3ef2462ae4a5739c36cad5aa7

src/renderer/components/opds/BrowserResult.tsx
danielweck: This comment has been addressed: https://github.com/readium/readium-desktop/pull/733/commits/2fcf478637b8b6d3ef2462ae4a5739c36cad5aa7

src/renderer/components/opds/EntryPublicationList.tsx
danielweck: This comment has been addressed: https://github.com/readium/readium-desktop/pull/733/commits/2fcf478637b8b6d3ef2462ae4a5739c36cad5aa7

src/renderer/components/opds/BrowserResult.tsx
danielweck: This comment has been addressed: https://github.com/readium/readium-desktop/pull/733/commits/e5ff8aa507a6d402750fbe186fd264565f3c0ad1

src/renderer/components/opds/PageNavigation.tsx
danielweck: This is a "no fix", as explained above (in the concluding paragraph)

src/renderer/components/opds/EntryPublicationList.tsx
panaC: Thank you Daniel, perfect explanation

src/main/converter/opds.ts
```diff
@@ -176,15 +197,29 @@ export class OpdsFeedViewConverter {
...
-            searchUrl: await this.getSearchUrlFromOpds1Feed(feed),
+            urls,
+            page,
         };
     }
 
+    private getUrlFromFeed(feed: OPDSFeed, linkRel: string): string {
+        return oc(feed.Links.find((link) =>
+            link.Rel && link.Rel.includes(linkRel),
+        )).Href(undefined);
```
danielweck: `oc` (`ts-optchain`) is useful for safely chaining "optional" (i.e. possibly `undefined` or `null`) object properties, in a type-strong manner.

However, in this particular case I don't think there is a strong justification for using `oc`, which instead seems to make the code harder to read, and to debug.

What about using "vanilla" coding style:
```
const linkWithRel = feed.Links.find((link) => link.Rel && link.Rel.includes(linkRel));
return linkWithRel ? linkWithRel.Href : undefined;
```

src/main/converter/opds.ts
```diff
@@ -176,15 +197,29 @@ export class OpdsFeedViewConverter {
...
             type,
             publications,
             navigation,
-            searchUrl: await this.getSearchUrlFromOpds1Feed(feed),
+            urls,
+            page,
         };
     }
 
+    private getUrlFromFeed(feed: OPDSFeed, linkRel: string): string {
```
danielweck: The return type of the `getUrlFromFeed()` function should be `string | undefined` (which indicates that there is a possibility that a `link` with the given `rel` has not been found).

src/renderer/components/opds/EntryPublicationList.tsx
danielweck: `strictPropertyInitialization` is a good one too ... but one thing at a time ;)

danielweck:
Bug: `currentResultPage` is not reset between feed browsing (i.e. when switching from one feed to another):

ThoriumOPDSpageCurrent

danielweck:
To reproduce the above bug:
Feedbooks OPDS, go to staff picks, 4th page, then go to new/noteworthy (scroll down to see page number 4, when in should be have been reset to 1)

danielweck:
This PR seems to break the search feature.
Test with the develop branch:
ThoriumSearch1
Test with the PR branch:
ThoriumSearch2

Scarsniik:
This issues are fixed in my last commits

Issue #721, opened 2019-09-23
Bookmarks section: list item's left-aligned icon can be clicked, but does not ativate hyperlink to saved reading location


danielweck:
![ThoriumBookmarkListClick](https://user-images.githubusercontent.com/625201/65444777-af24e680-de28-11e9-8018-3ac0ac1e671e.png)

danielweck:
Suggested fix:
1) either: remove the "click" cursor indicator when mouse hover (which in fact is misleadingly present on the entire list item, not just the icon on the left hand side)
2) or: keep the click cursor, and implement the hyperlink where it is missing (currently only on the thick horizontal line in the middle of the list item)

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/722

PR #734, opened 2019-09-25
add keyboard navigation for bookmarks and  click on th icon

### LINKED ISSUE #722: Bookmarks list: hyperlink activation only with mouse (must implement keyboard support)
The area highlighted by the red rectangle in the screenshot, corresponds to the only two buttons that can be accessed with the keyboard.

ThoriumBookmarkKeyboardAccess

Scarsniik:
Fixes #722 and #721

### DIFF
--- a/src/renderer/assets/styles/reader-app.css
+++ b/src/renderer/assets/styles/reader-app.css
@@ -514,10 +514,29 @@ button {
     font-size: 1rem;
     border-bottom: 1px solid #e5e5e5;
     display: flex;
+    position: relative;
+
+    & form {
+      position: absolute;
+      z-index: 100;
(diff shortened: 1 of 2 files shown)


danielweck:
Remaining problems:
1) the "hand" / "click" mouse cursor appears on the entire bookmark entry, but clicking does nothing (only when moving the mouse on one of the 3 buttons) See original issue description: https://github.com/readium/readium-desktop/issues/721#issuecomment-534182377
2) in "edit" mode, the text input is nested inside the button (`UpdateBookmarkForm`), which is illegal, but more importantly this means that the space bar and the mouse click trigger the `button` (activate the bookmark link)

PR #735, opened 2019-09-26
WIP DO NOT MERGE [fix #608] add opds feed catalog in add feed modal

### LINKED ISSUE #608: Create a meta-catalog of interesting OPDS feeds / design the new modal "add an opds feed"
In the context of the addition of an OPDS feed to the application, we could create a catalog of OPDS feeds, which would be available when users are online (logical when accessing OPDS feeds).
The meta-catalog would have sections (could be per language).

panaC:
I have made no design to implement this feature. just a simple select html element

I saw an html glitch on the first initialisation of component : the name placeholder isn't displayed. Like a space in input field, why ?

fixes #608


src/renderer/components/dialog/OpdsFeedAddForm.tsx
```diff
@@ -20,54 +20,96 @@ interface Props extends TranslatorProps {
...
         };
 
         this.add = this.add.bind(this);
         this.close = this.close.bind(this);
+        this.onSelectOpdsFeedChange = this.onSelectOpdsFeedChange.bind(this);
+    }
+
+    public async componentDidMount() {
+        try {
+            const response = await fetch(this.opdsFeedListServerUrl);
```
danielweck: Note that usage of the `window.fetch` Web API will probably be affected by the activation of `webSecurity` in the `BrowserWindow`. Possible workarounds:
1) ensure adequate CORS headers are added to HTTP responses (server side)
2) because `nodeIntegration` is enabled, switch from `window.fetch` to a NodeJS `request` API

More information:
https://github.com/readium/readium-desktop/issues/736

panaC: I think the CORS on server is enable, I activated it

danielweck: I see `Access-Control-Allow-Origin: *` in the HTTP response, so it should be okay.

src/renderer/components/dialog/OpdsFeedAddForm.tsx
```diff
@@ -20,54 +20,94 @@ interface Props extends TranslatorProps {
...
+    opdsFeedList: {
+        [name: string]: string;
+    } | undefined;
+    selectValue: string;
 }
 
 export class OpdsFeedAddForm extends React.Component<Props, State> {
-    constructor(props: any) {
+    private defaultSelectValue = "undefined";
+    private opdsFeedListServerUrl = "http://thorium-opds.edrlab.org/";
```
danielweck: I would strongly suggest using the secure HTTPS URL (there is no justification for using HTTP nowadays).
(for example, if the renderer process is moved to secure HTTPS in the future, then this non-HTTPS `window.fetch` request will fail)
https://www.eff.org/https-everywhere
:)

scripts/thorium_best_opds_feed/.readme
```diff
@@ -0,0 +1,4 @@
+
+npm install --global surge
+
+surge THIS_FOLDER thorium-reader.surge.sh
```
danielweck: Also ensure `--add EMAIL_ADDRESS` to ensure multiple "admins" can remotely control the contents of the deployed list of OPDS feeds (otherwise, only a single developer can edit this data, which can be problematic if an urgent change is required, and the developer is not reachable).

https://surge.sh/help/adding-collaborators

scripts/thorium_best_opds_feed/opdsfeed.json
```diff
@@ -0,0 +1,40 @@
...
+    "chitanka": "http://chitanka.info/catalog.opds",
+    "gitbook": "https://api.gitbook.com/opds/catalog.atom",
+    "lib rus ec": "http://lib.rus.ec/opds",
+    "zone 4 iphone": "http://www.zone4iphone.ru/catalog.php",
+    "flaschenpost": "https://flaschenpost.piratenpartei.de/catalog/",
+    "anarchist lib": "https://theanarchistlibrary.org/opds",
+    "aozora": "http://aozora.textlive.net/catalog.opds",
+    "prestigio plaza": "http://ebooks.prestigioplaza.com/feed",
+    "bnf.fr (gallica)": "http://gallica.bnf.fr/opds"
+}
```
danielweck: TODO @llemeurfr Curate an acceptable list of OPDS feeds (some of these are a bit, erm ... let's say "dodgy" :)

Original source:
https://github.com/readium/r2-streamer-js/blob/develop/docs/opds.md#a-selection-of-public-opds-feeds
(please contribute additions / modifications there as well, as this long list is great for testing edge-cases, including network errors, incomplete HTTP headers, invalid OPDS format, etc.)

src/renderer/components/dialog/OpdsFeedAddForm.tsx
panaC: https://thorium-opds.edrlab.org/ doesn't works. spoken with Laurent

scripts/thorium_best_opds_feed/.readme
panaC: What is your email address ?

danielweck: `daniel.weck@gmail.com`

src/renderer/components/dialog/OpdsFeedAddForm.tsx
danielweck: The HTTP CORS headers are missing in the root EDRLab URL, so `fetch()` Web API will reject the request.
Workaround: use `request` NodeJS API instead of `window.fetch` (`nodeIntegration` will be removed *at some point*, but we can still use it now).

danielweck: Also note the discrepancy in the context where the HTTP request is created (i.e. renderer processes for library and reader BrowserWindows):
* dev: `http://localhost:8090` (WebPack dev server)
* prod: `file://` (normal Electron URL from the filesystem / ASAR bundle in a packaged app)

danielweck:
Suggestion: use a catalog registry format/JSON syntax similar to the NYPL one:
https://libraryregistry.librarysimplified.org/libraries

PR #737, opened 2019-09-27
Fixes #723 : don't accept empty string for bookmark name

### LINKED ISSUE #723: Bookmarks list: label should not be allowed to be edited to an "empty" value (whitespaces only)
The edit feature allows entering only whitespaces. This should be rejected by the input validator, just like when entering an empty value.

Note that whitespaces are trimmed on the leading and trailing edge, as well as collapsed in the middle ... for *display* only (the actual underlying value retains the whitespaces).

ThoriumBookmarkLabel2

ThoriumBookmarkLabel1

Scarsniik:


src/renderer/components/reader/UpdateBookmarkForm.tsx
```diff
@@ -62,8 +62,12 @@ export class UpdateBookmarkForm extends React.Component<Props, State> {
     private submiteBookmark(e: any) {
         e.preventDefault();
         const { bookmark, updateBookmark } = this.props;
-        bookmark.name = this.inputRef.current.value;
-        updateBookmark(bookmark.identifier, bookmark.publication.identifier, bookmark.locator, bookmark.name);
+        const value: string = this.inputRef.current.value;
+        if (value.length > 0 && value.trim().length > 0) {
+            bookmark.name = value.trim();
```
danielweck: Ah, I have just realized the reason why bookmark names that contain multiple whitespace characters in the middle of the string are displayed as if they only had a single space (and also automatic removal of leading/trailing spaces): this is because of automatic HTML collapsing rules for insignificant whitespaces inside the `div` that renders `bookmark.name` (in `ReaderMenu.tsx`). However, in the database (or any backing store where bookmarks are saved), the name is not normalized, which could be problematic in other contexts. So, I would like to suggest applying the same normalization logic as we use for "tags":
`string.trim().replace(/\s\s+/g, " ")`

I would also like to suggest avoiding calling `value.trim()` more than once (as in the quoted line of code for this comment).

The result is:
```
let value: string = this.inputRef.current.value;
if (value && value.length) {
    value = value.trim().replace(/\s\s+/g, " ");
    if (value.length) {
        bookmark.name = value;
        updateBookmark(bookmark.identifier, bookmark.publication.identifier, bookmark.locator, bookmark.name);
    }
}
```

What do you think?

danielweck:
Hello Leo, time to sync your PR branch with `develop`! :)
(code refactoring ... probably a few conflicts to solve)

src/renderer/components/reader/UpdateBookmarkForm.tsx
danielweck: See: https://github.com/readium/readium-desktop/pull/737/commits/3e8afcda3b7abc7b5497610996194cd43d530031

src/renderer/components/reader/UpdateBookmarkForm.tsx
```diff
@@ -59,15 +59,18 @@ export class UpdateBookmarkForm extends React.Component<Props, State> {
...
-        apiAction("reader/updateBookmark",
-            bookmark.identifier,
-            bookmark.publication.identifier,
-            bookmark.locator,
-            bookmark.name,
-        )
-            .catch((error) => console.error("Error to fetch api reader/updateBookmark", error));
+        const value: string = this.inputRef.current.value;
+        const normalizedValue = value.trim().replace(/\s\s+/g, " ");
+        if (value.length > 0 && normalizedValue.length > 0) {
```
danielweck: Why the test for `value.length > 0` here?

PR #753, opened 2019-09-30
[fix #748] relative vs absolute opds feed url

### LINKED ISSUE #748: OPDS relative paths (images, EPUBs, etc.) fail (URL resolution)
See:
https://github.com/readium/readium-desktop/issues/744

panaC:
fixes #748 

test with http://proxy.flibusta.net/opds
http://cops-demo.slucas.fr/feed.php


src/main/converter/opds.ts
```diff
@@ -24,7 +25,10 @@ import { OPDSLink } from "@r2-opds-js/opds/opds2/opds2-link";
 import { OPDSPublication } from "@r2-opds-js/opds/opds2/opds2-publication";
 
 // Logger
-const debug = debug_("readium-desktop:main#services/lcp");
+const debug = debug_("readium-desktop:main/converter/opds");
+
+const urlPathResolve = (from: string, to: string) =>
+        to && !/^https?:\/\//.exec(to) ? resolve(from, to) : to;
```
danielweck: Edge case: OPDS images with `data://` base64 URLs (I don't have a concrete example off the top of my head, but I remember coming across such usage in a feed I was testing, a long time ago).

`url.resolve(from, to)` silently "ignores" absolute `to` URL arguments, by returning `to`. Verification:
`node -e "console.log(require('url').resolve('https://root.com/test/', 'http://sub'));"`
`node -e "console.log(require('url').resolve('https://root.com/test/', 'data://sub'));"`

That being said, it is better to be explicit in the code, and to be consistent with the existing `http[s]` check. So what about adding a `data:` check as well?
`to && !/^https?:\/\//.exec(to) && !/^data:\/\//.exec(to)`

danielweck: Side note:
The trailing slash in the `from` URL argument of `url.resolve(from, to)` is *very* important. Yet, some servers return the exact same OPDS feed, regardless of the trailing slash. For example:
http://proxy.flibusta.net/opds
vs.
http://proxy.flibusta.net/opds/

SO, depending on which URL the user enters in Thorium when adding an OPDS feed to the catalogs list, relative images/links can fail.

Examples:

`node -e "console.log(require('url').resolve('https://root.com/test/', 'sub'));"`
=>
`https://root.com/test/sub`

vs.

`node -e "console.log(require('url').resolve('https://root.com/test', 'sub'));"`
=>
`https://root.com/sub`

___

Or with the `.` path notation:

`node -e "console.log(require('url').resolve('https://root.com/test/', './sub'));"`
=>
`https://root.com/test/sub`

vs.

`node -e "console.log(require('url').resolve('https://root.com/test', './sub'));"`
=>
`https://root.com/sub`

___

Note that this makes no difference for `..`:

`node -e "console.log(require('url').resolve('https://root.com/test/', '../sub'));"`
=>
`https://root.com/sub`

vs.

`node -e "console.log(require('url').resolve('https://root.com/test', '../sub'));"`
=>
`https://root.com/sub`

panaC: we have just to end a slash at the end ?

panaC: in fact, path.resolve has right and we don't have to add a slash at the end, because otherwise it's a file and not a directory

danielweck: Yep :)
The potential problem of users entering the wrong OPDS URL (for example, copy+paste with or without the trailing slash) cannot be solved reliably by the Thorium code (i.e. we cannot guess whether or not a URL should have a trailing slash).

Just for the sake of completion: one possible workaround would be to detect HTTP 404 (or some other status code / bad response), at runtime, dynamically ... but obviously this is totally overkill, and error-prone. The edge-case does not justify the effort.

danielweck: What are your thoughts about handling `data://` URLs? (see my initial comment in this thread)

panaC: yes, i agree with you from beginning
I will fix that

panaC: 9e73da8

Issue #295, opened 2019-06-14
Themes: the whole reading window should be skined


llemeurfr:
Currently, the sepia and night themes only change the text of the publication, not the margins, not the app UI. 
It will be a much better experience when we manage to change that.

panaC:
related with #83

danielweck:
Duplicate of older issue: https://github.com/readium/readium-desktop/issues/83

PR #755, opened 2019-10-01
changing reader theme skins it

### LINKED ISSUE #83: night/dark mode in ReadiumCSS (EPUB HTML content documents) should be matched in UI

Scarsniik:
Fixes #83 and #295

### DIFF
--- a/src/renderer/assets/icons/open_book.svg
+++ b/src/renderer/assets/icons/open_book.svg
@@ -5,7 +5,7 @@
 	 xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 24 24"
 	 style="enable-background:new 0 0 24 24;" xml:space="preserve">
 <style type="text/css">
-	.st0{stroke:#000000;stroke-width:1.5;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:10;}
+	.st0{stroke:var(--reader-fontColor);stroke-width:1.5;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:10;}
 </style>
 <g transform="translate(0,-952.36218)">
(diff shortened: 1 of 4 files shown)


danielweck:
I searched for the character `#` inside the codebase, with file filter `*.css,*.svg` (using Visual Studio Code). There are many instances of RGB color codes (`#xxyyzz`).
Same remark when searching for `rgb` (e.g. `rgba(0, 0, 0, 0.1)` / `rgb(0, 188, 212)`).

Is that an oversight, or a deliberate design choice?

Scarsniik:
When I need transparency I prefere to use rgba because I think it's more readable but otherwise I use hexa code

danielweck:
Sorry, my original question was not clear: I was in fact wondering why are there so many hard-coded colour values, when the goal is to parametrize the CSS styles with variables / custom properties? (so that we can apply consistent dark / light theming across the entire UI)

Scarsniik:
Oh, ok. It's from the original integration we used at the begining. There was no variable so I didn't used id but in this PR I replaced a lot of hard coded color with variable.

danielweck:
So, you are satisfied with all the hard-coded colours that remain? (personally I do not know for sure, this is why I am asking :)

Just to double-check, could you please run a search for `#` and `rgb` inside the codebase, with the file filter `*.css,*.svg`. Then, if you are satisfied that none of the hard-coded colours need to be changed into dark/light variables, I will be happy to merge this PR based on your judgement (for me, it is quite hard to judge because I do not know the CSS/SVG styles in-depth)

Scarsniik:
Yes, I didn't change some values but I think themes I made are good. The hard coded colors was working with all the themes, it's why I keep them

danielweck:
Thanks, LGTM!

Sarvasri:
BlueFire reader on iOS (and perhaps other platforms) offers sliding controls for text and background color for both light and dark mode. This would be a good example to review. Years ago, Stanza offered the best EPUB reader with great controls. One great advantage of EPUB over PDF is the ability to customize the reading experience. This is so basic, yet it is not offered in many apps. I've been looking for a reader for desktop for a long time for Mac that offers a customizable UI. Thorium has good typography, but minimal user control of the reading experience. There should be a brightness control. Background texture selection would be nice. Colors for text and background for light and dark mode.

PR #757, opened 2019-10-02
[FIX #299] add tag from book

### LINKED ISSUE #299: Process tags from OPDS feeds
Currently, only the personal tags (entered by a user) appear in the bookshelf view. 
The tags imported with OPDS books should also appear there.

panaC:
- many type error commented

fixes #299

### DIFF
--- a/src/common/redux/states/import.ts
+++ b/src/common/redux/states/import.ts
@@ -10,8 +10,11 @@ export interface ImportState {
     downloadSample: boolean;
 }
 
+// FIXME : SHOULD BE linked with OpdsPublicationView (undefined)
+
 export interface ImportOpdsPublication {
     url: string;
     base64OpdsPublication: string;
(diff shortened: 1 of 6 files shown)


danielweck:
This works ... with a caveat: the OPDS-supplied tags can be deleted by the user. Is that a feature, or a bug? (i.e. should the tags be immutable when they are defined by the OPDS provider?)

ThoriumOPDSTags1

ThoriumOPDSTags2

ThoriumOPDSTags3

ThoriumOPDSTags4

panaC:
I think the user may remove the tag in his bookshelf

llemeurfr:
Agreed, this is a bonus for the user but he must be able to delete tags he does not like. Even if a tag may reappear if included in a new ebook he imports later. 

This is a way to attract attention on tags from pods feeds.

PR #758, opened 2019-10-03
Fix for screen readers (NVDA in particular) which do not detect webview "soft" refresh / new DOM in iframe (must be "hard" webview load-from-scratch, including inner links / intra document navigation)

danielweck:


danielweck:
Successfully tested on Windows 10 Pro (latest):
1) hardware device (laptop) with touch screen (this is important, as this has been known to trigger the accessibility tree in Electron/Chromium)
2) software emulation (VirtualBox) => NVDA is very choppy in this configuration, lots of silent pauses

danielweck:
Note that hyperlinking from the TOC (right-hand-side vertical slide-out menu) and with the keyboard shortcuts for spine-level navigation (CTRL+SHIFT LEFT/RIGHT (also ALT if not MacOS)) automatically brings focus to the "main" content area which wraps the EPUB HTML document webview. This means that screen readers can directly dive into the publication content when hyperlinking, with one step ahead (arrow keys, depending on VoiceOver vs. NVDA, etc.)

mutterback:
Hi Daniel, three minor, additional notes from our testers. Happy to file more accurately if this Issue isn't the right thread:

Text does not advance. The program says it has two settings for content, scrollable and paging.  I’d expect with scrollable, the text would automatically advance as I read but it doesn’t with any of the three screen reader’s I tried (JAWS, NVDA and Narrator). What I assume are buttons for paging, named left and right, didn’t change the text with any of the screen readers. [My interpretation of this is that the screen reader is reading text below of the visible text in the window, however the text displayed on the screen does not advance with the screen reader - Matt]

Narrator – When you load a book and turn on Narrator’s scan mode, it starts reading fine.  But if you hit ctrl+end to jump to the bottom of the text and start pressing up arrow, you get no speech from Narrator and the app behaves like it is hung. (only tested Narrator)

Announcements that a new book is imported by eKitabu are not announced in JAWS (only tested JAWS)

danielweck:
Thanks Matt!
Let's break this down into distinct actionable issues.
(I'll report back here in a few minutes)

danielweck:
First issue: https://github.com/readium/readium-desktop/issues/761

danielweck:
Second issue: https://github.com/readium/readium-desktop/issues/762

danielweck:
Third and final issue: https://github.com/readium/readium-desktop/issues/763

PR #767, opened 2019-10-04
Fixes #751 : fixes the percent encoding in opds search breadcrumb

### LINKED ISSUE #751: OPDS search is URL percent-escaped in breadcrumb path
ThoriumOPDSSearchSpaces

Scarsniik:


src/renderer/components/layout/BreadCrumb.tsx
```diff
@@ -56,10 +56,12 @@ class BreadCrumb extends React.Component<IProps> {
                             }}
                             title={item.name}
                         >
-                            {`${item.name} /`}
+                            {`${decodeURI(item.name)} /`}
```
danielweck: What about `title={item.name}` (just above this line)

More general question: why is `item.name` URL/percent-encoded in the first place? Shouldn't the `BreadCrumbItem.name` data *always* be "decoded" (when it corresponds to the terms of a search query)

Scarsniik: I correct it for the title.
The name for the serch is provided by the path data of the route, it's why it's percent encoded

src/renderer/components/layout/BreadCrumb.tsx
```diff
@@ -45,24 +45,26 @@ class BreadCrumb extends React.Component<IProps> {
...
                             key={index}
                             to={{
                                 pathname: item.path,
                                 search: `?displayType=${search.displayType}`,
                                 state: item.state,
                             }}
-                            title={item.name}
+                            title={name}
                         >
-                            {`${decodeURI(item.name)} /`}
```
panaC: did you forget to remove the second decodeURI ?
sorry if i'm wrong, i just look you code

Scarsniik: Oh yes...I checked it twice but apparently the morning is a difficult period. Thank you

Scarsniik: Ah, no it's ok, I did'nt forgot anything

src/renderer/components/layout/BreadCrumb.tsx
danielweck: Yes I understand that "under the hood", a route path "transports" the search criteria information (thus the percent-encoding), but my argument is that the `BreadCrumbItem` object should not reflect the URL implementation detail (in other words: this object should contain data that is "ready to use", no need to decode the URI syntax).

If this is too much work to implement, then please let me know and we will merge the PR fix as it is (and I will file a separate issue to describe the improvement we can make to the implementation)

danielweck:
@Scarsniik said:
> The name for the serch is provided by the path data of the route, it's why it's percent encoded

Yes, I understand that "under the hood", a route path "transports" the search criteria information (thus the percent-encoding), but my argument is that the BreadCrumbItem object should not reflect the URL implementation detail (in other words: this object should contain data that is "ready to use", no need to decode the URI syntax).

If this is too much work to implement, then please let me know and we will merge the PR fix as it is (and I will file a separate issue to describe the improvement we can make to the implementation)

src/renderer/components/layout/BreadCrumb.tsx
```diff
@@ -45,22 +45,26 @@ class BreadCrumb extends React.Component<IProps> {
...
                             to={{
                                 pathname: item.path,
                                 search: `?displayType=${search.displayType}`,
                                 state: item.state,
                             }}
-                            title={item.name}
+                            title={name}
                         >
-                            {`${item.name} /`}
+                            {`${decodeURI(name)} /`}
```
danielweck: `decodeURI` is called twice on `item.name`, see this line of code above `const name = decodeURI(item.name);`

Scarsniik:
I did it like that because there is a lot of file using the breadcrumb and I found it easier to do it just in the breadcrumb file instead of in all the files but I'll change it, it will be better and it's not a big change

Issue #697, opened 2019-09-11
Windows (AppX deployed to Microsoft Store): no "open with" option for EPUB files?


danielweck:
I have another Electron app installed "manually" (not via the Store) which has working EPUB file extension associations (i.e. I can choose the app in the "open with" context menu from the file explorer). However, this does not seem to work with Thorium ... is it because the app is not yet public in the Store?

https://github.com/readium/readium-desktop/blob/9799be240bd18a6dde9bc064b8c762e77953d08b/package.json#L113-L124

panaC:
With appx : "open with" doesn't found thorium and worse it appears only the thorium path in list, without action on double click on it.
With msi : "open with" works it appears on the top list, in main windows the publication path is C:\WINDOWS\reader.exe ??

Both doens't works

danielweck:
Current `electron-builder` config for `win` `target` is `msi`:

https://github.com/readium/readium-desktop/blob/1d46e2f6f108b6172e9b1ed146c094c098d84ba9/package.json#L125-L131

... instead, should probably be `nsis`:

https://www.electron.build/configuration/nsis

panaC:
Previous comment : "With msi : "open with" works it appears on the top list, in main windows the publication path is C:\WINDOWS\reader.exe ??" -> fix by replace msi with nsis

as far as appx : 
when i install thorium from microsoft store no key in registry is added for file association..
It's a big deal.

danielweck:
https://github.com/electron-userland/electron-builder/issues/3718

panaC:
https://github.com/electron-userland/electron-builder/issues/3718

danielweck:
We can (maybe) inject a patch via `afterPack`:
https://www.electron.build/configuration/configuration#afterpack

PR #776, opened 2019-10-09
[FIX] replace msi with nsis in package.json

panaC:
It appears that msi do not register any key in windows registry. The registry key to associate extension with executable launch doesn't works with msi installer but only with nsis.

With nsis,  thorium is added like default software with extension epub, epub3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
     "win": {
       "target": [
         "appx",
-        "msi"
+        "nsis"
       ],
       "icon": "dist/assets/icons/icon.ico"
     },


danielweck:
Related issue / comment:
https://github.com/readium/readium-desktop/issues/697#issuecomment-534040779

panaC:
effectively

do you know why msi doesn't write on registry ? Laurentd said me that msi should do it

danielweck:
No idea. Maybe a `electron-builder` bug? Anyway, NSIS is the recommended target, MSI is legacy.

Issue #496, opened 2019-07-30
Follow-up to (PouchDB / LevelDB / LevelDown): MaxListenersExceededWarning, Error: once called more than once


danielweck:
Follow-up to:
https://github.com/readium/readium-desktop/issues/406

According to feedback from other developers on the PouchDB issue tracker (see https://github.com/pouchdb/pouchdb/issues/7819#issuecomment-516551035 and https://github.com/pouchdb/pouchdb/issues/7825 ), this `package.json` workaround is effective:

```
 "resolutions": { "leveldown": "5.0.2", "levelup": "4.0.2" },
```

As you can see, this forces specific versions of LevelDown and LevelUp, instead of what we are doing now: setting a fixed version `7.0.*` for `pouchdb-adapter-leveldb` (we also freeze `leveldown` to `5.*.*`):

https://github.com/readium/readium-desktop/blob/59cf6aff032f50f2d631642edb308bfcd81034aa/package.json#L172
https://github.com/readium/readium-desktop/blob/59cf6aff032f50f2d631642edb308bfcd81034aa/package.json#L263-L264
https://github.com/readium/readium-desktop/blob/59cf6aff032f50f2d631642edb308bfcd81034aa/src/package.json#L36-L37

danielweck:
For what it's worth, I posted an issue in the LevelDown tracker:
https://github.com/Level/leveldown/issues/686

danielweck:
Ah, a better workaround which allows us to use the latest `leveldown` (`^5.3.0`):
https://github.com/readium/readium-desktop/commit/480dac7a13644ea380b26fa7b181e11c53ae8120
See: https://github.com/Level/leveldown/issues/686#issuecomment-541844695

Tested successfully on MacOS with:
`rm -rf dist && rm -rf node_modules && rm package-lock.json && npm install && npm run __postinstall && npm run start`

danielweck:
Follow-up: the latest `leveldown` version `5.4.0` does not fix the issue, so we are keeping the workaround for now. See:
https://github.com/readium/readium-desktop/commit/3331b92529d9bdd93b064f6a464fa808837347d3
https://github.com/Level/leveldown/issues/686#issuecomment-544602542

PR #792, opened 2019-10-13
Electron v6

danielweck:
Most of the critical changes are in the `r2-navigator-js` component, but all R2 packages have been updated to track the NodeJS version 12 API. This has been tested with LCP in `r2-testapp-js` (recompiled native plugin).


danielweck:
Fixes https://github.com/readium/readium-desktop/issues/496

PR #800, opened 2019-10-18
Fixes #773 : focus button of card when close export or delete dialog

### LINKED ISSUE #773: Popup menu (library view), keyboard focus only returns to menu button with "about this book" (not "delete" and "export"))
ThoriumMenuPopupFocusBack

Scarsniik:


danielweck:
Thank you, the proposed solution works for the "grid view" (cover image thumbnails), but not for the list view. The "export" menu item remains focused after the dialog is closed, but the first two menu items (in the horizontally-laid menu) disappear after the dialog is closed, and the menu "root" button is not focused (instead, keyboard focus moves to the top of the app).

ThoriumMenuFocus

src/main/services/catalog.ts
```diff
@@ -251,11 +253,12 @@ export class CatalogService {
                 mainWindow = window;
             }
         }
-
+        store.dispatch(exportDialogOpen());
```
danielweck: Rather than having usage-specific actions for each call to `dialog.showOpenDialog()` (and specific store properties, like `export`, `publication info`, etc.), I think it would be better to have a single "dialog open/close" action type, with a parameter indicating its usage. This generic design would allow reuse of a single construct in different usage contexts (wherever a modal dialog is open/closed). In fact this could/should be generalized even further to any dialog/popup menu/etc. which needs to return keyboard focus to a calling site.

danielweck: I am proposing to close this PR, which unfortunately has several key conflicts to resolve due to recent codebase evolutions (notably: strongly-typed Redux actions).

danielweck:
Closing this PR, as per the above comments.

PR #819, opened 2019-10-23
[fix #785]add subtitle in opds catalog link

### LINKED ISSUE #785: OPDS link titles (used in feed browser for subsection labels, and breadcrumb) are truncated
The current logic is to discard any text that follows a line break:

https://github.com/readium/readium-desktop/blob/80dbecabf4d8ea453c0917bb590020b32655744d/src/main/converter/opds.ts#L49-L54

http://opoto-ed.org/catalogue/opds
=>
http://readium2.herokuapp.com/opds-v1-v2-convert/http%3A%2F%2Fopoto-ed.org%2Fcatalogue%2Fopds
https://readium2.now.sh/opds-v1-v2-convert/http%3A%2F%2Fopoto-ed.org%2Fcatalogue%2Fopds

Note how an OPDS1 link with `Title` =  `By Author` and `Content` = `All books sorted by author` becomes an OPDS2 link with `title` = `By Author [LINE_BREAK] All books sorted by author` (i.e. naive string concatenation).

We can fix/improve this logic in the `r2-opds-js` converter, but I am not sure how. Do we discard the OPDS1 `link.Content` information (only preserve `link.Title`)? Do we add an extra field to the OPDS2 `link` content model in order to store the `link.Content`?

Another option is to improve the `link.title` parsing + displaying in Thorium. For example `link.title.trim().replace(/\s\s+/g, " ")` in order to collapse insignificant whitespace inside the string. Problem: the breadcrumb labels become long. So, perhaps this display logic should only be used for the subsections labels?

ThoriumOPDSLinkTitle1

ThoriumOPDSLinkTitle2

panaC:
The subtitle is stuck on the arrow, need a little css.

fixes #785

### DIFF
--- a/src/main/converter/opds.ts
+++ b/src/main/converter/opds.ts
@@ -53,12 +53,10 @@ export class OpdsFeedViewConverter {
         const title = titleParts[0].trim();
         const subtitle = titleParts[1] && titleParts[1].trim();
 
-        // FIXME : Does publicationCount works? I never see it in react
         return  {
             title,
             subtitle,
             url: urlPathResolve(url, link.Href),
-            publicationCount: link.Children && link.Children.length,
         };
     }
 
--- a/src/main/converter/opds.ts
+++ b/src/main/converter/opds.ts
@@ -49,14 +49,16 @@ export class OpdsFeedViewConverter {
     public convertOpdsLinkToView(link: OPDSLink, url: string): OpdsLinkView {
         // Title could be defined on multiple lines
         // Only keep the first one
-        let title = link.Title;
-        const titleParts = title.split("\n");
-        title = titleParts[0];
+        const titleParts = link.Title.split("\n").filter((text) => text);
+        const title = titleParts[0].trim();
+        const subtitle = titleParts[1] && titleParts[1].trim();
 
+        // FIXME : Does publicationCount works? I never see it in react
         return  {
             title,
+            subtitle,
             url: urlPathResolve(url, link.Href),
-            publicationCount: (link.Children) ? link.Children.length : null,
+            publicationCount: link.Children && link.Children.length,
         };
     }
 
--- a/src/renderer/components/opds/Entry.tsx
+++ b/src/renderer/components/opds/Entry.tsx
@@ -48,16 +48,6 @@ class Entry extends React.Component<EntryProps, undefined> {
                                 </span> :
                                 <></>
                         }
-                        {
-                            (entry.publicationCount) ?
-                                (
-                                    // FIXME livres in french
-                                    <span className={styles.flux_subtitle}>
-                                        {entry.publicationCount}
-                                    </span>
-                                ) :
-                                (<></>)
-                        }
                         <div className={styles.flux_image}>
                             <SVG svg={ArrowIcon} />
                         </div>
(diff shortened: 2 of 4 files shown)


src/renderer/components/opds/Entry.tsx
```diff
@@ -41,9 +41,11 @@ class Entry extends React.Component<EntryProps, undefined> {
                         to={ route }
                     >
                         <span className={styles.flux_title}>{entry.title}</span>
+                        {entry.subtitle ? <span className={styles.flux_subtitle}>{entry.subtitle}</span> : <></>}
                         {
                             (entry.publicationCount) ?
                                 (
+                                    // FIXME livres in french
```
danielweck: Let's just remove "livres" altogether. No need for a label here, just a number. IMO :)

src/main/converter/opds.ts
```diff
@@ -49,14 +49,16 @@ export class OpdsFeedViewConverter {
     public convertOpdsLinkToView(link: OPDSLink, url: string): OpdsLinkView {
         // Title could be defined on multiple lines
         // Only keep the first one
-        let title = link.Title;
-        const titleParts = title.split("\n");
-        title = titleParts[0];
+        const titleParts = link.Title.split("\n").filter((text) => text);
+        const title = titleParts[0].trim();
+        const subtitle = titleParts[1] && titleParts[1].trim();
```
danielweck: if `subtitle` == `title`, then set `subtitle` to `undefined`, otherwise there are duplicate UI labels:
ThoriumOPDSsubtitle

src/main/converter/opds.ts
```diff
@@ -49,14 +49,16 @@ export class OpdsFeedViewConverter {
...
 
+        // FIXME : Does publicationCount works? I never see it in react
         return  {
             title,
+            subtitle,
             url: urlPathResolve(url, link.Href),
-            publicationCount: (link.Children) ? link.Children.length : null,
+            publicationCount: link.Children && link.Children.length,
```
danielweck: `link.Children` is not used in the OPDS model. `publicationCount` is always zero.

OPDS feeds can provide the Atom "Thread Count" (`ThrCount`) number:
https://readium2.herokuapp.com/opds-v1-v2-convert/https%3A%2F%2Fcatalog.feedbooks.com%2Fcategory%2FFBFIC000000%2Fsub.atom

...but as you can see, this is only available in the parsed XML (left column, OPDS1), there is no equivalent in the OPDS2 JSON conversion (right column).

danielweck: I have asked the question: https://github.com/opds-community/drafts/issues/30

panaC: fix

src/main/converter/opds.ts
panaC: well seen, fixed

src/renderer/components/opds/Entry.tsx
panaC: fixed

src/main/converter/opds.ts
```diff
@@ -53,12 +53,10 @@ export class OpdsFeedViewConverter {
         const title = titleParts[0].trim();
         const subtitle = titleParts[1] && titleParts[1].trim();
 
-        // FIXME : Does publicationCount works? I never see it in react
```
danielweck: Note that I have re-implemented this feature in https://github.com/readium/readium-desktop/commit/a7c90060500e280de5328e458e89c5a55549e0fd

danielweck:
Follow-up: https://github.com/readium/readium-desktop/commit/a7c90060500e280de5328e458e89c5a55549e0fd

PR #836, opened 2019-11-03
replace apiActions and apiSubscribe with reduxConnect only

panaC:
@danielweck what do you think ?


danielweck:
Discussed and approved with @panaC Merging this PR now.

PR #867, opened 2019-12-04
WIP typing router state from daniel work

panaC:


src/renderer/components/catalog/GridTagButton.tsx
```diff
@@ -19,10 +21,10 @@ interface IBaseProps {
 // ReturnType<typeof mapStateToProps>
 // ReturnType<typeof mapDispatchToProps>
 // tslint:disable-next-line: no-empty-interface
-interface IProps extends IBaseProps, RouteComponentProps {
+interface IProps extends IBaseProps, ReturnType<typeof mapStateToProps> {
 }
 
-class GridTagButton extends React.Component<IProps, undefined> {
+class GridTagButton extends React.Component<IProps> {
```
danielweck: Can we please continue to use the exact same consistent notation for `extends React.Component<PROPS, STATE>`, which is to use `{}` for unspecified PROPS (not `any`), and `undefined` for unspecified STATE (not `null` or empty).

src/renderer/components/catalog/Catalog.tsx
```diff
@@ -95,11 +94,12 @@ const mapStateToProps = (state: RootState) => ({
         "publication/updateTags",
         "reader/setLastReadingLocation",
     ]),
+    location: state.router.location,
```
danielweck: Why use `mapStateToProps` instead of `withRouter()` HOC + `RouteComponentProps`? Is this a personal preference, or is there a technical reason?

src/renderer/redux/reducers/index.ts
```diff
@@ -13,25 +13,28 @@ import { importReducer } from "readium-desktop/common/redux/reducers/import";
...
+    combineReducers({
+        i18n: i18nReducer,
+        reader: readerReducer,
+        opds: opdsReducer,
+        win: winReducer,
+        net: netReducer,
+        update: updateReducer,
+        api: apiReducer,
+        dialog: dialogReducer,
+        router: connectRouter<IRouterLocationState>(history),
```
danielweck: Great :)

src/renderer/components/catalog/GridTagButton.tsx
panaC: if you want, but it's not really important, it is `{}` by default

panaC: of course not `any`or `null`

src/renderer/components/catalog/SearchForm.tsx
```diff
@@ -62,18 +62,20 @@ class Search extends React.Component<IProps, undefined> {
...
         const value = this.inputRef?.current?.value;
         if (!value) {
-            this.props.redirect({
-                pathname: "/library/search/all",
-                search: undefined,
+            this.props.history.push({
+                ...this.props.location,
+                ...{
+                    pathname: "/library/search/all",
+                },
```
danielweck: In my tests, and according to the React Router documentation, the `history` API does not handle `state`, thus why I used the `<Redirect/>` React Component (as a HOC, but we can also use it directly inside each implicated component class).
I will have to run tests again with this PR again to find the edge cases where the state (`displayType`) is not preserved across navigations, for example inside OPDS (search and browsing), inside local bookshelf (search and browsing), as well as when navigating across OPDS/local bookshelf.

src/renderer/components/catalog/GridTagButton.tsx
danielweck: It is just for code style consistency, search and replace, etc.

PR #870, opened 2019-12-04
[READY TO MERGE][improve #864] react router

panaC:
extract and improve from #864 

I extracted from #864 what concerns react-router

with this PR, we can in the future not used at all `withRouter` HOC but used instead `react-redux connect` HOC.
this allows a better standardisation of our components


danielweck:
Could you please resolve the conflicts? (since you copied my code from the other PR in order to improve it in this PR, there have been multiple merges into the develop branch)

So, I would like to see this PR supersede my original one (which I have now closed).

I agree that instead of using the React Router `<Redirect/>` component in order to forward state into visited routes (for example, our primary use-case is to maintain the `displayType` Grid / List), there is indeed the cleaner approach (code-wise, not necessarily architecturally, in the React sense) of using `history.push|replace()`. However, contrary to the official documentation (see links below), there seems to be a difference between passing a single location parameter versus a route plus a state. We need to test this further.

Bottom line: it would be nice if we could remove the heavy-handed (and error-prone, and non-typed) URL query parameter parsing and forwarding when transitioning between routes, for `displayType`, search, etc. Instead, using a typed state means the code will be a lot more robust during refactorings, etc. (and easier to read for new developers)

https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/history.md

https://github.com/ReactTraining/history/blob/master/docs/Navigation.md

src/renderer/components/layout/LibraryHeader.tsx
```diff
@@ -108,13 +111,25 @@ class Header extends React.Component<IProps, undefined> {
...
         return (
-            <li className={classNames(...styleClasses)} key={ index }>
-                <Link to={ item.route } replace={true}>
-                    { translate("header." + item.label) }
+            <li className={classNames(...styleClasses)} key={index}>
+                <Link
+                    to={{
+                        ...this.props.location,
+                        ...{
+                            pathname: item.route,
```
panaC: remove spread

panaC:
@danielweck I think the PR is ready, It works very well, the state is preserved all long the application lifecycle

src/renderer/routing.ts
```diff
@@ -37,64 +36,72 @@ export interface ILibrarySearchText {
...
+    } as Route,
     "/opds/browse": {
         // IOpdsBrowse
         path: "/opds/:opdsId/browse/:level/:name/:url",
         exact: true,
         component: Browser,
-    },
-    "/settings/languages": {
+    } as Route,
+    "/settings/languages": Object.freeze<Route>({
```
danielweck: Forgotten `Object.freeze<Route>()` needs to be replaced with `as Route`.

danielweck:
`newRouteUrl` in `PageNavigation.tsx` is broken (to reproduce, test Feedbooks OPDS feed and use "Next" page)

danielweck:
Also, we need to check that the programmatic call of `history.push()` preserves the state (`displayType`) when using left/right arrow keys for previous/next pagination in OPDS feeds:

```
            if (previous && e.key === "ArrowLeft") {
                this.props.history.push(newRouteUrl(location.pathname, next));
            } else if (next && e.key === "ArrowRight") {
                this.props.history.push(newRouteUrl(location.pathname, previous));
            }
```

panaC:
Tanks Daniel, I fixed PageNavigation. It was totally broken

Issue #860, opened 2019-12-02
Allow self-signed certificates in dev mode


iisisrael:
When running in dev mode (`npm run start:dev`), self-signed certificates should be allowed so requests can be made to a dev / test catalog provider.  Currently, such requests result in:
```
[2]   readium-desktop:main#redux/sagas/api Error: self signed certificate
[2]   readium-desktop:main#redux/sagas/api     at TLSSocket.onConnectSecure (_tls_wrap.js:1317:34)
[2]   readium-desktop:main#redux/sagas/api     at TLSSocket.emit (events.js:203:13)
[2]   readium-desktop:main#redux/sagas/api     at TLSSocket._finishInit (_tls_wrap.js:792:8)
[2]   readium-desktop:main#redux/sagas/api     at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:606:12) +41ms
```
with a white screen of death in the application.  Cert .pem is installed locally as a trusted source.

llemeurfr:
To be discussed.

danielweck:
This was fixed some time ago via HTTPS agent z`rejectUnauthorizedz`
https://github.com/edrlab/thorium-reader/blob/5b095fcec33edce32fdc066474954ac4605de91a/src/main/network/http.ts#L234-L237

PR #874, opened 2019-12-04
Allows self-signed certs on OPDS feed requests in dev

iisisrael:
Not sure how kosher / best practice this is, but resolves issue #860 .

### DIFF
--- a/src/common/utils/http.ts
+++ b/src/common/utils/http.ts
@@ -10,6 +10,8 @@ import { JsonMap } from "readium-desktop/typings/json";
 import * as request from "request";
 import { Url } from "url";
 
+import { _NODE_ENV } from "readium-desktop/preprocessor-directives";
+
 type TRequestCoreOptionsRequiredUriUrl = request.CoreOptions & request.RequiredUriUrl;
 type TRequestCoreOptionsOptionalUriUrl = request.CoreOptions & request.OptionalUriUrl;
 
@@ -66,6 +68,7 @@ export async function httpGet<TBody extends JsonMap | string = string , TData = 
             method: "GET",
             encoding: undefined,
             headers,
+            rejectUnauthorized: (_NODE_ENV === "development") ? false : true,
         },
     );
 
--- a/src/main/services/downloader.ts
+++ b/src/main/services/downloader.ts
@@ -20,6 +20,8 @@ import { tmpNameSync } from "tmp";
 import { URL } from "url";
 import * as uuid from "uuid";
 
+import { _NODE_ENV } from "readium-desktop/preprocessor-directives";
+
 type TRequestCoreOptionsRequiredUriUrl = request.CoreOptions & request.RequiredUriUrl;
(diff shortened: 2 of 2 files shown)


src/main/services/opds.ts
```diff
@@ -96,6 +98,7 @@ export class OpdsService {
 
         return httpGet(url, {
             timeout: 10000,
+            rejectUnauthorized: (_NODE_ENV === "development") ? false : true,
```
danielweck: Good idea, although I would suggest adding this in the `httpGet` implementation as a default in `{}`:
https://github.com/readium/readium-desktop/blob/17c866a684fe3ac2ef6dcd46723856fb8315507e/src/common/utils/http.ts#L61-L70

...and also in the download service:

https://github.com/readium/readium-desktop/blob/17c866a684fe3ac2ef6dcd46723856fb8315507e/src/main/services/downloader.ts#L136-L145

This way, this could cover a complete OPDS/LCP/LSD HTTP test environment using self-signed certs.

iisisrael: Okay, can do.  Didn't know how comprehensive it should be, given that `http.ts` is in the util namespace, if that should be expected to be environment-aware.

iisisrael:
Squashed, which removes the original changes.

danielweck:
It has just occurred to me that we should probably use this instead:

`const request = require( 'request' ).defaults({rejectUnauthorized: _NODE_ENV === "development"});`

However, unfortunately this is not a global default that would apply to all instances of `request` in a given Javascript bundle / process (e.g. the Electron main process), which would be ideal because there are other uses of `request` in Readium2 dependencies/components (e.g. LCP LSD register / renew / return interactions, etc.).

So instead this just creates a wrapper to a specific `request` instance that will use the given defaults. For this to work more globally in Thorium's own codebase, we must expose the instance as a const, somewhere. For this to work with underlying Readium2 components, the APIs would need to accept a custom `request` instance to be used instead of a regular `import * as request from "request"` (IIRC).

danielweck:
New file: `src/utils/request.ts`
```
import * as rawRequest from "request";
export const request = rawRequest.defaults({ rejectUnauthorized: _NODE_ENV === "development" });
```


...then in `src/main/services/downloader.ts` and `src/common/utils/http.ts` replace `import * as request from "request";` with `import { request } from "readium-desktop/utils/request";`

What do you think?

danielweck:
I'm not sure how this would work with `request-promise-native` which is used in some cases instead of `request` inside the Readium2 LCP/LSD code.
https://github.com/request/request-promise-native

danielweck:
Closing this PR, as `rejectUnauthorized` is now enabled in DEV mode for `httpGet()` (wrapper for `request()` which calls Node `fetch()`), as well as the special case `request.post()` (OPDS OAuth). Note that the LCP/LSD internal `request()` calls (`request` lib, not a Thorium-specific function wrapper) do not enable `rejectUnauthorized` (see comment above https://github.com/edrlab/thorium-reader/pull/874#issuecomment-562784062 ).

PR #875, opened 2019-12-05
Adds OPDS authentication content type header matching

iisisrael:
Adds check for Content-Type header per the [Authentication for OPDS draft spec](https://drafts.opds.io/authentication-for-opds-1.0#22-content-conformance), 2.2 Content Conformance.


danielweck:
Excellent point!
Thorium currently implements support for OPDS2 authentication (via OAuth access tokens) based on HTTP `401` responses ("Unauthorized") and the `application/opds+json` HTTP header content-type ... but we definitely need to let `application/opds-authentication+json` through as well.
Thanks :)

PR #879, opened 2019-12-09
[fixed #878] navigation history

### LINKED ISSUE #878: Top header tabs (books, catalogs, settings) reset total state of view, does not preserve current OPDS location, search / tags filters, etc.

panaC:
fixed #878


src/renderer/redux/sagas/history.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { LOCATION_CHANGE, LocationChangeAction } from "connected-react-router";
+import { winActions } from "readium-desktop/renderer/redux/actions/";
+import { TLocationRouter } from "readium-desktop/renderer/routing";
+import { all, call, put, take } from "redux-saga/effects";
+
+function* historyWatcher() {
+    while (true) {
+        const result: LocationChangeAction = yield take(LOCATION_CHANGE);
+        yield put(winActions.history.build(result.payload.location as TLocationRouter));
```
danielweck: I would like to suggest using `takeTyped` as follows:
```
        const action = yield* takeTyped(onRouterLocationChanged);
        yield put(winActions.history.build(action.payload.location));
```

`takeTyped` and `onRouterLocationChanged` are imported as such:
```
import { takeTyped } from "readium-desktop/common/redux/typed-saga";
import { onRouterLocationChanged } from "readium-desktop/renderer/routing";
```
`onRouterLocationChanged` is defined as:

```
// LOCATION_CHANGE action of connected-react-router is type:
// export type TLocationChangeAction = LocationChangeAction<TLocationRouter>;
// (when using "yield* takeTyped(onRouterLocationChanged)")
// tslint:disable-next-line: max-line-length
type TActionPattern = <T extends IRouterLocationState>(location: Location<T>, action: RouterActionType, isFirstRendering?: boolean) => LocationChangeAction<T>;
export const onRouterLocationChanged = onLocationChanged as TActionPattern;
```

...with some additional imports:
```
import {
    LocationChangeAction, onLocationChanged, push, RouterActionType,
} from "connected-react-router";
import { Location, LocationDescriptorObject } from "history";
```

danielweck: Please can we also update the other use of untyped `LOCATION_CHANGE` here:
https://github.com/readium/readium-desktop/blob/a3db8a39fff0da2451fc2059f1d6ba44fdbb16ca/src/renderer/redux/sagas/opds.ts#L33-L34
Thank you :)

panaC: I'm not a super fan of this way of doing things.
Indeed, we move the typing issue on the other code base side. if the 'connected-router' module change his type, we don't infer it in your base code. it's force casted.

danielweck: Yes, that is indeed what the current code does:
`const result: LocationChangeAction = yield take(LOCATION_CHANGE);`
(`: LocationChangeAction`)
+
`yield put(winActions.history.build(result.payload.location as TLocationRouter)); `
(`as TLocationRouter`)

=> two force-cast / type coercion!

I am suggesting a cleaner method which is consistent with the rest of the Thorium codebase, and which will trigger a compile error if/when TypeScript type definitions are updated in the library dependencies.

danielweck: PS: ideally, `onLocationChanged as TActionPattern` would simply refine the existing generic type of the `onLocationChanged` function from `connected-react-router`, but I do not know how to do this. Do you know if it is possible?

danielweck: @panaC  thank you for bringing to my attention the missing `toString()`! Indeed this has to be added in `readium-desktop/renderer/routing`:
```
onLocationChanged.toString = () => LOCATION_CHANGE;
```

danielweck: All that being said, if I understand correctly you are against the suggested "improvement", right?
If so, I maintain my position based on the arguments expressed in this comment thread (sorry, you have not convinced me otherwise :), but I will be happy to merge this PR as-is, if you wish (we have more important problems to tackle, let us not block this PR because of relatively minor TypeScript typing disagreements).

panaC: don't worries Daniel, I fixed it with a style that you should love

danielweck: I like it indeed :)
I hope you do too ;)

One minor downside: there is now a chain of function calls which introduces runtime costs (`build()` -> `onLocationChanged()`) instead of just specifying types which only introduce compile-time costs. However, I think this is a very minor trade-off, easily outweighed by the improved code clarity and type safety.

danielweck:
Compared with the `develop` branch, this PR breaks the `displayType` state forwarding between routes (i.e. `grid` vs. `list` is not preserved between local bookshelf and remote OPDS catalogs).

PR #881, opened 2019-12-09
opds2 : groups, facets, metadatas

panaC:


panaC:
FIXME : add pointer hover on 'a' links in formatContributor.tsx

https://stackoverflow.com/questions/2409836/how-to-set-cursor-style-to-pointer-for-links-without-hrefs

src/common/views/opds.ts
```diff
@@ -78,10 +87,34 @@ export interface OpdsAuthView {
...
+    numberOfItems?: number | undefined;
+    priceValue?: number | undefined;
+    priceCurrency?: OPDSCurrencyEnum | undefined;
+    holdTotal?: number | undefined;
+    holdPosition?: number | undefined;
+    copyTotal?: number | undefined;
+    copyAvailable?: number | undefined;
+    availabilityState?: OPDSAvailabilityEnum | undefined;
+    availabilitySince?: string | undefined;
+    availabilityUntil?: string | undefined;
```
danielweck: Just a thought: why `?` optional property as well as `undefined` union type?

panaC: not the same thing , but in this case is useless

danielweck: Because TypeScript `strictNullCheck` is not currently used, we have these kinds of inconsistencies. So, for now, let's accept this redundant syntax to help with Intellisense in Visual Studio Code.

PR #902, opened 2019-12-26
[fix #882, #897] header translation

### LINKED ISSUE #882: When changing language, tab headers only refresh when switching tab (regression)
I think this is a regression from this PR:
https://github.com/readium/readium-desktop/pull/879/files#diff-a1228900df8ff573cefb675232bfceb2
(not 100% sure, but the `__` translate props looks suspicious)

### LINKED ISSUE #897: wrong navigation header language when the app is launched
Move the app in French. Close it. 
Reopen it. The navigation header + the second level (search / All books) are in English, the other parts are in French. 
Select Catalogs -> the navigation header is now in French. 

It seems to be a bad initialization of the navigation header.

panaC:
fixes #882 #897

### DIFF
--- a/src/renderer/redux/sagas/i18n.ts
+++ b/src/renderer/redux/sagas/i18n.ts
@@ -13,8 +13,11 @@ import { all, call } from "redux-saga/effects";
 function* localeWatcher() {
     while (true) {
         const action = yield* takeTyped(i18nActions.setLocale.build);
+
         const translator = diRendererGet("translator");
         translator.setLocale(action.payload.locale);
+
+        document.documentElement.setAttribute("lang", action.payload.locale);
     }
 }
 
--- a/src/renderer/redux/sagas/win.ts
+++ b/src/renderer/redux/sagas/win.ts
@@ -5,14 +5,19 @@
 // that can be found in the LICENSE file exposed on Github (readium) in the project repository.
 // ==LICENSE-END==
 
-import { SagaIterator } from "redux-saga";
-import { put, take } from "redux-saga/effects";
-
+import { i18nActions } from "readium-desktop/common/redux/actions/";
 import { winActions } from "readium-desktop/renderer/redux/actions";
+import { SagaIterator } from "redux-saga";
+import { all, put, take } from "redux-saga/effects";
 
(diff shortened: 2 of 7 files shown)


src/renderer/redux/sagas/win.ts
```diff
@@ -5,14 +5,19 @@
...
 
 export function* winInitWatcher(): SagaIterator {
     while (true) {
-        yield take(winActions.initRequest.ID);
+
+        yield all({
+            win: take(winActions.initRequest.ID),
+            i18n: take(i18nActions.setLocale.ID),
+        });
+
```
danielweck: When looking at this PR #902 in isolation, it is not clear why combining `winActions.initRequest` and `i18nActions.setLocale` is so important to trigger `winActions.initSuccess`.

So, I would like to suggest closing this PR, and instead promote PR #903 as a *complete* self-explanatory solution to issues #882 and #897. The critical code change is evident in this particular diff:
https://github.com/readium/readium-desktop/pull/903/files#diff-c5e985b08bbba519929349eeff28018fR19

@panaC what do you think?

src/renderer/redux/sagas/i18n.ts
```diff
@@ -13,8 +13,11 @@ import { all, call } from "redux-saga/effects";
 function* localeWatcher() {
     while (true) {
         const action = yield* takeTyped(i18nActions.setLocale.build);
+
         const translator = diRendererGet("translator");
         translator.setLocale(action.payload.locale);
+
+        document.documentElement.setAttribute("lang", action.payload.locale);
```
danielweck: I do not know if performing DOM mutations from a Redux Saga is "good development practice" (as opposed to doing this from a React component / `*.tsx` file) ... but as far as I am concerned this is an acceptable coding "shortcut", because DOM mutations are hacky in the first place anyway (much like relying extensively on React DOM refs). So, unless somebody has a strong opinion against migrating this minor DOM mutation into renderer-process Redux Saga, I am fine with the proposed change.

src/renderer/redux/sagas/win.ts
panaC: you're right

Issue #554, opened 2019-08-12
QueryString parsing (duplicate / redundant techniques)


danielweck:
`import { parseQueryString } from "readium-desktop/utils/url";`
https://github.com/readium/readium-desktop/blob/4e5ec305a4694ad75b691c98314de7d3295c44c9/src/utils/url.ts#L15-L29

_______
vs.
_______
`import * as qs from "query-string";`
https://github.com/readium/readium-desktop/blob/4e5ec305a4694ad75b691c98314de7d3295c44c9/package.json#L186

_______
vs.
_______
`import * as qs from "qs";`
https://github.com/readium/readium-desktop/blob/4e5ec305a4694ad75b691c98314de7d3295c44c9/package.json#L185
https://github.com/readium/readium-desktop/blob/4e5ec305a4694ad75b691c98314de7d3295c44c9/package.json#L227

danielweck:
Note that some `r2-xxx-js` packages also use an ad-hoc utility function (created a long time ago):

```
export interface IStringMap { [key: string]: string; }

export const getURLQueryParams = (search?: string): IStringMap => {
    const params: IStringMap = {};

    let query = search || window.location.search;
    if (query && query.length) {
        query = query.substring(1);
        const keyParams = query.split("&");
        keyParams.forEach((keyParam) => {
            const keyVal = keyParam.split("=");
            if (keyVal.length > 1) {
                params[keyVal[0]] = decodeURIComponent(keyVal[1]);
            }
        });
    }

    return params;
};
```

...and there is some redundancy with `querystring.[un]escape()` and these ad-hoc RFC-compliant utility functions:

```
export function encodeURIComponent_RFC3986(str: string): string {
    return encodeURIComponent(str)
        .replace(/[!'()*]/g, (c: string) => {
            return "%" + c.charCodeAt(0).toString(16);
        });
}

export function encodeURIComponent_RFC5987(str: string): string {
    return encodeURIComponent(str).
        replace(/['()]/g, querystring.escape). // i.e., %27 %28 %29
        replace(/\*/g, "%2A").
        // |`^
        replace(/%(?:7C|60|5E)/g, querystring.unescape);
}
```

!

danielweck:
Updated code references:

https://github.com/readium/readium-desktop/blob/aacfe6bbd33db5623a01cfb2939713b6015c8790/src/renderer/components/reader/ReaderMenu.tsx#L9

[...]

PR #903, opened 2019-12-26
[READY] centralization of the starting point for reader and main app

### LINKED ISSUE #882: When changing language, tab headers only refresh when switching tab (regression)

panaC:
fixes #882 #897


src/common/services/translator.ts
```diff
@@ -132,6 +132,10 @@ export class Translator {
     public translate: I18nTyped = this._translate;
     private locale: string = "en";
 
+    constructor(defaultLocale = "en") {
```
danielweck: Is this change necessary? The Dependency Injection Container will not automatically create a `Translator` instance with a constructor parameter, so this would need to be an explicit `new Translator("fr")` (for example) from `di.ts` ... but this is currently not the case:

https://github.com/readium/readium-desktop/blob/aacfe6bbd33db5623a01cfb2939713b6015c8790/src/main/di.ts#L167

src/renderer/components/App.tsx
```diff
@@ -18,23 +17,13 @@ import * as dialogActions from "readium-desktop/common/redux/actions/dialog";
...
 import DownloadsPanel from "./DownloadsPanel";
 import ToastManager from "./toast/ToastManager";
 
 export default class App extends React.Component<{}, undefined> {
 
-    @lazyInject(diRendererSymbolTable.store)
-    private store: Store<RootState>;
-
-    @lazyInject(diRendererSymbolTable.history)
-    private history: History<IRouterLocationState>;
```
danielweck: This change is interesting. Are you recommending removing *all* occurrences of `@lazyInject` in the renderer process code? (and instead replace with `diRendererGet("xx")` wherever required?) Do you think `@lazyInject` is unreliable, or is this change motivated by personal coding taste?

src/common/services/translator.ts
panaC: you're right c2ca422276e16d13109655e3cec416cacbad3969

src/renderer/redux/sagas/win.ts
```diff
@@ -5,14 +5,55 @@
...
+
         yield put(winActions.initSuccess.build());
     }
 }
+
+let notRendered = true;
+
+function* winStartWatcher(): SagaIterator {
+
+    while (true) {
```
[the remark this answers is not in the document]
panaC: you're right 1730aa3f473e58fb3eebc7d928d7d9b2d22e00e4

src/renderer/redux/sagas/i18n.ts
```diff
@@ -13,8 +13,11 @@ import { all, call } from "redux-saga/effects";
 function* localeWatcher() {
     while (true) {
         const action = yield* takeTyped(i18nActions.setLocale.build);
+
         const translator = diRendererGet("translator");
         translator.setLocale(action.payload.locale);
+
+        document.documentElement.setAttribute("lang", action.payload.locale);
```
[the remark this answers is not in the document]
panaC: ok

src/renderer/components/reader/Reader.tsx
```diff
@@ -76,10 +74,7 @@ import optionsValues, {
 //     supportFetchAPI: true,
 // });
 
-/**
- * WHY lot of const variable not in constructor ?
- */
-const queryParams = getURLQueryParams();
+const queryParams = () => getURLQueryParams();
```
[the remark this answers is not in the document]
panaC: yes , it's necessary when imported from redux-saga `win.ts`.
What is your proposal ?

src/renderer/components/App.tsx
panaC: `lazyInject` causes problems when importing from redux-saga

danielweck: Ok, so this PR removes all usages of `@lazyInject`.

src/renderer/components/reader/Reader.tsx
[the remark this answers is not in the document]
danielweck: Related issue: https://github.com/readium/readium-desktop/issues/554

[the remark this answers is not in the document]
danielweck: `window.location.search` should be "statically" / "immediately" available to the `BrowserWindow`'s URL, see:
https://github.com/readium/readium-desktop/blob/aacfe6bbd33db5623a01cfb2939713b6015c8790/src/main/redux/sagas/reader.ts#L130-L149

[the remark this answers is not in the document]
panaC: resolved by technical call

src/renderer/components/reader/Reader.tsx
```diff
@@ -145,19 +138,22 @@ const computeReadiumCssJsonMessage = (): IEventPayload_R2_EVENT_READIUMCSS => {
...
 };
-setReadiumCssJsonGetter(computeReadiumCssJsonMessage);
 
-const publicationJsonUrl = queryParams.pub.startsWith(READIUM2_ELECTRON_HTTP_PROTOCOL) ?
-    convertCustomSchemeToHttpUrl(queryParams.pub) : queryParams.pub;
+const queryParams = getURLQueryParams();
+const lcpHint = queryParams.lcpHint;
+// pub is undefined when loaded in dependency injection by library webview.
+// Dependency injection is shared between all the renderer view
+const publicationJsonUrl = queryParams.pub?.startsWith(READIUM2_ELECTRON_HTTP_PROTOCOL)
```
danielweck: Is the question mark `?` and the code comment necessary here? If I am not mistaken, the URL query parameter `pub` is available immediately and reliably in the `BrowserWindow` / `webContents` URL:

https://github.com/readium/readium-desktop/blob/aacfe6bbd33db5623a01cfb2939713b6015c8790/src/main/redux/sagas/reader.ts#L130-L149

panaC: no, it's always necessary here, the Reader code is imported in DI. And the DI is shared between reader and library. I am working to improve that in the future PR

src/renderer/di.ts
```diff
@@ -16,6 +16,9 @@ import { diRendererSymbolTable as diSymbolTable } from "readium-desktop/renderer
 import { RootState } from "readium-desktop/renderer/redux/states";
 import { initStore } from "readium-desktop/renderer/redux/store/memory";
 import { Store } from "redux";
+
+import * as mainApp from "./components/App";
+import * as readerApp from "./components/reader/App";
```
danielweck: I would like to suggest changing this to a cleaner syntax:
`import MainApp from "./components/App";`

...and simply use `MainApp` instead of `mainApp.default` in the code below.

src/renderer/components/reader/Reader.tsx
danielweck: Ah yes of course, WebPack's statically-created bundle (at build time) imports the entire code hierarchy  ... so we need a separate PR to split the reader / library scopes.

src/renderer/di.ts
panaC: 1e07d8e9

danielweck:
Note that this PR messes-up the reader/library CSS due to combining both scopes into the same WebPack bundle. This is because we are staging multiple consecutive (small) PRs in order to solve the bigger issues, incrementally. So the develop branch currently runs but has some rendering bugs.

PR #905, opened 2020-01-02
fixes #889

### LINKED ISSUE #889: JS error, uncaught exception (on Windows)
This happened on a Windows machine. The app was in background (maybe after being closed) and there were issues with the internet connection on this machine.  
Bug thorium

panaC:
fixes #889

### DIFF
--- a/src/common/utils/http.ts
+++ b/src/common/utils/http.ts
@@ -79,13 +79,6 @@ export async function httpGet<TBody extends JsonMap | string = string , TData = 
         await new Promise((resolve, reject) => {
             request(requestOptions, (err, response) => {
 
-                debug("HTTP RESPONSE:");
-                debug(err);
-                debug(response.statusCode);
-                debug(response.statusMessage);
-                debug(response.url);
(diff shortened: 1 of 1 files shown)


src/common/utils/http.ts
```diff
@@ -79,13 +79,6 @@ export async function httpGet<TBody extends JsonMap | string = string , TData =
         await new Promise((resolve, reject) => {
             request(requestOptions, (err, response) => {
 
-                debug("HTTP RESPONSE:");
-                debug(err);
-                debug(response.statusCode);
-                debug(response.statusMessage);
-                debug(response.url);
-                debug(response.body?.substr ? response.body.substr(0, 800) : response.body);
-
```
danielweck: I would prefer to leave all logging above `if (err)`, so that any HTTP `err` is logged (With the proposed change, logging of `err` is skipped if `err` is truthy). Consequently, I would suggest simply adding the missing `?` for each `debug(response.xxx);` (e.g. `debug(response?.xxx);`)

PR #907, opened 2020-01-03
Panac/fix/i18n set locale promise

panaC:


src/main/redux/sagas/i18n.ts
```diff
@@ -7,23 +7,32 @@
...
             identifier: LocaleConfigIdentifier,
-            value: { locale: action.payload.locale },
-        }));
-    }
+            value: {
+                locale: action.payload.locale,
+            },
+        });
+    const translatorSetLocale = () =>
+        translator.setLocale(action.payload.locale);
```
danielweck: Missing `await` when invoking a Promise-returning function (this would have been flagged-up if the TypeScript project was configured with stricter checks).
PS: `async` must be added in `    const translatorSetLocale = async () =>` too.
PS: Redux `call` will automatically handle the `Promise<void>`-returning `translatorSetLocale` function.

Same remark with `configRepository.save()` (above).

src/renderer/redux/sagas/i18n.ts
```diff
@@ -6,16 +6,20 @@
...
-        const action = yield* takeTyped(i18nActions.setLocale.build);
-        const translator = diRendererGet("translator");
+function* setLocale(action: i18nActions.setLocale.TAction) {
+    const translator = diRendererGet("translator");
+
+    const translatorSetLocale = () =>
         translator.setLocale(action.payload.locale);
-    }
+
+    yield call(translatorSetLocale);
```
danielweck: Same remark here about `await` on the `setLocale()` Promise-returning function.

danielweck:
Note that in addition to the PR review comments, the codebase has conflicts that need solving:
* `src/renderer/components/reader/Reader.tsx`
* `src/renderer/redux/sagas/i18n.ts`

src/renderer/redux/sagas/i18n.ts
panaC: https://github.com/readium/readium-desktop/pull/907/commits/73dbaf736242aa558ac43cc7b35d05960b1c9af6

src/main/redux/sagas/i18n.ts
panaC: https://github.com/readium/readium-desktop/pull/907/commits/73dbaf736242aa558ac43cc7b35d05960b1c9af6

PR #910, opened 2020-01-06
[Proposal] api decorator + redux connect decorator + translator decorator

panaC:
Ready to merge after some deep tests and the merge of #930.

I plan to used it in first with the refactor of the reader react components


panaC:
proposal which will remain a proposal

Issue #898, opened 2019-12-19
GitHub release check, rate limiting


danielweck:
I am seeing HTTP requests to GitHub at regular intervals:
```
[2]   readium-desktop:common:utils:http HTTP RESPONSE: +809ms
[2]   readium-desktop:common:utils:http null +0ms
[2]   readium-desktop:common:utils:http 403 +0ms
[2]   readium-desktop:common:utils:http Forbidden +0ms
[2]   readium-desktop:common:utils:http  +0ms
[2]   readium-desktop:common:utils:http {
[2]   readium-desktop:common:utils:http   message: "API rate limit exceeded for 217.44.31.75. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
[2]   readium-desktop:common:utils:http   documentation_url: 'https://developer.github.com/v3/#rate-limiting'
```
```
[2]   readium-desktop:common:utils:http HTTP REQUEST: +2m
[2]   readium-desktop:common:utils:http {
[2]   readium-desktop:common:utils:http   timeout: 5000,
[2]   readium-desktop:common:utils:http   json: true,
[2]   readium-desktop:common:utils:http   headers: {
[2]   readium-desktop:common:utils:http     'user-agent': 'readium-desktop',
[2]   readium-desktop:common:utils:http     'accept-language': 'en,en-US;q=0.7,en;q=0.5'
[2]   readium-desktop:common:utils:http   },
[2]   readium-desktop:common:utils:http   url: 'https://api.github.com/repos/edrlab/readium-desktop/releases/latest',
[2]   readium-desktop:common:utils:http   method: 'GET',
[2]   readium-desktop:common:utils:http   encoding: undefined
[2]   readium-desktop:common:utils:http } +0ms
```

danielweck:
Code culprit (check every `120000` milliseconds?):

https://github.com/readium/readium-desktop/blob/aacfe6bbd33db5623a01cfb2939713b6015c8790/src/main/redux/sagas/update.ts#L17-L60

danielweck:
Note that the Redux "update" action is not used anywhere, so I would suggest disabling this "ping" code until application auto-updates are enabled somewhere in the codebase (we could use Electron Builder's own capabilities):
https://github.com/readium/readium-desktop/blob/aacfe6bbd33db5623a01cfb2939713b6015c8790/src/common/redux/actions/update/latestVersion.ts#L11-L33

danielweck:
[...]

PR #912, opened 2020-01-08
disable net and update redux state

panaC:


danielweck:
Fixes https://github.com/readium/readium-desktop/issues/898

PR #913, opened 2020-01-08
...

### LINKED ISSUE #911: new copy of state in each redux reducer loop
src/renderer/redux/reducers/reader.ts - ln 46
src/main/redux/reducers/reader.ts - ln 46

panaC:
fixes #911


panaC:
@danielweck What do you think to push computeReadiumCssJsonMessage in dedicated common file. ?
But doesn't work in reader window , error in pagination

panaC:
i have a bug

Scrollable and Paginated is reversed

image

image

@danielweck any idea ?

PR #926, opened 2020-01-21
[WIP]{REFACTORING} Reader

panaC:


src/common/models/reader.ts
```diff
@@ -21,7 +21,11 @@ export interface Reader extends Identifiable {
     filesystemPath: string;
     manifestUrl: string;
     publicationIdentifier: string;
+
+    // should be removed, not used and useless
     browserWindow: BrowserWindow;
+
+    // should be removed, set in win root state
```
danielweck: This is used, and it is a critical piece of information for managing `BrowserWindow` instances.

src/common/models/reader.ts
```diff
@@ -21,7 +21,11 @@ export interface Reader extends Identifiable {
     filesystemPath: string;
     manifestUrl: string;
     publicationIdentifier: string;
+
+    // should be removed, not used and useless
```
danielweck: Correct :)

src/renderer/reader/components/Reader.tsx
```diff
@@ -605,7 +605,11 @@ class Reader extends React.Component<IProps, IState> {
     }
 
     private async handleReadingLocationChange(loc: LocatorExtended) {
+
+        // Why findBookmark here ? can be removed
```
danielweck: I agree, but we have to make sure that `checkBookmarks()` is called (automatically from `componentDidUpdate()` if `oldState.bookmarks !== this.state.bookmarks`).

src/renderer/reader/components/Reader.tsx
```diff
@@ -605,7 +605,11 @@ class Reader extends React.Component<IProps, IState> {
...
         this.findBookmarks();
+
+        // every 250ms on user scrolling. This will be saved on db through redux
```
danielweck: The navigator component *debounces*  on the trailing edge of a fast sequence of interactions / events.

PR #930, opened 2020-01-22
[WIP][NEW FEATURES WAITING DEEP TESTS BEFORE MERGING]added session management and a publication register | improved asynchronism logic in the main process with code transfer to redux saga only | first attempt to use redux as a state store with persistence

### LINKED ISSUE #911: new copy of state in each redux reducer loop

panaC:
fixes #911 #947 #1030 #1035 #1037 #1036 

315b1e49116daff0eb4e242207075378ef1b79aa


panaC:
note to myself : handle "reader/setLastReadingLocation"

src/common/error.ts
```diff
@@ -0,0 +1,13 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END==
+
+import { dialog } from "electron";
+
+export function error(title: string, err: string) {
+
+    dialog.showErrorBox("Internal Thorium ERROR", `${title}\nERROR: ${err}\n\nYou should probably restart the application`);
+}
```
danielweck: Great :)
Just for discussion:
It would be interesting to run a search in the codebase for all `await` function calls, and to identify how many of them do not have a `try+catch` handler for Promise rejection. I think there are many instances of unhandled rejections.

The problem is that developers assume that missing `try+catch` is fine because the exception simply gets cascaded up the call chain, and that at some point it will be handled. This is only partially true.

For better error handling practice, it is better to think of `try+catch` as the equivalent of `promise.then().catch()` ... which makes it much more obvious that promise rejections should be handled "locally" (using the contextual logic of the promise's intended purpose), rather than assuming that a higher-level / less-local logical layer of the code will somehow automagically handle the rejection appropriately.

What do you think?

panaC:
https://github.com/readium/readium-desktop/pull/513/files#r383369057

danielweck:
> https://github.com/readium/readium-desktop/pull/513/files#r383369057

Yes, the fix is quite simple: https://github.com/readium/readium-desktop/commit/36fbca41c4307b4a0703164cd9873bc7364d47ea

(feel free to remove commented-out code in your cleanup)

panaC:
### 03/05 pull request status

I've run some deep-test
- with session enabled and not enabled
- with many detached publication reader windows in // library windows
- the behaviour between enabled session and not enabled session on closing library window
    - session enabled and user choice to save session : save all reader to restart them during next starting process.
    - session not enabled : close all reader windows and don't restart them during next starting process
- lcp publication : renew, revoke, borrow, remove, lpcl import.
- publication-info in lib and reader
- settings
- opds catalogs: browse, download, ...
- search

#### in pending in this PR :

- @danielweck can you help me about css bug in reader window.
   - I don't understand why the menu size is reduced and narrowed
- complete the i18n translation.

otherwise, I think it's on track to merge it in the future.

danielweck:
Great, thank you for testing methodically. I will take a look at the CSS regression.

src/renderer/reader/components/Reader.tsx
```diff
@@ -266,19 +205,35 @@ class Reader extends React.Component<IProps, IState> {
         ensureKeyboardListenerIsInstalled();
         this.registerAllKeyboardListeners();
 
+        const store = diReaderGet("store");
+
+        const pubId = store.getState().reader.info.publicationIdentifier;
+        const locator = store.getState().reader.locator;
+        const manifestUrl = store.getState().reader.info.manifestUrl;
```
danielweck, line 212: This causes a regression: localhost URL instead of special Readium HTTPS custom protocol. I am looking into it.

danielweck: Fixed via https://github.com/readium/readium-desktop/commit/4e2fe146699bc88e39619d8cc35108f778b54dcd

PR #957, opened 2020-02-13
Add Spanish translation

nvdaes:


nvdaes:
@danielweck  : I have added the spanish translation as commented on email list.
Hope this helps. Thanks for your attention and work.

panaC:
thank you very much for your PR.

Do you know how to connect the new language in the code ?
So that your PR can be integrated into thorium.

check this file : https://github.com/readium/readium-desktop/blob/develop/src/common/services/translator.ts

Otherwise we will integrate it

nvdaes:
I modified the mentioned file.  Please let me know if something needs to be added or changed. I suppose that locale-names files are automatically generated. Hope a procedure for contributing can be documented, maybe helpful :)

nvdaes:
With my last commit, checks failed. Seems that src\resources\locale-names/esLang.json needs to be created to translate language names. In this case, I will push another commit for this, based on
http://utils.mucattu.com/iso_639-1.html
Please, let me know if this is not needed.

panaC:
yes, you are right, @Scarsniik used this [script](https://github.com/readium/readium-desktop/blob/develop/scripts/csvToJson.py) to convert `csv` from wikipedia to `json`.
If that can help you. I guess these ISO codes are not easy to get.

nvdaes:
I have followed the following criteria: 
- Capitalize initial character of language names, as done in other languages, though in Spanish is only required in certain situations.
- Use the first or most wnown or easy name when various names can be used for the same language. 
Please, review and merge when appropiate. Thanks

panaC:
perfect, thank-you. I tested it works well.

@danielweck will be able to merge if he agrees.

nvdaes:
Thanks, if/when @danielweck  merges this, I will make a post to share this with spanish users.

danielweck:
Thank you! Merging this PR into a staging branch in order to check against the latest updates (keyboard shortcut editor new labels, etc.)

danielweck:
Follow-up PR: https://github.com/readium/readium-desktop/pull/971

Issue #155, opened 2018-05-31
Keyboard shortcuts


gregoriopellegrino:
OS: Windows 10
Assistive technology: NVDA 2018.1.1

It is quite difficult to move in the UI between panels (TOC, pages, content, header): it would be very useful to implement the keyboard shortcut F6, that is "native" in Windows.

danielweck:
Note: in Readium "1" (cloud / web reader and Chrome app) a number of keyboard interactions / shortcuts are implemented in order to facilitate key-driven UI navigation.
Documentation: https://github.com/readium/readium-js-viewer/wiki/Keyboard

llemeurfr:
Reading the R1 document, I see mention of Windows function keys only where stated that F8 is not supported. So, first, should we support Windows function keys here ? then what is the list of function keys we should support, for which action (for which assistive technology)?

llemeurfr:
From what I find on the Web, we could map some Windows function keys to useful actions: 

F1 | Brings up a Help window (but we don't have a Help so far)
F3 | Opens search box in browsers (could move to the search field in the bookshelf view, open a search box in the reading view)
F6 | Cycles through the screen elements in a window (see Gregorio's request)
F10 | Activates menu bar options (use to be discussed)
F11 | Toggles between full screen and normal display (could be a trigger to the Zen mode in the reading view)

llemeurfr:
From George's comments on the use of Thorium vs Daisy tests, it seems we are lacking a keyb shortcut to move between chapters without going through the ToC.

llemeurfr:
Only F6 seems really interesting in practice. But not easy to implement. Slack has this feature. 

F3: people use **Ctrl-F** in practice, and it does not seem really useful in the bookshelf view as the search field is always displayed. It will be useful for implementing search inside a book, later.

panaC:
How to do implement the F6 feature ?

The F6 allow to focus on each main screen block at each hit

danielweck:
Note that `F3` / `CTRL-F` (and the "reverse" variant using the `SHIFT` key modifier) can be used to cycle through search results too, depending on the context. So this is not just about quick-access to a text input field / search box.

danielweck:
`F6` (and `SHIFT` + `F6`) is about cycling through GUI landmarks. Note that screen readers typically layer additional logical navigation, based on structural-semantics in the HTML markup (e.g. heading levels, ARIA roles, etc.).

danielweck:
[...]

PR #958, opened 2020-02-13
configurable keyboard shortcuts

danielweck:
Fixes issue https://github.com/readium/readium-desktop/issues/155


src/main/redux/sagas/keyboard.ts
```diff
@@ -35,32 +41,65 @@ function* setShortcuts(action: keyboardActions.setShortcuts.TAction) {
...
 function* keyboardShowWatcher() {
     yield takeEvery(keyboardActions.showShortcuts.build, showShortcuts);
 }
+// function* keyboardReloadWatcher() {
+//     yield takeEvery(keyboardActions.reloadShortcuts.build, reloadShortcuts);
+// }
 function* keyboardReloadWatcher() {
-    yield takeEvery(keyboardActions.reloadShortcuts.build, reloadShortcuts);
+    while (true) {
+        const action = yield* takeTyped(keyboardActions.reloadShortcuts.build);
```
panaC: but why ??

you should may be surround `yield takeEvery` with `yield all([takeEvery])` ,
perhaps it's the raison.

src/common/redux/reducers/keyboard.ts
```diff
@@ -22,13 +22,26 @@ export function keyboardReducer(
...
+                    ...{
+                        shortcuts: action.payload.shortcuts,
+                    },
+                }
+            );
         case keyboardActions.showShortcuts.ID:
-            return Object.assign({}, state);
+            return (
+                {
+                    ...state,
```
panaC: it may be useless, right ?

src/main/redux/sagas/keyboard.ts
danielweck: There are several issues of this type in this PR. I propose we fix them in a second pass (subsequent PR).

src/common/redux/reducers/keyboard.ts
danielweck: There are several issues of this type in this PR. I propose we fix them in a second pass (subsequent PR).

danielweck:
The current set of keyboard shortcuts is described in this Wiki page:
https://github.com/readium/readium-desktop/wiki/Keyboard-shortcuts

PR #961, opened 2020-02-25
re-instated publication identifier in metadata info dialog

danielweck:
The publication identifier metadata was removed in this Pull Request:
#881

@llemeurfr I seem to remember that you advocated for the removal of the publication unique identifier (ISBN or otherwise) in the information / metadata popup dialog, but I cannot find a record of related conversations / decisions. Sorry if I am being wrong :)

Personally, I feel that even though the unique publication identifier is likely irrelevant / useless to most (human) readers, it may be important to some people, for some usage scenarios. So I would be in favour of re-instating it.

Thorium_publication-metadata-id_1

Thorium_publication-metadata-id_2


llemeurfr:
Indeed I proposed its removal, because useless for most users. Especially if it's a uuid. Only an ISBN may be useful for "standard" users, for whom Thorium is tailored. 
But let's see what others think...

danielweck:
There has not been a strong demand for this feature. Closing.

PR #971, opened 2020-03-16
Spanish translation

danielweck:
Follow-up to https://github.com/readium/readium-desktop/pull/957
(full credits to @nvdaes, I just added missing labels related for the new keyboard shortcuts editor)


nvdaes:
Thanks @danielweck . I'm very grateful for the spanish translation and I am sharing this with spanish mailing lists.
Could we update the translation about keyboard? I have seen some mistakes, like the word translated as "Espectáculo", which is one of the possible translations for "show", but this doesn't make sense in the current context. Also the word translated as "Salvar", which should be translated as "Guardar".
Please let me know how we should proceed with this. I will take care ASAP (may take some time).

danielweck:
Thank you for suggesting corrections! For simple edits of existing localization keys, if you create a Pull Request it will be very easy and rapid to merge.

nvdaes:
Great! I will do it ASAP, probably in the weekend, maybe earlier. Also, I would like to place json entries in different lines. This would be easier to see diffs and manage files.

danielweck:
The JSON is automatically ordered (canonical alphabetical sorting of hierarchical translation keys)

PR #973, opened 2020-03-17
[feature][fix #940] improve audiobook info in pubInfo dialog

### LINKED ISSUE #940: audiobooks: duration metadata display in publication info dialog
Follow-up to PR https://github.com/readium/readium-desktop/pull/938

panaC:
fixes #940

### DIFF
--- a/src/common/views/publication.ts
+++ b/src/common/views/publication.ts
@@ -19,6 +19,13 @@ export interface CustomCoverView {
     bottomColor: string;
 }
 
+export interface ITimeDuration {
+    days: number;
+    hours: number;
+    seconds: number;
+    minutes: number;
+}
+
 export interface PublicationView extends Identifiable {
     title: string;
     authors: string[];
@@ -30,6 +37,9 @@ export interface PublicationView extends Identifiable {
     publishedAt?: string; // ISO8601
     cover?: CoverView;
     customCover?: CustomCoverView;
+
+    duration: ITimeDuration;
+    nbOfTracks: number;
 
     lcp?: LcpInfo;
     lcpRightsCopies?: number;
--- a/src/common/views/publication.ts
+++ b/src/common/views/publication.ts
@@ -20,7 +20,6 @@ export interface CustomCoverView {
 }
 
 export interface ITimeDuration {
-    days: number;
     hours: number;
     seconds: number;
     minutes: number;
--- a/src/common/views/publication.ts
+++ b/src/common/views/publication.ts
@@ -38,8 +38,8 @@ export interface PublicationView extends Identifiable {
     cover?: CoverView;
     customCover?: CustomCoverView;
 
-    duration: ITimeDuration;
-    nbOfTracks: number;
+    duration?: ITimeDuration;
+    nbOfTracks?: number;
 
     lcp?: LcpInfo;
     lcpRightsCopies?: number;
--- a/src/main/converter/publication.ts
+++ b/src/main/converter/publication.ts
@@ -7,7 +7,9 @@
 
 import { injectable } from "inversify";
 import * as moment from "moment";
-import { CoverView, PublicationView } from "readium-desktop/common/views/publication";
+import {
+    CoverView, ITimeDuration, PublicationView,
+} from "readium-desktop/common/views/publication";
 import {
     convertContributorArrayToStringArray,
 } from "readium-desktop/main/converter/tools/localisation";
@@ -61,9 +63,29 @@ export class PublicationViewConverter {
             lcp: document.lcp,
             lcpRightsCopies: document.lcpRightsCopies,
 
-            // doc: r2Publication.Metadata,
+            duration: r2Publication.Metadata.Duration &&
+                this.convertSecondToDaysHoursMinutes(r2Publication.Metadata.Duration),
+            nbOfTracks: r2Publication.Metadata.AdditionalJSON?.tracks as number,
+
+            // doc: r2Publiction.Metadata,
 
             r2PublicationBase64,
+        };
+    }
+
+    private convertSecondToDaysHoursMinutes(seconds: number): ITimeDuration {
+        const days = Math.floor(seconds / (60 * 60 * 24));
+        seconds %= 60 * 60 * 24;
+        const hours = Math.floor(seconds / (60 * 60));
+        seconds %= 60 * 60;
+        const minutes = Math.floor(seconds / 60);
+        seconds %= 60;
+
+        return {
+            days,
+            hours,
+            minutes,
+            seconds,
         };
     }
 }
--- a/src/main/converter/publication.ts
+++ b/src/main/converter/publication.ts
@@ -74,15 +74,18 @@ export class PublicationViewConverter {
     }
 
     private convertSecondToDaysHoursMinutes(seconds: number): ITimeDuration {
-        const days = Math.floor(seconds / (60 * 60 * 24));
-        seconds %= 60 * 60 * 24;
-        const hours = Math.floor(seconds / (60 * 60));
-        seconds %= 60 * 60;
-        const minutes = Math.floor(seconds / 60);
-        seconds %= 60;
+
+        const secondsPerMinute = 60;
+        const minutesPerHours = 60;
+        const secondsPerHour = minutesPerHours * secondsPerMinute;
+
+        const hours = Math.floor(seconds / secondsPerHour);
+        seconds %= secondsPerHour;
+
+        const minutes = Math.floor(seconds / secondsPerMinute);
+        seconds %= secondsPerMinute;
 
         return {
-            days,
             hours,
             minutes,
             seconds,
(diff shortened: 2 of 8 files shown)


src/main/converter/publication.ts
```diff
@@ -61,9 +63,29 @@ export class PublicationViewConverter {
...
         };
     }
+
+    private convertSecondToDaysHoursMinutes(seconds: number): ITimeDuration {
+        const days = Math.floor(seconds / (60 * 60 * 24));
+        seconds %= 60 * 60 * 24;
+        const hours = Math.floor(seconds / (60 * 60));
+        seconds %= 60 * 60;
+        const minutes = Math.floor(seconds / 60);
+        seconds %= 60;
```
danielweck: Just nit-picking about code style :)
Would you mind introducing variables? (to replace duplicated calculations and make the code more readable)
For example:

```
const secondsPerMinute = 60;
const minutesPerHours = 60;
const hoursPerDay = 24;
const secondsPerHour = minutesPerHours * secondsPerMinute;
const secondsPerDay = secondsPerHour * hoursPerDay;

const days = Math.floor(seconds / secondsPerDay);
seconds %= secondsPerDay;

const hours = Math.floor(seconds / secondsPerHour);
seconds %= secondsPerHour;

const minutes = Math.floor(seconds / secondsPerMinute);
seconds %= secondsPerMinute;
```

src/common/views/publication.ts
```diff
@@ -31,6 +38,9 @@ export interface PublicationView extends Identifiable {
     cover?: CoverView;
     customCover?: CustomCoverView;
 
+    duration: ITimeDuration;
+    nbOfTracks: number;
```
danielweck: These fields should be optional, I think.
(I mean: question mark `?` before the colon character `:`)

src/main/converter/publication.ts
```diff
@@ -61,9 +63,29 @@ export class PublicationViewConverter {
             lcp: document.lcp,
             lcpRightsCopies: document.lcpRightsCopies,
 
-            // doc: r2Publication.Metadata,
+            duration: r2Publication.Metadata.Duration &&
+                this.convertSecondToDaysHoursMinutes(r2Publication.Metadata.Duration),
+            nbOfTracks: r2Publication.Metadata.AdditionalJSON?.tracks as number,
```
danielweck: As mentioned in a previous code comment, I think `nbOfTracks` and `duration` should be optional properties, so if I am not mistaken `as number` can be removed and the TypeScript compiler will accept `AdditionalJSON?.tracks` as being potentially `undefined` (note: not `null` ... but this doesn't matter right now because Thorium does not have `strictNullChecks` in its `tsconfig.js` ;)

src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
```diff
@@ -29,6 +30,41 @@ export interface IProps {
...
+    const { duration, __ } = props;
+
+    let sentence = "";
+    sentence += duration?.days ? `${duration.days} ${__("publication.duration.days")} ` : "";
+    sentence += duration?.hours ? `${duration.hours} ${__("publication.duration.hours")} ` : "";
+    sentence += duration?.minutes ? `${duration.minutes} ${__("publication.duration.minutes")} ` : "";
+    sentence += sentence ? `${__("publication.duration.and")} ` : "";
+    sentence += duration?.seconds ? `${duration.seconds} ${__("publication.duration.seconds")}` : "";
+
+    sentence = sentence.trim();
```
danielweck: I do not have an objection against this time format, but I think we should get some feedback to check that this is an acceptable representation depending on language locale.

Another common notation for time values that do not reach days is `HH:MM:SS` (here, ignoring milliseconds, for brevity), with the advantage of being relatively locale-agnostic.
```
const str = `${hours > 0 ? (hours.toString().padStart(2, "0") + ":") : ``}${minutes > 0 ? (minutes.toString().padStart(2, "0") + ":") : `00:`}${seconds > 0 ? (seconds.toString().padStart(2, "0")) : `00`}`;
```

src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
```diff
@@ -29,6 +30,41 @@ export interface IProps {
...
+
+    const isDefined = sentence.length > 0;
+
+    return (isDefined ?
```
danielweck: Code consistency: at line `44` there is a ternary expression which checks for "falsy" `sentence` (i.e. empty string) `sentence += sentence ? `${__("publication.duration.and")} ` : "";` ... so I would suggest removing `isDefined` and simply replace with `return (sentence ? ...`.

src/resources/locales/en.json
```diff
@@ -158,12 +158,23 @@
...
+            "tracks": "Tracks"
+        },
         "cancelledLcp": "This book can't be read because the LCP license has been cancelled.",
         "certificateRevoked": "Certificate is revoked",
         "certificateSignatureInvalid": "Certificate signature is invalid",
         "cover": {
             "img": "cover image"
         },
+        "duration": {
+            "and": "et",
```
danielweck: This should be `and` (currently `et`)

src/resources/locales/en.json
```diff
@@ -158,12 +158,23 @@
...
+            "days": "day(s)",
+            "hours": "hour(s)",
+            "minutes": "minute(s)",
+            "seconds": "second(s)",
+            "title": "Durations"
```
danielweck: I think this should be singular, not plural (in all language locales)

src/resources/locales/en.json
panaC: https://github.com/readium/readium-desktop/pull/973/commits/e6f8ded19d62fddf5d75efe4b943ae65f613bc7f

src/main/converter/publication.ts
panaC: https://github.com/readium/readium-desktop/pull/973/commits/b38967b14f3898ee4788b5a41a2392815fe9294e

src/common/views/publication.ts
panaC: https://github.com/readium/readium-desktop/pull/973/commits/b38967b14f3898ee4788b5a41a2392815fe9294e

src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
panaC: Please put your algo in mind directly in the issue tracker. I'm not in your mind :)

src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
panaC: https://github.com/readium/readium-desktop/pull/973/commits/4f30f388a60e72c529070853f1cd62b2dbd68f4d

src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
danielweck: It's a "human-friendly" version (i.e. minutes always shown as `00` even if zero, padded string to ensure consistent double-digit `**` even if `<10`) of NPT Normal Play Time from RFC 2326:
https://www.w3.org/TR/media-frags/#naming-time
https://tools.ietf.org/html/rfc2326#page-17

src/main/converter/publication.ts
danielweck: Oh, my bad, I forgot that `AdditionalJSON` is `AnyJson`, so naturally `as number` is needed! Thanks.

src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
panaC: I'm totally agree with you. 

thanks you Daniel for your proposition

src/typings/en.translation.d.ts
```diff
@@ -381,7 +381,16 @@ declare namespace typed_i18n {
...
   readonly "seeMore": string,
-  readonly "userKeyCheckInvalid": string
+  readonly "userKeyCheckInvalid": string,
+  readonly "duration": {
+    readonly "title": string,
+    readonly "days": string,
+    readonly "minutes": string,
+    readonly "hours": string,
+    readonly "seconds": string,
+    readonly "and": string
```
danielweck: TODO: `npm run i18next-typed`

src/resources/locales/en.json
```diff
@@ -158,12 +158,18 @@
...
+            "tracks": "Tracks"
+        },
         "cancelledLcp": "This book can't be read because the LCP license has been cancelled.",
         "certificateRevoked": "Certificate is revoked",
         "certificateSignatureInvalid": "Certificate signature is invalid",
         "cover": {
             "img": "cover image"
         },
+        "duration": {
+            "title": "Durations"
```
danielweck: I think this should be singular, not plural.

src/resources/locales/fr.json
```diff
@@ -158,12 +158,18 @@
         "shelf": "Étagère"
     },
     "publication": {
+        "audio": {
```
danielweck: TODO: Spanish. Can we just use Google Translate (or equivalent free service) to obtain the labels for audio "Track" and "Duration"?

PR #975, opened 2020-03-18
[fix #698 #554] Reinstate `no-floating-promises` in `tslint.json` and remove qs and query-string packages

### LINKED ISSUE #698: Reinstate `no-floating-promises` in `tslint.json`
(general principle: stricter static TypeScript compiler checks)

See originating discussion:
https://github.com/readium/readium-desktop/pull/659#discussion_r323306491

panaC:
fixes #698 #554


danielweck:
I suspect that this PR has introduced a regression bug (I am checking now). My fault for merging it too quick, I approved the removal of the `qs` packages without double-checking that the await/async changes could have a negative impact on runtime (I wrongly assumed linting resolution only). Next time, let's separate such distinct changes into different PRs, so we can review + test more easily.

src/renderer/common/apiAction.ts
```diff
@@ -15,58 +15,59 @@ import { v4 as uuidv4 } from "uuid";
...
-            }).catch((error) => {
-                reject(error);
-            }).finally(() => {
-                if (storeUnsubscribe) {
-                    storeUnsubscribe();
-                }
-                if (timeoutId) {
-                    clearTimeout(timeoutId);
-                }
+                await promise.then((result) => {
```
danielweck: Ah, I think this is the culprit for the regression bug. Fixing this now...

danielweck: Well, this needs fixing (which I did, and I will push the change), but the problem is elsewhere...

danielweck: Ok, problem solved. Pushing fix now...

danielweck:
Hotfix: https://github.com/readium/readium-desktop/commit/a09bd2c8c80a81d0286827c474da5c4fef00dea0

PR #981, opened 2020-03-24
fix function name

panaC:


src/main/converter/publication.ts
```diff
@@ -65,7 +65,7 @@ export class PublicationViewConverter {
 
             RDFType: r2Publication.Metadata.RDFType,
             duration: r2Publication.Metadata.Duration &&
```
danielweck: should a ternary operator be used here?
if r2Publication.Metadata.Duration is zero (number), null, or undefined ... the duration field will take these values instead of an instance of the HMS data structure ... this would be revealed by the TypeScript compiler if we had strictNullChecks :)
So, as a rule of thumb, I prefer to always ensure the resulting type of the boolean comparison, and a ternary operator allows us to normalise falsy (zero, null, undefined) to whatever we need (for example, undefined).

panaC: you're right

PR #985, opened 2020-03-25
[feature #979] add W3C audiobook support

### LINKED ISSUE #979: audiobooks: support for W3C Audiobooks
When the app imports a W3C Audiobook, it contains a W3c WebPub manifest named `publication.json` and may contain an html ToC in `index.html`. A import time, both files are processed, a Readium WebPub manifest (`manifest.json`) is generated and included inside the zip package. The zip file is renamed `.audiobook`. The file is now compliant with the internal format of the Readium project, without any loss of W3C information.  

The specification of W3C Audiobooks is [here](https://www.w3.org/TR/audiobooks/). The spec of the companion LPF format is [here](https://www.w3.org/TR/lpf/).

panaC:
fix #979 

[armand_durand_sample.zip](https://github.com/readium/readium-desktop/files/4396964/armand_durand_sample.zip)

rename it to `*.lpf`


src/common/extension.ts
```diff
@@ -0,0 +1,16 @@
...
+// Licensed to the Readium Foundation under one or more contributor license agreements.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END=
+
+export const acceptedExtension = (ext: string) =>
+    /\.epub[3]?$/.test(ext) ||
+    /\.audiobook$/.test(ext) ||
+    /\.lpf$/.test(ext) ||
+    ext === ".lcpl";
```
danielweck: Can this be changed to `/\.lcpl$/.test(ext)`?

src/common/extension.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const acceptedExtensionArray = [
+    ".lcpl", ".epub", ".epub3", ".audiobook", ".lpf",
```
danielweck: Could this array be iterated in the `acceptedExtension()` function, with this check inside the loop (untested code, just for illustration):
```
if (new RegExp(`\\.${acceptedExtensionArray[i]}$`).test(ext)) { return true; }
```

src/resources/locales/de.json
```diff
@@ -84,7 +84,7 @@
         "deleteFeed": "Willst du diesen Katalog wirklich löschen?",
         "deletePublication": "Willst du diese Publikation wirklich löschen?",
         "import": "Willst du diese EPUB-Dateien wirklich importieren?",
-        "importError": "Ein Fehler ist aufgetreten, überprüfen Sie die Dateiendung (.epub / .lcpl / .audiobook).",
+        "importError": "Ein Fehler ist aufgetreten, überprüfen Sie die Dateiendung (.epub / .lcpl / .audiobook / .lpf).",
```
danielweck: With the `i18next` "translator" we can easily handle interpolation, so could we remove the duplication of file extensions with `acceptedExtensionArray.join(" / ")`?
https://www.i18next.com/translation-function/interpolation

src/common/extension.ts
```diff
@@ -0,0 +1,28 @@
...
+// Licensed to the Readium Foundation under one or more contributor license agreements.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END=
+
+import { ObjectKeys, ObjectValues } from "readium-desktop/utils/object-keys-values";
+
+export const acceptedExtensionObject = {
+    lcpLicence: ".lcpl",
+    epub2: ".epub",
```
danielweck, line 12: The `epub2` key should be `epub`, as in fact this covers EPUB v2, v3, etc.

Explanation: the `.epub3` extension is a non-standard oddity implemented by https://standardebooks.org :)

src/main/lpfConverter.ts
```diff
@@ -0,0 +1,304 @@
...
+import { Publication as R2Publication } from "@r2-shared-js/models/publication";
+import { zipLoadPromise } from "@r2-utils-js/_utils/zip/zipFactory";
+import { iso8601DurationsToSeconds } from "readium-desktop/utils/iso8601";
+
+// Logger
+const debug = debug_("readium-desktop:main#lpfConverter");
+
+async function copyAndRenameLpfFile(lpfPath: string): Promise<string> {
+
+    const userPath = app.getPath("userData");
```
danielweck, line 33: I would suggest using a proper temporary directory instead.
`import { tmpNameSync } from "tmp";`
```
const tmpFolderPath = tmpNameSync({
            dir: "LPF",
            prefix: "readium-desktop-",
            postfix: ".extension"});
```

src/main/lpfConverter.ts
```diff
@@ -0,0 +1,304 @@
...
+    }
+    {
+        const iso8601 = `${w3cManifest.duration || ""}`;
+        const second = iso8601DurationsToSeconds(iso8601);
+        if (second > 0) {
+            publication.Metadata.Duration = second;
+        }
+    }
+    {
+        const ressources = w3cManifest.resources as JsonMap;
```
danielweck, line 190: Single `s` in `ressources` :)

src/main/lpfConverter.ts
```diff
@@ -0,0 +1,304 @@
...
+                            case ".mp3":
+                                link.TypeLink = "audio/mpeg";
+                                break;
+                            case ".wav":
+                                link.TypeLink = "audio/wav";
+                                break;
+                            case ".opus":
+                                link.TypeLink = "audio/ogg";
+                                break;
+                            default:
```
danielweck, line 243: I would add `.m4a` => `audio/m4a` as well.

danielweck: Also, just to be on the safe side (I don't expect to encounter audiobooks using these codecs, but...): `.ogg`, `.oga` => `audio/ogg`

src/main/lpfConverter.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    const stream = await openAndExtractPublicationFromLpf(lpfRenameInAudiobookPath);
+
+    const buffer = await streamToBufferPromise(stream);
+    const rawData = buffer.toString("utf8");
+    const w3cManifest = JSON.parse(rawData) as JsonMap;
+
+    const readiumManifest = w3cPublicationManifestToReadiumPublicationManifest(w3cManifest);
+
+    const manifestBuffer = Buffer.from(JSON.stringify(readiumManifest));
```
danielweck, line 300: I would suggest formatting the JSON output:
`JSON.stringify(readiumManifest, null, 4)`

panaC: done :)

src/main/lpfConverter.ts
panaC: done :)

src/main/lpfConverter.ts
panaC: thanks

src/common/extension.ts
panaC: done :)

src/main/lpfConverter.ts
panaC: `tmpNameSync` is the synchronous version of tmpName and should be `tmpName` instead

https://github.com/benjamingr/tmp-promise#synchronous-filename-generation
https://github.com/benjamingr/tmp-promise#asynchronous-filename-generation

this library store the file in the OS tmp directory: It's used the `os.tmpdir()` method to this.
https://github.com/raszi/node-tmp/blob/61f9a7ddda5d183ca790a214ee5981ef470ddb33/lib/tmp.js#L578

i just switch `app.getPath("userData");` to `os.tmpdir()` and should be the same.

danielweck:
Hotfix for bad indentation (linting error): https://github.com/readium/readium-desktop/commit/994b69cd5687f73567e31ec7f2dba36bccd7581c

PR #1001, opened 2020-04-12
[fix] lpf-converter

panaC:
- mime type
- deepClone additionalJson from w3cmanifest to rwpm
- inLanguage accept array
- set ES2019 lib to tsconfig.json

fixes from the Laurent code review


tsconfig.json
```diff
@@ -55,6 +55,8 @@
     "skipLibCheck": false,
     "module": "commonjs",
     "lib": [
+      "es2019",
+      "es2018",
```
danielweck, line 59: can you please explain what these additional libs bring to the transpiled es6 target build?

src/utils/deepClone.ts
```diff
@@ -0,0 +1,25 @@
...
+        {});
+
+const ObjFromEntries = Object.fromEntries || fromEntries;
+
+export const deepClone = (o: unknown): any => {
+
+    const _clone = (_o: unknown): Array<[string, unknown]> =>
+        Object.entries(_o).map<[string, unknown]>(
+            ([_key, _value]) =>
+                typeof _value === "object" ? [_key, _clone(_value)] : [_key, _value]);
```
danielweck, line 21: if typeof _value is array, potentially the array items can be objects, so need recursing too?

danielweck: sorry, my bad, from entries does that.

danielweck: actually i rest my case. i need to run sone tests.

src/main/lpfConverter.ts
```diff
@@ -100,110 +101,110 @@ async function injectManifestToZip(audiobookPath: string, manifest: Buffer) {
...
         const links = convertW3CpublicationLinksToReadiumManifestLink(readingOrder);
         if (links.length) {
             publication.Spine = links;
             // https://github.com/readium/r2-shared-js/blob/develop/src/models/publication.ts#L63
         }
     }
+    {
+
+        if (Object.keys(w3cManifest).length) {
+            publication.Metadata.AdditionalJSON = deepClone(w3cManifest);
```
danielweck, line 205: additionalJson is meant to contain the properties of the object, not the root itself. to be verified.

tsconfig.json
panaC: Object.fromEntries pardi ! :)

src/main/lpfConverter.ts
panaC: It contains all the root properties not yet parsed

src/utils/deepClone.ts
panaC: I will add some tests

tsconfig.json
danielweck: `Object.fromEntries()` is available in the current version of Electron used in Thorium (`8.2.1`), both in Node (v`12.13.0`) and in Chromium's V8 (v`8.0.426.30`).
See:
https://node.green/#ES2019-features
and:
https://github.com/mdn/browser-compat-data/blob/036d00b5853b8711d94a2b163dfba0fcafba06e2/javascript/builtins/Object.json#L703-L754

src/utils/deepClone.ts
```diff
@@ -0,0 +1,25 @@
...
+        (pv, [_key, _value]) =>
+            pv[_key] = typeof _value === "object" ? fromEntries(_value) : _value,
+        {});
+
+const ObjFromEntries = Object.fromEntries || fromEntries;
+
+export const deepClone = (o: unknown): any => {
+
+    const _clone = (_o: unknown): Array<[string, unknown]> =>
+        Object.entries(_o).map<[string, unknown]>(
```
danielweck, line 19: If `_value` is `null`, the program crashes.

`node -e "console.log(Object.entries(null))"`

(also tested with actual `deepClone()` call)

src/utils/deepClone.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const fromEntries = (it: Array<[string, any]>): any =>
+    it.reduce<any>(
+        (pv, [_key, _value]) =>
+            pv[_key] = typeof _value === "object" ? fromEntries(_value) : _value,
+        {});
+
+const ObjFromEntries = Object.fromEntries || fromEntries;
+
+export const deepClone = (o: unknown): any => {
```
danielweck, line 16: This function does not seem to work as intended, the clone is not structured as per the original object (I think `Object.fromEntries() should be called recursively too):

Repro test:

```javascript
const deepClone = (o) => {

    const _clone = (_o) =>
        Object.entries(_o).map(
            ([_key, _value]) =>
                typeof _value === "object" ? [_key, _clone(_value)] : [_key, _value]);

    const entries = _clone(o);
    return Object.fromEntries(entries);
};

const obj = {
    zbool: true,
    zstring: "str1",
    znumber: -1,
    zarray: [true, "str3", 111, {foo: "bar", deep: ["another array"]}, [1, 2, 3, {n:4}]],
    zobj: {
        zbool: false,
        zstring: "str2",
        znumber: 999,
        zarray: [false, "str5", 333, {num: 555, nest: [{nul: "null", und: undefined}]}, [1, 2, 3, {n:4}]],
    },
};
console.log(JSON.stringify(obj, null, 4));
console.log(JSON.stringify(deepClone(obj), null, 4));
```

Output:

```json
{
    "zbool": true,
    "zstring": "str1",
    "znumber": -1,
    "zarray": [
        true,
        "str3",
        111,
        {
            "foo": "bar",
            "deep": [
                "another array"
            ]
        },
        [
            1,
            2,
            3,
            {
                "n": 4
            }
        ]
    ],
    "zobj": {
        "zbool": false,
        "zstring": "str2",
        "znumber": 999,
        "zarray": [
            false,
            "str5",
            333,
            {
                "num": 555,
                "nest": [
                    {
                        "nul": "null"
                    }
                ]
            },
            [
                1,
                2,
                3,
                {
                    "n": 4
                }
            ]
        ]
    }
}
```json
```
{
    "zbool": true,
    "zstring": "str1",
    "znumber": -1,
    "zarray": [
        [
            "0",
            true
        ],
        [
            "1",
            "str3"
        ],
        [
            "2",
            111
        ],
        [
            "3",
            [
                [
                    "foo",
                    "bar"
                ],
                [
                    "deep",
                    [
                        [
                            "0",
                            "another array"
                        ]
                    ]
                ]
            ]
        ],
        [
            "4",
            [
                [
                    "0",
                    1
                ],
                [
                    "1",
                    2
                ],
                [
                    "2",
                    3
                ],
                [
                    "3",
                    [
                        [
                            "n",
                            4
                        ]
                    ]
                ]
            ]
        ]
    ],
    "zobj": [
        [
            "zbool",
            false
        ],
        [
            "zstring",
            "str2"
        ],
        [
            "znumber",
            999
        ],
        [
            "zarray",
            [
                [
                    "0",
                    false
                ],
                [
                    "1",
                    "str5"
                ],
                [
                    "2",
                    333
                ],
                [
                    "3",
                    [
                        [
                            "num",
                            555
                        ],
                        [
                            "nest",
                            [
                                [
                                    "0",
                                    [
                                        [
                                            "nul",
                                            "null"
                                        ],
                                        [
                                            "und",
                                            null
                                        ]
                                    ]
                                ]
                            ]
                        ]
                    ]
                ],
                [
                    "4",
                    [
                        [
                            "0",
                            1
                        ],
                        [
                            "1",
                            2
                        ],
                        [
                            "2",
                            3
                        ],
                        [
                            "3",
                            [
                                [
                                    "n",
                                    4
                                ]
                            ]
                        ]
                    ]
                ]
            ]
        ]
    ]
}
```

File: src/utils/deepClone.ts (hunk 2 above)
Comment (danielweck) [reply src/utils/deepClone.ts:21]: I added a minimal repro test in the comment above.

File: tsconfig.json (hunk 1 above)
Comment (panaC) [reply tsconfig.json:59]: I know

but without "es2019" :

`[2] ERROR in [at-loader] ./src/utils/deepClone.ts:14:31 
[2]     TS2339: Property 'fromEntries' does not exist on type 'ObjectConstructor'.
[2] 
[2] ERROR in [at-loader] ./src/main/lpfConverter.ts:153:53 
[2]     TS2339: Property 'fromEntries' does not exist on type 'ObjectConstructor'.`

Comment (danielweck) [reply tsconfig.json:59]: Ah yes, because the TypeScript compiler transpiles to ES6 in the current configuration, some constructs are modified in the output JS (I guess `Object.fromEntries` becomes `ObjectConstructor` something ...)

File: src/utils/deepClone.ts (hunk 5 above)
Comment (panaC) [reply src/utils/deepClone.ts]: thank you

File: src/utils/deepClone.ts (hunk 4 above)
Comment (panaC) [reply src/utils/deepClone.ts]: thank you

File: src/utils/deepClone.ts (hunk 2 above)
Comment (panaC) [reply src/utils/deepClone.ts]: i have fixed `deepClone` and added some tests
feel free to check again

File: src/main/lpfConverter.ts (hunk 3 above)
Comment (danielweck) [reply src/main/lpfConverter.ts:205]: Yes, this is fine, sorry.

File: test/deepClone.test.ts [hunk 6]
```diff
@@ -0,0 +1,170 @@
...
+    const object: any = {
+        zbool: true,
+        zstring: "str1",
+        znumber: -1,
+        zarray: [true, "str3", 111, { foo: "bar", deep: ["another array"] }, [1, 2, 3, { n: 4 }]],
+        zobj: {
+            zbool: false,
+            zstring: "str2",
+            znumber: 999,
+            zarray: [false, "str5", 333, { num: 555, nest: [{ nul: "null", und: undefined }] }, [1, 2, 3, { n: 4 }]],
```
danielweck, line 160: Note that `nul: "null"` is a string because `null` was crashing in my tests with the previous code. May I suggest to remove this unit test, as I just used it as an example to quickly test the deep nesting of value type, using `JSON.stringify(obj, null, 4)` to visualize the results. But frankly it is not a good "unit" test :)

test/deepClone.test.ts
```diff
@@ -0,0 +1,170 @@
...
+    const test = isCloned(expected, actual);
+
+    expect(expected).toEqual(actual);
+    expect(test).toBe(true);
+});
+
+test("should clone object with `date` value", () => {
+
+    // @ts-ignore
+    const object: any = { key: new Date() };
```
danielweck, line 136: In my tests (not unit tests, just NodeJS REPL from the command line), `Date` object becomes `{}` when cloned with the `deepClone()` function. Obviously, this doesn't matter when inputing data from raw JSON (where `Date` is not a primitive type), but the clone function should nonetheless not loose the stringified date?

src/utils/deepClone.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const deepClone = (_o: any): any =>
+        _o && typeof _o === "object"
+            ? Array.isArray(_o)
+                ? _o.map((_value) => deepClone(_value))
+                : Object.fromEntries(
+                    Object.entries(_o).map(
+                        ([_key, _value]) =>
+                            [_key, deepClone(_value)]))
+            : _o;
```
danielweck, line 16: Just a note: this last segment of the top-level ternary operator captures all "falsy" `_o` values (which is fine, but worth noting, and perhaps worth commenting in the code, for the avoidance of doubt for future readers / code reviewers): `undefined`, `null`, `""`, `[]`, `0` (as well as any "truthy" value that is not an `object` (including `Array`)).

src/utils/deepClone.ts
```diff
@@ -0,0 +1,16 @@
...
+// ==LICENSE-END==
+
+export const deepClone = (_o: any): any =>
+        _o && typeof _o === "object"
+            ? Array.isArray(_o)
+                ? _o.map((_value) => deepClone(_value))
+                : Object.fromEntries(
+                    Object.entries(_o).map(
+                        ([_key, _value]) =>
+                            [_key, deepClone(_value)]))
```
danielweck, line 15:
```
Object.fromEntries(
                    Object.entries(_o).map(
                        ([_key, _value]) =>
                            [_key, deepClone(_value)]))
```

=> we could easily remove the dependency on `Object.fromEntries` (for which the TypeScript compiler injects a ES2018/19 polyfill, as we target ES6), simply by using `reduce`:

`Object.entries(o).reduce((obj, [key, value]) => ({...obj, [key]: deepClone(value)}), {})`

test/deepClone.test.ts
danielweck: I used both `JSON.stringify()` and `util.inspect()` to discover this Date bug. Here is my test script:
```
const util = require("util");

const deepClone = (o) =>
    o && typeof o === "object"
        ? Array.isArray(o)
            ? o.map((value) => deepClone(value))
            : Object.fromEntries(
                Object.entries(o).map(
                    ([key, value]) =>
                        [key, deepClone(value)]))
        : o;
const deepClone_ = (o) =>
    o && typeof o === "object"
        ? Array.isArray(o)
            ? o.map((value) => deepClone_(value))
            : Object.entries(o).reduce((obj, [key, value]) => ({...obj, [key]: deepClone(value)}), {})
        : o;
const obj = {
    zund: undefined,
    zfunc1: (...args) => { console.log("hello", args); },
    zfunc2(...args) { console.log("hello", args); },
    zdate: new Date(),
    zbool: true,
    zemptyarr: [],
    zstring: "str1",
    znul: null,
    zsym: Symbol("zzz"),
    znumber: -1,
    zarray: [true, "str3", 111, {foo: "bar", deep: ["another array"]}, [[], 1, 2, 3, {n:4}]],
    zobj: {
        zbool: false,
        zstring: "str2",
        znumber: 999,
        zarray: [false, "str5", 333, {num: 555, zsymb: Symbol("yyy"), nest: [{nul: null, und: undefined}]}, [1, 2, 3, {n:4}]],
    },
};
console.log(JSON.stringify(obj, null, 4));
console.log(util.inspect(obj, { colors: true, depth: null, compact: false }));

console.log(JSON.stringify(deepClone(obj), null, 4));
console.log(util.inspect(deepClone(obj), { colors: true, depth: null, compact: false }));

console.log(JSON.stringify(deepClone_(obj), null, 4));
console.log(util.inspect(deepClone_(obj), { colors: true, depth: null, compact: false }));

```

src/utils/deepClone.ts
danielweck: PS: my little test script (pasted in a previous comment) can be used to check both implementations for equality.

tsconfig.json
panaC: indeed

src/main/lpfConverter.ts
danielweck, line 205: Ah! `w3cManifest` is *not* an immutable object, in fact keys are deleted from this simple JSON structure (which is simply parsed from string) inside this very function. As the object is created outside this function, in principle the code in the function body breaks encapsulation, however in practice `w3cManifest` is an implementation detail inside "private" code, so it does not matter. Consequently, there is no need to clone `w3cManifest`, it can simply be assigned: `publication.Metadata.AdditionalJSON = w3cManifest;`. I would therefore suggest dropping the "deep clone" ad-hoc implementation, and instead in the future use `lodash` or some other robust micro lib. Alternatively, where performance is not an issue, this works fine too in the JSON case (i.e. only primitive types with array and object composition): `const clone = JSON.parse(JSON.stringify(obj));` ;)

src/utils/deepClone.ts
panaC: nice Tip :) 👍

src/utils/deepClone.ts
panaC: I added a function helper.
feel free to correct it

test/deepClone.test.ts
panaC: i added a shallow copy by their copy constructor of Date, Set, Map

src/main/lpfConverter.ts
panaC: ah i just read your message
OK let's leave the ship :)

src/main/lpfConverter.ts
```diff
@@ -100,110 +100,110 @@ async function injectManifestToZip(audiobookPath: string, manifest: Buffer) {
...
-                {});
-            if (ObjectKeys(titleObj).length) {
-                publication.Metadata.Title = titleObj;
+            const nameObjArray = (Array.isArray(name) ? name : [name]) as IW3cName[];
+            const nameMap = nameObjArray.reduce(
+                (pv, cv: IW3cName) =>
+                    cv.value && cv.language ? pv.set(`${cv.language}`, `${cv.value}`) : pv,
+                new Map<string, string>());
+            if (nameMap.size) {
+                publication.Metadata.Title = Object.fromEntries(nameMap);
```
danielweck, line 152: Suggestion: get rid of `Object.fromEntries()` so that Thorium's output ES6 bundle does include TypeScript's ES2018/19 polyfill libs, and simply replace `new Map<string, string>()` with `{}` as the `array.reduce()` initial object, and of course instead of `pv.set(`${cv.language}`, `${cv.value}`)` just return `{ ...pv, [cv.language]: cv.value }` (i.e. the more recent key overrides the previous one, if any ... just like the `Map`).
Thoughts?

danielweck: Total code:
```javascript
publication.Metadata.Title = nameObjArray.reduce((pv, cv: IW3cName) => {
if (!cv.value || !cv.language) {
    return pv;
}
return { ...pv, [cv.language]: cv.value };
}, {});
```

danielweck: What is your analysis about `Object.fromEntries()` (which in my view, is quite a heavyweight solution) vs. `array.reduce()` (which I think is a more lightweight and elegant solution, yielding more readable code)

It is absolutely fine if you have a different opinion / coding preference :)
I just want to understand the merit of the proposed technical solution.

panaC: yes you're right. Your solution is much better - simpler
Thanks you

the running time should be quite similar.

danielweck: In principle, each `array.reduce()` iteration allocates a new object in the memory stack (or is it the heap / somewhere in the closure ... I'm not sure), but dealing with a `Map` and `Object.fromEntries()` to finalize the object from an array structure is costly too .. I guess.

danielweck:
Shall we merge this PR now? Looks good to me.

PR #1002, opened 2020-04-14
cover images, misc aspect ratios, fixes #993

### LINKED ISSUE #993: non-rectangular cover images in the bookshelf, different aspect ratios (particularly: square-ish audio cover images)
Because audiobook covers are square (usually at least), there is a large space between the 'Resume listening' label and the first line of covers. 

We'd like to make this space smaller. But will we always have square covers for audiobooks? nobody knows. Any idea on how to solve it?

danielweck:
(click to see full / native size, otherwise you will not see the borders)

## BEFORE:

ThoriumCoverImages_BEFORE

## AFTER:

ThoriumCoverImages_AFTER

### DIFF
--- a/src/renderer/assets/styles/myBooks.css
+++ b/src/renderer/assets/styles/myBooks.css
@@ -26,7 +26,9 @@
 		display: flex;
 		justify-content: space-between;
 		padding-top: 40px;
-		padding-bottom: 35px;
+		padding-bottom: 5px;
+		margin-bottom: 15px;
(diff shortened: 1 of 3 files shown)


danielweck:
Updated proposal (decreased header bottom margin, fixed first-cover-image right margin, created emphasis on vertical rhythm by introducing faded header underline):

ThoriumBookshelf

danielweck:
Shall we get this merged? It is a definite improvement on the current layout, we can always tweak it again later if necessary.

PR #1006, opened 2020-04-15
feat(l10n): update spanish translation

nvdaes:


danielweck:
tested and merged. thank you very much, Noelia!

PR #1009, opened 2020-04-15
[ready][fix] lpf-converter

panaC:


src/main/lpfConverter.ts
```diff
@@ -6,15 +6,16 @@
...
 
 import * as debug_ from "debug";
-import { copy, ensureDir, move } from "fs-extra";
-import { contentType } from "mime-types";
+import { promises as fsp } from "fs";
 import * as moment from "moment";
 import * as os from "os";
-import { basename, extname, join } from "path";
+import { basename, dirname, extname, join } from "path";
+import { Contributor } from "r2-shared-js/dist/es6-es2015/src/models/metadata-contributor";
```
danielweck, line 13: TODO: fix `r2-shared-js/dist/es6-es2015/` import (elsewhere as well, not just here)

panaC: ah damn
automatic import error

can you improve that ?

danielweck: Import path aliases are defined for both TypeScript and WebPack (see code references below).

Perhaps this is a bug in  Visual Studio Code?
https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

PS: "TypeScript Import Sorter" is the Visual Studio Code extension I personally use, but it does not seem to have an "autofixer" for incorrect import aliasing:
https://github.com/SoominHan/import-sorter

https://github.com/readium/readium-desktop/blob/6a6fad8faa33e5a1c517beda4057ad0f3e5f5314/tsconfig.json#L78-L95

https://github.com/readium/readium-desktop/blob/6a6fad8faa33e5a1c517beda4057ad0f3e5f5314/webpack.config.main.js#L16-L21

https://github.com/readium/readium-desktop/blob/6a6fad8faa33e5a1c517beda4057ad0f3e5f5314/webpack.config.renderer-library.js#L12-L17

https://github.com/readium/readium-desktop/blob/6a6fad8faa33e5a1c517beda4057ad0f3e5f5314/webpack.config.renderer-reader.js#L12-L17

src/main/lpfConverter.ts
```diff
@@ -201,24 +368,95 @@ function w3cPublicationManifestToReadiumPublicationManifest(w3cManifest: JsonMap
...
+            publication.Metadata.AccessibilityHazard = rawArrayFiltered;
+        }
+    }
+    {
+        const raw = pop("accessibilitySummary");
+        const loc = convertW3cLocalisableStringToReadiumManifestTitle(raw);
+        if (loc) {
+            const locMap = (Array.isArray(loc) ? loc : { und: loc }) as IStringMap;
+            const summary = Object.values(locMap);
+            publication.Metadata.AccessibilitySummary = summary;
```
danielweck, line 453: In `r2-shared-js`, the `schema:accessibilitySummary` EPUB2/3 metadata is now correctly parsed as a potentially-multilingual string (i.e. union type of plain `string`, or `IStringMap` with language keys and default to `und`).
My next commit to the develop branch will include the updated package.

src/main/lpfConverter.ts
```diff
@@ -201,24 +368,95 @@ function w3cPublicationManifestToReadiumPublicationManifest(w3cManifest: JsonMap
...
+                (pv, cv) => [
+                    ...pv,
+                    Array.isArray(cv?.itemListElement)
+                        ? cv.itemListElement.join(",")
+                        : undefined,
+                ], [])
+            .filter((f: any) => f);
+
+        if (rawArrayFormated.length) {
+            publication.Metadata.AccessModeSufficient = rawArrayFormated;
```
danielweck, line 428: In `r2-shared-js`, the `schema:accessModeSufficient` EPUB2/3 metadata is now parsed as comma-separated values, and normalized as an ordered list of non-repeating strings without any left/right whitespace (i.e. trimmed). There may be whitespaces within individual tokens, there is no automatic whitespace collapsing.
In short: `publication.Metadata.AccessModeSufficient` is now an array of array-of-strings.

My next commit to the develop branch will include the updated package.

danielweck: Okay, develop branch updated. The `itemListElement.join(",")` can now be removed.

https://github.com/readium/readium-desktop/commit/9e3b13a84e5ef0f3bdbb506e73c50e59719b1cb4

https://github.com/readium/r2-shared-js/compare/v1.0.34...v1.0.35

src/main/lpfConverter.ts
danielweck: The develop branch is now updated. `publication.Metadata.AccessibilitySummary` is now `string | IStringMap`, please update the conversion code accordingly (if there is only a single `accessibilitySummary` in the W3C document without specific language / default language, then `string` should be favoured instead of the language map).

https://github.com/readium/readium-desktop/commit/9e3b13a84e5ef0f3bdbb506e73c50e59719b1cb4

https://github.com/readium/r2-shared-js/compare/v1.0.34...v1.0.35

src/main/lpfConverter.ts
panaC: https://github.com/readium/readium-desktop/pull/1009/commits/31453cf6a1f6afcd723f92c27ee6dcdcfc0439bb

src/main/lpfConverter.ts
panaC: https://github.com/readium/readium-desktop/pull/1009/commits/31453cf6a1f6afcd723f92c27ee6dcdcfc0439bb

PR #1013, opened 2020-04-19
fix(a11y: don't use aria-pressed for show/hide keyboard

nvdaes:


nvdaes:
Using aria-pressed in show/hide keyboard settings produces inconsistent results, since this button ha different labels for show and hide.

danielweck:
Follow-up to commit https://github.com/readium/readium-desktop/commit/6a6fad8faa33e5a1c517beda4057ad0f3e5f5314

danielweck:
Thank you Noelia, I was wondering about this too. Maybe ARIA expanded could be useful in this particular case, but I am not sure where to place the attribute (button, or container div below?). We must make sure the screen reader announcement, or UI discovery as the user scans the linear structure of UI controls, is effective. Let me know please if you have additional thoughts.

nvdaes:
Hi Daniel, for me aria-expanded doesn't need to be used here since the button for show and hide has different labels. As an alternative, we may use just a label and for example aria-pressed. For me aria-expanded has more sense in menus.
I have some thoughts, though I don't know if I should open an issue for each concern. My concerns are:
- Use headings with appropiate levels, for example, starting with h1, then h2, etc in Thorium interface.`
``Use a close button in windows that can be closed or add proper documentation about the usage of Escape.
- Don't check the bookmark checkbox if the bookmark hasn't been added, and don't show the alert "Add button". Instead send a notification about the current state of the bookmark. Also, a message showing the reason causing the error adding a bookmark can be useful too, for example: "Bookmark added yet".
- Translate the wiki about keyboard shortcuts.
- Create tips for screen reader, explaining how the display (paginated or refloable) can change the visible text and how this can affect the possibility of adding bookmarks.
Let me know how i should proceed about these concerns. I could try to create PR for some of them (not sure). I don't want to break the visual appearance ofelements in case I createa PR for headings, for instance.

Issue #97, opened 2018-03-08
Code signing for Mac


danielweck:
As reported by @CircularKen : 

By default Mac will show

<img width="532" alt="Screenshot 2020-04-24 06 56 09" src="https://user-images.githubusercontent.com/17338941/80179800-2c038300-85f9-11ea-9fc9-5e25c80da77a.png">

Right Click on the app and Open to see

<img width="532" alt="Screenshot 2020-04-24 06 56 26" src="https://user-images.githubusercontent.com/17338941/80179791-2ad25600-85f9-11ea-9af1-012858f57d51.png">

danielweck:
Let's close this. Thorium is now signed for deployment into the Windows10 app store.

danielweck:
Actually, we still need to sign the code for MacOS and Windows NSI/MSI app installers (i.e. software not deployed through app store, that is to say via a trusted developer account with valid certificates).

So, @llemeurfr are there certificates for Thorium, for MacOS (which requires an Apple developer account), and for Windows (EV Code Signing Certificate, for example from DigiCert)?

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/125

llemeurfr:
No, we didn't buy this; it is true that getting no user warning for the .DMG would be good, and in this case for a .MSI also (even if most Windows users will get Thorium from the MS store). 
I propose to address this issue at the same time we address multi-platforms automatic builds via Jenkins or alike, for each released version. -> from August.

danielweck:
Useful links:

https://www.electron.build/code-signing

https://github.com/electron/electron-notarize
https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/
https://medium.com/@TwitterArchiveEraser/notarize-electron-apps-7a5f988406db

https://samuelmeuli.com/blog/2019-04-09-publishing-an-electron-app-on-the-mac-app-store/
https://david.dev/electron-cannot-build-mas-on-os-x-mojave

danielweck:
> nb: we have an Apple dev account.

It would be interesting to see if you could export EDRLab's Apple certificate containing the private key for signing DMG apps, so that the developer in charge of creating the build artefacts (at this point in time: me) can import it into his/her KeyChain.

danielweck:
Update: we now have a working setup for signing and notarizing Thorium for Mac OS, with LCP enabled. Next: Mac App Store deployment (which requires a different certificate and provisioning profile).

[...]

PR #1018, opened 2020-04-25
[PARKED] (DO NOT MERGE) support for MAS Mac App Store

danielweck:
Currently fails.
See: https://github.com/readium/readium-desktop/issues/97#issuecomment-618898849


CircularKen:
Daniel - I have recent experience of jumping through the latest Catalina notarisation hoops for Apple. If you need help let me know!

danielweck:
Thanks Ken. The problems we are experiencing are specific to Electron, but generally-speaking the MAS process can fail at various points in the build chain, and the error feedback from Apple's tooling / server is pretty sparse (dare I say, "useless" at times). Oh, and setting up the correct entitlements + provisioning profiles depending on the build type is pretty cumbersome :)

MAS is not a super high priority for Thorium right now, but we would like figure-out what the blockers are (if any). Thorium ships a couple of native NodeJS libraries which can be problematic, if improperly signed or handled in the sandbox context of a MAS app. There are also Electron APIs that simply do not pass the MAS compliance filter ("single instance lock", for example). Plus, the `readium-desktop` SDK launches its own HTTP(S) server on a localhost port, so there's that as well.

Fun stuff ;)

PR #1020, opened 2020-04-27
[fix] bad conformsTo url in lpfConverter

panaC:


danielweck:
Missing fix in the unit tests: https://github.com/readium/readium-desktop/blob/40ed7ac45f02aba5fd4f590c2890cf22f9f2cdcf/test/lpfConverter.test.ts#L128

PR #1032, opened 2020-05-01
[WIP][NEW FEATURE] text search in reader

panaC:


panaC:
Hello @danielweck 

some questions : 

To implement a text search in publication reader. How do we proceed ?

### first step : input : 
I think we have to catch the inner text from each html file in the publication.
In this case i think we have several way to do it : 
-  make a request to streamer for each html resources files in publication.
   - I tried to make a request to the streamer in thorium and i receive an empty body response. I don't really understand how works the access guard in streamer. (can you explain me, yet :), please )
 -  share text between publication reader iframe and reader browser view. The publication webview stream the publication and send to reader browser view process the publication text.
    - An issue will be to send to reader browser view process all html inner text from all html resources. To do a full search in publication all text must be scanned.
 - From thorium point of view consume an api from r2-navigator-js that returns all html inner text.
    - easiest solution that need I guess a new dev work in r2-navigator.
    - or simpler an API which returns only the search locator.

### second step : processing : 
I found a popular algorithm to search sub-string in big string.
https://en.wikipedia.org/wiki/Rabin–Karp_algorithm

I've started to implement it.

### third step : output :
We could connect the output of algorithm to your [highlight](https://github.com/readium/r2-navigator-js/blob/develop/src/electron/renderer/highlight.ts) API 

how work the highlighting API ? What is the best output format to connect highlight api ?

Tell me what do you think about it, how to you see things ?

what about LCP protections for accessing all internal html text in the thorium process outside the publication webview ?

panaC:
I close it because replaced with #1049

PR #1041, opened 2020-05-07
[fix #1039] map lib.close function on ctrl-q/command-q shortcut

### LINKED ISSUE #1039: "save session" dialog prompt only when closing window, not when quitting app using Apple-Q
https://github.com/readium/readium-desktop/blob/321c18b97bdee43f3619de085d2a733f5ad6ccc2/src/main/redux/sagas/win/session/library.ts#L20-L49

=>

https://github.com/readium/readium-desktop/blob/321c18b97bdee43f3619de085d2a733f5ad6ccc2/src/main/redux/sagas/win/library.ts#L245-L249

=>

https://github.com/readium/readium-desktop/blob/321c18b97bdee43f3619de085d2a733f5ad6ccc2/src/main/redux/sagas/win/library.ts#L126-L221

Screenshot 2020-05-06 at 18 53 59

panaC:
fix #1039

### DIFF
--- a/src/main/redux/sagas/app.ts
+++ b/src/main/redux/sagas/app.ts
@@ -8,7 +8,7 @@
 import * as debug_ from "debug";
 import { app, protocol } from "electron";
 import * as path from "path";
-import { diMainGet } from "readium-desktop/main/di";
+import { diMainGet, getLibraryWindowFromDi } from "readium-desktop/main/di";
 import { call } from "redux-saga/effects";
 
 // Logger
@@ -50,4 +50,12 @@ export function* init() {
         },
     );
 
+    // track command-q or ctrl-q
+    app.on("before-quit", (e) => {
+
+        e.preventDefault();
+
+        const libraryWin = getLibraryWindowFromDi();
+        libraryWin.close();
+    });
 }
--- a/src/main/redux/sagas/app.ts
+++ b/src/main/redux/sagas/app.ts
@@ -19,8 +19,8 @@ import { clearSessions } from "@r2-navigator-js/electron/main/sessions";
 
 import { streamerActions } from "../actions";
 import {
-    getBeforeQuitChannel, getQuitEventChannel, getWindowAllClosedEventChannel,
-} from "./getExitEventChannel";
+    getBeforeQuitEventChannel, getQuitEventChannel, getWindowAllClosedEventChannel,
+} from "./getEventChannel";
 
 // Logger
 const filename_ = "readium-desktop:main:saga:app";
@@ -73,7 +73,7 @@ export function* init() {
 export function exit() {
     return spawn(function*() {
 
-        const beforeQuitEventChannel = getBeforeQuitChannel();
+        const beforeQuitEventChannel = getBeforeQuitEventChannel();
         const windowAllClosedEventChannel = getWindowAllClosedEventChannel();
         const quitEventChannel = getQuitEventChannel();
         let shouldExit: boolean = process.platform !== "darwin" || IS_DEV;
@@ -89,14 +89,6 @@ export function exit() {
             beforeQuitEventChannel,
             (e: Electron.Event) => {
 
-                // track ctrl-q/command-q
-                shouldExit = true;
-
-                e.preventDefault();
-
-                const libraryWin = getLibraryWindowFromDi();
-                libraryWin.close();
-
                 debug("#####");
                 debug("#####");
                 debug("#####");
@@ -106,6 +98,15 @@ export function exit() {
                 debug("#####");
                 debug("#####");
                 debug("#####");
+
+                // track ctrl-q/command-q
+                shouldExit = true;
+
+                e.preventDefault();
+
+                const libraryWin = getLibraryWindowFromDi();
+                libraryWin.close();
+
             },
         );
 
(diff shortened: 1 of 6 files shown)


danielweck:
I think there are several problems (edge cases) related to Electron window vs. app lifecycle:

* `window-all-closed` is not necessarily called, so better use this only for non-critical UX-related behaviour, for example typically, a Mac app *can* be active without any windows at all, but in both `readium-desktop` and `r2-testapp-js` we make sure that "no windows" equates "no app" (in `r2-testapp-js` there is a user confirmation dialog, whereas in `readium-desktop` the app automatically closes). Related code snippets: https://github.com/readium/readium-desktop/issues/1039#issuecomment-624803690 Doc: https://www.electronjs.org/docs/api/app#event-window-all-closed
* `app.exit()` (which is used in `readium-desktop/src/main/redux/sagas/persist.ts`) is extremely problematic (as previously discussed for the CLI implementation), because the `[will/before-]quit` events are used in both `readium-desktop` and `r2-navigator-js` to perform graceful teardown of internal services (for example, the `r2-streamer-js` server must be shut down correctly, and the `r2-navigator-js` webview sessions must clear the appropriate caches). Doc: https://www.electronjs.org/docs/api/app#appquit
* `r2-navigator-js` currently relies on the `will-quit` event, which handler performs `preventDefault()` followed by `app.quit()`. Aside: note that `r2-testapp-js` uses `quit` to shut down the `r2-streamer-js` HTTP server gracefully just before the app exits.

So, we clearly need to review the Thorium's Electron lifecycle carefully in the coming days, to eliminate ambiguity and ensure that all services complete their teardown before the app process is terminated.

danielweck:
Related issue: https://github.com/readium/r2-navigator-js/issues/51

src/main/redux/sagas/index.ts
```diff
@@ -61,6 +61,9 @@ export function* rootSaga() {
     // send initSucess first
     yield put(appActions.initSuccess.build());
 
+    // watch all electon exit event
+    yield app.exit();
```
danielweck, line 65: This is extremely confusing for people reading / skimming the code (because of Electron's own `app.exit()`), much like when code uses "reserved" NodeJS packages for generic variable names like `path`, `os`, `fs`, etc.

So, would you mind swapping the `import { app as appElectron } from "electron"` around, for example `import * as appSaga from "./app";`? Thank you

src/main/redux/sagas/app.ts
```diff
@@ -50,4 +61,101 @@ export function* init() {
...
+
+                yield put(streamerActions.stopRequest.build());
+
+                yield race({
+                    a: take(streamerActions.stopSuccess.ID),
+                    b: take(streamerActions.stopError.ID),
+                });
+
+                if (shouldExit) {
+                    app.exit(0);
```
danielweck, line 140: As a rule of thumb, Electron apps should be cleanly shutdown using `app.quit()` not `app.exit()` (unless in CLI mode), because of subtle differences on Mac, Windows, Linux related to process lifecycle. What is the motivation for using `app.exit()` here?

src/main/redux/sagas/app.ts
```diff
@@ -50,4 +61,101 @@ export function* init() {
...
+                    call(clearSessions),
+                    call(needToPersistState),
+                ]);
+
+                yield put(streamerActions.stopRequest.build());
+
+                yield race({
+                    a: take(streamerActions.stopSuccess.ID),
+                    b: take(streamerActions.stopError.ID),
+                });
```
danielweck, line 137: In the Electron app lifecycle, `window-all-closed` would typically be used to differentiate a Mac app which can "survive" without any active window. As good practice, this event handler should not contain teardown logic, it should instead simply `app.quit()` (or not, for example on Mac). Then, the `[will|before]quit` events can be used to capture the teardown phase, for all platforms (Linux, Windows, Mac).
I would therefore recommend migrating any teardown logic (i.e. graceful shutdown of internal services) to one of the `quit` events. This way, all scenarios are covered: user closing windows one by one until there are no more, or user closing library view while there are reader windows open (which results in cascade-closing all reader windows), or force-closing using Apple-Q / Alt-F4.

src/main/redux/sagas/index.ts
panaC: fixed

src/main/redux/sagas/app.ts
panaC: i'm manually handle the thorium end cycle. Make a call to app.quit twice launches 'before-quit' and 'will-quit' event.
When the user press `ctrl-q`, electron call app.quit(). This function dispatch a 'before-quit' event and try to close every browser windows and finally close the main process.
In thorium the ctrl-q shortcut should be like a press of the library window close button. To catch and stop the propagation of app.quit process, I prevent default the event in 'before-quit' event. In this case I fully controls the end-cycle when a developer make a call to app.quit or an user close the app.

The end-cycle workd like this:
- library 'close' event is trapped
    - closes every render windows
- app event 'window-all-closed' is trapped
    - calls the thorium closed process
    - if 'before-quit' event has been previously called or os=linux/window -> forces app exit
    - else in mac wait app event 'appActivate' to recreate a library window.

panaC: i tried to answer, but tell me how you see thorium end-cycle ?

danielweck: Thank you for the clarification, I think your analysis covers all the use cases in Mac Windows and Linux.

PR #1054, opened 2020-05-18
cleaning install extension code

panaC:


src/main/redux/sagas/app.ts
```diff
@@ -7,6 +7,9 @@
 
 import * as debug_ from "debug";
 import { app, protocol } from "electron";
+import installExtension, {
+    REACT_DEVELOPER_TOOLS, REDUX_DEVTOOLS,
+} from "electron-devtools-installer";
```
danielweck, line 12: There is a little bit of "side effect" when using the top-level ES6 import, compared with the old `require()` call inside the `IS_DEV` condition:
https://github.com/MarshallOfSound/electron-devtools-installer/blob/d692149da53ef15d04b8c30e75f630f7f3cf0fd4/src/index.js#L9-L17

I wonder if WebPack (in the particular case of Thorium's build configuration) automatically eliminates dead code when bundling for production. We need to verify this before merging this PR. In other words, when `IS_DEV` is true, import-level "tree shaking" should eliminate the `electron-devtools-installer` import completely.

panaC: ah you're right.
I have not seen this side effects
thanks to report me

I close the PR, it doesn't matter to fix this.

PR #1057, opened 2020-05-20
[WIP] Search

### LINKED ISSUE #173: Publication Search engine
https://github.com/fergiemcdowall/search-index

panaC:
Fixes #173


src/utils/search/searchWithDomSeek.ts
```diff
@@ -0,0 +1,157 @@
...
+            (doc as any).getCssSelector,
+            (_node: Node) => ""); // computeElementCFI
+
+        searchResults.push({
+            textMatch: collapseWhitespaces(matches[0]),
+            textBefore,
+            textAfter,
+            rangeInfo,
+            href,
+            uuid: nanoid(),
```
danielweck, line 152: In my stress-tests I routinely generate many thousands (10-20) of search results, using large publications (many documents in the reading order), large documents (hundreds of "pages" each), and short frequently-occurring text tokens like "the" in the English language.

So, with that in mind, I am wondering whether the `uuid` of `ISearchResult` can simply be the concatenation of the stable short hash of `href`, and the incremental index `[0, n]` of the search result itself.

Rationale: I really like `nanoid` compared with `uuid.v4()`, but in this particular case I think we are wasting compute cycles for no good reason with randomly-generated IDs, because we can solidly rely on the ordered index of search results.

Thoughts?

danielweck: A parallel: in the React 'key' for iterated JSX fragments, we typically use the index, not a random number.

panaC: fix with https://github.com/edrlab/thorium-reader/pull/1057/commits/0f68469fe7a8bfe5d6b29ad61194040d961968f0

src/utils/counter.ts
```diff
@@ -0,0 +1,15 @@
+// ==LICENSE-BEGIN==
+// Copyright 2017 European Digital Reading Lab. All rights reserved.
+// Licensed to the Readium Foundation under one or more contributor license agreements.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END=
+
+const counter = () => {
+    let _counter = 0;
```
danielweck, line 9: I think `counter` should be reset to zero everytime a new search is performed. This will help debugging (i.e. checking DOM or other IDs), but also avoid super long numbers when running many searches in the same Thorium executable session.

panaC: I don't reset counter to 0 specially to avoid a collision. For example If the remove highlight procedure fail between 2 search word.
It can be also useful to have a unique id number in the window for the highlight annotation or whatever unique number.

panaC: I can also add a reset test on Number.MAX_SAFE_INTEGER condition

panaC: https://github.com/edrlab/thorium-reader/pull/1057/commits/e69a0c78448197d888cb6580631f67c947ca3058

src/utils/counter.ts
```diff
@@ -0,0 +1,15 @@
...
+
+    return () => {
+        return Number.isSafeInteger(++_counter) ? counter : _counter = 1;
```
danielweck, line 12: `counter` is a function in `? counter :`, is that intentional?

panaC: my bad

danielweck:
Quick review (UI UX testing):
* in search results list, before/after text has extraneous space character (I fixed this in my local branch)
* when using mouse to activate link in search results list, no automatic closing of nav panel (that's fine), but when using keyboard, should close and focus (just like the TOC, bookmarks, etc.) (I fixed this in my local branch)
* in search results list, activating the same link twice generates a duplicate highlight (weird color because of double overlay, should be coral), in fact resetting search results or finding a new keyword shows that the "ghost" / orphan highlight is still present.
* CTRL-F can be used whilst CTRL-N is active (navigation panel is open), unlike CTRL-B bookmark which is correctly deactivated.
* when mouse drag fast to move the search panel at the top of the document, drag is interrupted
* in search results list, using keyboard-tab to navigate the links is really slow (only when many DOM items are rendered => we really need to virtualize the list for large data sets)

panaC:
> in search results list, activating the same link twice generates a duplicate highlight (weird color because of double overlay, should be coral), in fact resetting search results or finding a new keyword shows that the "ghost" / orphan highlight is still present. same when clicking twice with the mouse.

I will fix this

> CTRL-F can be used whilst CTRL-N is active (navigation panel is open), unlike CTRL-B bookmark which is correctly deactivated.

https://github.com/edrlab/thorium-reader/blob/233283ed85d3f3f120cabee58e6200265d88a365/src/renderer/reader/components/ReaderHeader.tsx#L333-L340
https://github.com/edrlab/thorium-reader/blob/233283ed85d3f3f120cabee58e6200265d88a365/src/renderer/reader/components/ReaderHeader.tsx#L365-L370

Isn't this the cause of the problem?

> when mouse drag fast to move the search panel at the top of the document, drag is interrupted

effectively, like discussed together i'm "broke my teeth" on it. And you said me that you re-thinking a better design. Tell me later.

> in search results list, using keyboard-tab to navigate the links is really slow (only when many DOM items are rendered => we really need to virtualize the list for large data sets)

virtualise is a new concept for me. I've already memoize the rendering list, but not really virtualise. I will try to implement this in the future.

danielweck:
Over time, I found a few issues with this PR (PanaC branch), which I have fixed in my test PR (local EDRLab branch):
https://github.com/edrlab/thorium-reader/pull/1086

Latest issue, case-sensitive filesystem on Linux:
https://github.com/edrlab/thorium-reader/runs/909679407

danielweck:
This PR includes https://github.com/edrlab/thorium-reader/pull/1056 and https://github.com/edrlab/thorium-reader/pull/1055
Closing, as now superseded by https://github.com/edrlab/thorium-reader/pull/1086 (which includes the Git history from this PR's branch, no squashing, just a regular merge)

PR #1077, opened 2020-06-08
[fix #1075]

### LINKED ISSUE #1075: OPDS download is broken with Gutenberg feed
http://m.gutenberg.org/ebooks.opds/

(see https://readium2.herokuapp.com/opds-v1-v2-convert/http%3A%2F%2Fm.gutenberg.org%2Febooks.opds%2F )

```
[2]   readium-desktop:main#services/downloader oauth +0ms
[2]   readium-desktop:main#services/downloader NotFoundError { message: 'document not found' } +1ms
```

panaC:
fix download error on http://m.gutenberg.org/ebooks.opds/
but don't download it anyway

fixes #1075

### DIFF
--- a/src/main/services/downloader.ts
+++ b/src/main/services/downloader.ts
@@ -140,6 +140,9 @@ export class Downloader {
                 method: "GET",
                 encoding: undefined,
                 headers,
+                agentOptions: {
+                    rejectUnauthorized: false,
+                },
             },
         );
 
--- a/src/main/services/downloader.ts
+++ b/src/main/services/downloader.ts
@@ -234,6 +234,9 @@ export class Downloader {
                         outputStream.end(null, null, () => {
                             return resolve(download);
                         });
+                    } else {
+
+                        return reject("output stream error");
                     }
                 });
             });
@@ -253,6 +256,9 @@ export class Downloader {
                     outputStream.end(null, null, () => {
                         return reject(error);
                     });
+                } else {
+
+                    return reject(error);
                 }
             });
         });
--- a/src/main/services/downloader.ts
+++ b/src/main/services/downloader.ts
@@ -258,6 +258,7 @@ export class Downloader {
                     });
                 } else {
 
+                    debug("request error: ", error);
                     return reject(error);
                 }
             });


panaC:
https://github.com/Microsoft/vscode/commit/20ff90c932596961c710ae7d0441705694e2b0a9 ?

src/main/services/downloader.ts
```diff
@@ -140,6 +140,9 @@ export class Downloader {
                 method: "GET",
                 encoding: undefined,
                 headers,
+                agentOptions: {
+                    rejectUnauthorized: false,
```
danielweck, line 144: Why are we seeing this problem now, and not before? Has something changed in our code that triggers this new behaviour? Or is that a change in the HTTP servers?

src/main/services/downloader.ts
```diff
@@ -234,6 +237,9 @@ export class Downloader {
                         outputStream.end(null, null, () => {
                             return resolve(download);
                         });
+                    } else {
+
+                        return reject("output stream error");
```
danielweck, line 242: This should be `resolve()`, not reject (because even though `outputStream` is non-existent, we are still producing a valid `download` data structure).

danielweck: There is a `reject()` missing in `if (response.statusCode < 200 || response.statusCode > 299) {` ... I am implementing several bug fixes, I will cherry-pick the `agentOptions` / `rejectUnauthorized` solution from your PR.

src/main/services/downloader.ts
danielweck: This should really only be used in development mode, not for production (security hole with man-in-the-middle attacks, if we break the certificate validation chain).

danielweck: Answering my own question:
https://www.sslshopper.com/ssl-checker.html#hostname=https://lcp-prod.edrlab.org
"One of the root or intermediate certificates has expired (10 days ago)."

danielweck:
PR superseded by https://github.com/edrlab/thorium-reader/commit/4a1abf87bb23c3f73f551778dd8b94f5b2341fdb
(Thank you Pierre for putting me on the right track, the workaround is valid, but I correct the issues mentioned in this PR comments, and I added a few other checks)

PR #1094, opened 2020-06-11
create an exit function that compact db before to quit

panaC:
- enhance try-catch around clearSession and needToPersistState

@danielweck ?

### DIFF
--- a/src/main/di.ts
+++ b/src/main/di.ts
@@ -13,6 +13,7 @@ import { Container } from "inversify";
 import * as path from "path";
 import * as PouchDBCore from "pouchdb-core";
 import { Translator } from "readium-desktop/common/services/translator";
+import { PromiseAllSettled } from "readium-desktop/common/utils/promise";
 import { CatalogApi } from "readium-desktop/main/api/catalog";
 import { LcpApi } from "readium-desktop/main/api/lcp";
 import { OpdsApi } from "readium-desktop/main/api/opds";
@@ -156,6 +157,15 @@ if (!fs.existsSync(publicationRepositoryPath)) {
 // end of create database
 //
 
+// https://pouchdb.com/guides/compact-and-destroy.html
+const compactDb = () => PromiseAllSettled([
+    publicationDb.compact(),
+    opdsDb.compact(),
+    configDb.compact(),
+    locatorDb.compact(),
+    lcpSecretDb.compact(),
+]);
+
 //
 // Depedency Injection
 //
@@ -287,6 +297,7 @@ interface IGet {
 const diGet: IGet = (symbol: keyof typeof diSymbolTable) => container.get<any>(diSymbolTable[symbol]);
 
 export {
+    compactDb,
     diGet as diMainGet,
     getLibraryWindowFromDi,
     getReaderWindowFromDi,
--- a/src/main/di.ts
+++ b/src/main/di.ts
@@ -158,13 +158,32 @@ if (!fs.existsSync(publicationRepositoryPath)) {
 //
 
 // https://pouchdb.com/guides/compact-and-destroy.html
-const compactDb = () => PromiseAllSettled([
-    publicationDb.compact(),
-    opdsDb.compact(),
-    configDb.compact(),
-    locatorDb.compact(),
-    lcpSecretDb.compact(),
-]);
+const compactDb = async () => {
+    const res = await PromiseAllSettled([
+        publicationDb.compact(),
+        opdsDb.compact(),
+        configDb.compact(),
+        locatorDb.compact(),
+        lcpSecretDb.compact(),
+    ]);
+
+    const done = res.reduce((pv, cv) => pv && cv.status === "fulfilled", true);
+    if (!done) {
+        throw JSON.stringify(res);
+    }
+};
+
+const closeProcessLock = (() => {
+    let lock = false;
+
+    return {
+        get isLock() {
+            return lock;
+        },
+        lock: () => lock = true,
+        release: () => lock = false,
+    };
+})();
 
 //
 // Depedency Injection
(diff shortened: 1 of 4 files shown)


src/main/redux/sagas/app.ts
```diff
@@ -87,6 +87,20 @@ export function exit() {
...
+
+            // clean the db just before to quit
+            compactDb()
+                .then(() => {
+
+                    debug("EXIT NOW");
+                    app.exit(0);
+                })
+                .catch(() => {
+                    // ignore
```
danielweck, line 100: missing app.edit(0) ?

panaC: catch won't happen but you're right

src/main/redux/sagas/app.ts
```diff
@@ -78,7 +78,7 @@ export function exit() {
         const shutdownEventChannel = getShutdownEventChannel();
         const windowAllClosedEventChannel = getWindowAllClosedEventChannel();
         const quitEventChannel = getQuitEventChannel();
-        let shouldExit: boolean = process.platform !== "darwin" || IS_DEV;
+        let shouldExit = process.platform !== "darwin" || IS_DEV;
```
danielweck, line 81: shall we remove the dev condition here, so that prod and dev use the exact same teardown logic?

panaC: i just a preference to kill app in dev. If the app stay in dock i should make a CTRL-C in dev console. In dev we make many re-start.  Are you agree ?

danielweck: ok

src/main/redux/sagas/app.ts
```diff
@@ -87,6 +87,18 @@ export function exit() {
         - quit
         */
 
+        const exitNow = () => {
+
+            // clean the db just before to quit
+            // tslint:disable-next-line: no-floating-promises
+            compactDb()
+                .finally(() => {
```
danielweck, line 95: I would like to suggest capturing all possibilities (note that the SQLite3 DB is *not* compacted even after the successful `compactDb()` `Promise.all()`):
```
            compactDb().then(() => {
                debug("Database(s) compacted successfully.");
            }).catch((err) => {
                debug("Database(s) compact failed:");
                debug(err);
            })
            .finally(() => {
                debug("EXIT NOW");
                app.exit(0);
            });
```

danielweck: I am now trying with `npm run start` to test LevelDown (SQLite3 is `npm run start:dev`).

danielweck: Tip: to test the `config` DB uncontrolled growth, simply play an audiobook for a few minutes :)

danielweck: Interesting: with LevelDown the config's `*.log` file grows (several MBs), and also does not get clean after `compact()`.

danielweck: The compaction code is invoked successfully:
https://github.com/rh389/pouchdb-adapter-websql-core/blob/e58ceaf4b73b7afcaffe695a76d88d1109c834af/src/index.js#L977-L1003

No `websqlError(callback)`, and the final `callback();` is invoked.

danielweck: Ah, the actual compaction routine is never called! `digestsToCheck` is empty:
https://github.com/rh389/pouchdb-adapter-websql-core/blob/e58ceaf4b73b7afcaffe695a76d88d1109c834af/src/utils.js#L92-L98

danielweck: The only `DELETE` is see when debugging is:
https://github.com/rh389/pouchdb-adapter-websql-core/blob/e58ceaf4b73b7afcaffe695a76d88d1109c834af/src/utils.js#L142-L144
...but the actual `config` SQLite3 DB file remains big.

danielweck: Okay so after analysing the SQLite3 database with a browser app, I can see that the `by-sequence` and `document-store` tables have been cleaned-up successfully, in the sense that there are no backup copies of the JSON data, but instead there are many `status` = `missing` references (and only a single `available` instance). I imagine that what takes space inside the binary file is the indexes? Not sure. At any rate, I am happy to merge this PR (once the `.then()` and `.catch()` are added)

panaC: 👍

src/main/di.ts
```diff
@@ -156,6 +157,34 @@ if (!fs.existsSync(publicationRepositoryPath)) {
...
+
+const closeProcessLock = (() => {
+    let lock = false;
+
+    return {
+        get isLock() {
+            return lock;
+        },
+        lock: () => lock = true,
+        release: () => lock = false,
```
danielweck, line 184: there can be race conditions on the lock/release cycles, so instead I would suggest implementing a mutex for the critical section based on an incremental lock index, for example https://github.com/mgtitimoli/await-mutex/blob/master/src/mutex.js

...or an abstraction with try/finally to ensure the lock is released:
```
export class Mutex {
  private mutex = Promise.resolve();

  lock(): PromiseLike<() => void> {
    let begin: (unlock: () => void) => void = unlock => {};

    this.mutex = this.mutex.then(() => {
      return new Promise(begin);
    });

    return new Promise(res => {
      begin = res;
    });
  }

  async dispatch(fn: (() => T) | (() => PromiseLike<T>)): Promise<T> {
    const unlock = await this.lock();
    try {
      return await Promise.resolve(fn());
    } finally {
      unlock();
    }
  }
}
```
see full article source: https://spin.atomicobject.com/2018/09/10/javascript-concurrency/

danielweck: or if you want to go crazy on process reentrance, etc.
https://github.com/rogierschouten/async-lock/blob/master/lib/index.js
(this one is an NPM package)

panaC: https://github.com/edrlab/thorium-reader/pull/1094/commits/36997799e397fd00c5398eb18117f2d88585266e

PR #1096, opened 2020-06-11
fix #1095: LCP RENEW with toast error

### LINKED ISSUE #1095: LCP loan extension refused, no user feedback
If a loan extension is refused by the server (e.g. because the loan maximum end date has been reached), the user has no visual indication of that fact.

panaC:


danielweck:
Note: I have added the same toast notifications for LCP-LSD `return`: https://github.com/edrlab/thorium-reader/commit/a372e6a38e2dff65427934a97c960d8d00ea385d

PR #1093, opened 2020-06-11
feat(l10n): Japanese translation (original PR #1034)

danielweck:
Full credits to the original PR author @peaceroad : https://github.com/edrlab/thorium-reader/pull/1034

PR #1099, opened 2020-06-13
feat(l10n): Upadate Japanese translation (locales/ja.json)

peaceroad:
Pointed out in #1093 , I updated the Japanese translation of the missing label.
In addition, I fixed translation mistakes


danielweck:
This is great, thank you! :)

Issue #1014, opened 2020-04-19
Hide useless elements for screen readers


nvdaes:
Hide elements such as:
- SVG like Import publication, bookmark and other elements which can be available as buttons, for a more consistent and predictable operability and less verbosity with screen readers.
- Hide the toolbar, which cannot be accessed and produces a wrong tree of landmarks with Elements list dialog in NVDA (NVDA+f7).

nvdaes:
Looking at the code, seems that there is a div labelled as toolbar with role region, and this div has even the main div as children. Also, there is a link with role region as the first child of this div... In short, seems that aria labels and roles should be improved. Now this produces too verbosity with NVDA, not always making sense. I think that regions should be restructured.

danielweck:
Hello Noelia, I believe the UI controls accessibility has greatly improved in recent builds, we fixed labelling issues in the toolbar notably. I am closing this issue but of course feel free to re-open to explain what still doesn't work well, if anything <smile> Regards, Daniel

nvdaes:
Thanks Daniel. I"ll open a new issue if I have any suggestion.

PR #1101, opened 2020-06-13
[WIP] (NEEDS REVIEW) Improve UX for screen readers

nvdaes:
- Use h1 (don't start with h2) in the library.
- Hide label for Import publication since this appears after the button..

Related issue: #1014


nvdaes:
I have hidden bookmark svg and checkbox since screen reader need to use the cursor to locate these controls, so the locator is moved and for us just the keyboard shortcut (`control+b`) is useful, and the checkbox and graphic can be anoying or useless.
@danielweck , you can review, test and merge if you want and, if needed, I may create incremental PR if I find more issues with screen readers. Be aware that I have tested this just with NVDA.
Thanks

nvdaes:
Unfortunately, seems Travis check is failing. I have committed and updated dependencies from Windows. Not sure if this can be affecting.
Thanks

nvdaes:
I have reverted a commit causing Travis failure. In case you are interested, please review and merge if appropriate. Import publication graphic, shown below the button, is hidden to screen readers. Also, now heading level 2 are replaced by heading level 1, since generally webpages shouldn't start with heading level 2.
Thanks.

PR #1103, opened 2020-06-16
Import prompt pronoun

mattgarrish:
Minor nit, but when you drag and drop files, the prompt asks you to confirm "those files", but there is already a specific list to be confirmed. Should be "these".


danielweck:
Thanks Matt :)

PR #1128, opened 2020-07-17
divina integration

panaC:


src/common/isManifestType.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const isAudiobookFn = (item: TIsRdfType | undefined) =>
+    (item?.RDFType && /http[s]?:\/\/schema\.org\/Audiobook[\/]?$/.test(item?.RDFType)) ||
+    (item?.Metadata?.RDFType && /http[s]?:\/\/schema\.org\/Audiobook[\/]?$/.test(item?.Metadata?.RDFType));
+
+export const isDivinaFn = (item: TIsRdfType | undefined) =>
+    (item?.Metadata?.RDFType &&
+        (/http[s]?:\/\/schema\.org\/ComicStrip[\/]?$/.test(item?.Metadata?.RDFType) ||
+            /http[s]?:\/\/schema\.org\/ComicStory[\/]?$/.test(item?.Metadata?.RDFType) ||
+            /http[s]?:\/\/schema\.org\/VisualNarrative[\/]?$/.test(item?.Metadata?.RDFType))) ||
```
danielweck, line 25: What about `VisualArtwork`? (I'm not sure, but this is currently used in `r2-shared-js`, so I need to align the vocabularies)
Note that further down the line, there will be a stricter canonical definition of the "Divina" publication type, which means we will be able to remove the "type sniffing" hacks.

danielweck: Code to align: https://github.com/readium/r2-shared-js/blob/develop/src/parser/divina.ts

danielweck: Discussion context: https://github.com/readium/webpub-manifest/issues/65#issuecomment-672859394

See also: https://github.com/readium/architecture/issues/134

PR #1138, opened 2020-07-30
[WIP] (NEEDS REVIEW) Improve toast

nvdaes:
This is not finished and will need improvements from developers. For now:

1. toastManager is a div with a role of "complementary", so that it can be reviewed with cursor by screen readers.
2. Button of toast is aria-hidden="true" since it is useless and doesn't have any effect with screen readers.
3. In order to review the notifications, they won't be closed in five secconds.


nvdaes:
For now I have finished this work. Future improvements like the ability to move the focus to the notifications region, maybe with the posibility (as an independent feature) of chosing if screen reader users prefer or not to save the previous position to came back quickly (though with NVda this can be achieved with the review cursor), may be added later.
for any reason checks are failing, but I managed to install Visual Studio 2017 on Windows and I have tested this.

nvdaes:
I have removed my changes trying to make messages translatable, so that all checks can be successful. If you are interested, please review this and add the needed modifications to translate aria labels or let me know how to do it.

PR #1142, opened 2020-08-08
replace request with node-fetch

panaC:

PR #1145, opened 2020-08-11
[1] replace request with node-fetch (#1142)

panaC:


danielweck:
This PR partially fixes https://github.com/edrlab/thorium-reader/issues/970
(there are other uses of `request` in the codebase, including R2 packages)

panaC:
#1161

PR #1161, opened 2020-09-05
[1-4] downloader & w3c-audiobooks & packager

panaC:
fix: 
- cli import https://github.com/edrlab/thorium-reader/blob/bfd2e9102e3d25456be12978d8eb2effcad7465b/src/main/cli/process.ts  https://github.com/edrlab/thorium-reader/pull/1161/commits/bfd2e9102e3d25456be12978d8eb2effcad7465b


.editorconfig
```diff
@@ -11,8 +11,6 @@ indent_style = ignore
 indent_size = ignore
 
 [*.{js,ts,jsx,tsx}]
-indent_style = space
-indent_size = 4
```
danielweck, line 15: Could you please add an exclude glob to `eclint` instead? https://github.com/edrlab/thorium-reader/blob/c66a2a4572a9ee26be84a98d142a2241cace0d38/package.json#L32

panaC: done https://github.com/edrlab/thorium-reader/pull/1161/commits/99c61226748afbb8e18439757670d0240fa486dc

panaC:
#1168

Issue #1136, opened 2020-07-28
Error when trying to run 1.4.0 on Windows


ways2read:
Windows 10 version 2004, on a machine that has run various editions of Thorium Reader previously.

Install from the GitHub download 1.4.0 https://github.com/edrlab/thorium-reader/releases/tag/v1.4.0

On first run and subsequently, I get the Internal Thorium error in main:saga:i18n

![image](https://user-images.githubusercontent.com/9092859/88681006-03ec5a00-d0e9-11ea-8061-a0ff6bf15235.png)
 
Uninstalled and ran the 1.4.1 alpha withou difficulty. Reinstalled 1.4.0 and got the same error.

danielweck:
I have just tested Thorium version 1.4 from the Windows Store, with existing books in the database, and then also after deleting Thorium's entire configuration folder in order to emulate an "install from scratch" (i.e. without any prior Thorium launches)
I am not able to reproduce this bug. I am now going to try the standalone Windows installer again (I did test the build artefact before uploading the release, but let's try again ...).

danielweck:
No, I am afraid I have been unable to reproduce this bug.
Both Thorium 1.4 from the Windows Store and as a standalone installer have their settings + database in:
`C:\Users\ME\AppData\Roaming\EDRLab.ThoriumReader`
The automated build Thorium 1.4.1-alpha.1.XXX also share the same folder, but inside it there are separate locations for the settings and database (`db-dev-sqlite` and `publications-dev` instead of `db` and `publications`).

Richard, if you are so inclined, you may try to delete (or move) the `db` and `publications` folders, and launch Thorium 1.4 standalone installer again.

ways2read:
I deleted everything under C:\Users\ME\AppData\Roaming\EDRLab.ThoriumReader and reinstalled 1.4.0, and now everything is OK.

The language switching in read aloud works fine also.

Richard

danielweck:
We have another bug report with this exact same “i18n” error message. It appears this related to the user interface locale, which defaults to the user’s operating system language, if I remember correctly.

llemeurfr:
A new identical bug report: Thanks a lot for the wonderful reading tool of Thorium Reader. Just to report that the GitHub Windows version 1.3.1 does not display spacing between paragraphs; and that when I tried to install v. 1.4.0, an error message popped up: "Internal Thorium Error, An error has occurred in readium-desktop:main:saga:i18n, Error message: {}".

I have asked the Windows locale.

[...]

PR #1165, opened 2020-09-09
potential fix: i18n error on thorium startup after upgrade

panaC:
#1136


danielweck (approved):
This seems like a reasonable safeguard. Let's merge.

PR #1182, opened 2020-09-22
[WIP] PDF support in reader window

panaC:


src/renderer/reader/components/ReaderFooter.tsx
```diff
@@ -23,6 +23,7 @@ import {
 import { LocatorExtended } from "@r2-navigator-js/electron/renderer/index";
 import { Locator as R2Locator } from "@r2-shared-js/models/locator";
 import { Publication as R2Publication } from "@r2-shared-js/models/publication";
+import { Link } from "r2-shared-js/dist/es6-es2015/src/models/publication-link";
```
danielweck, line 26: Alias `@r2-shared-js/models/publication-link`

src/renderer/reader/pdf/common/eventBus.ts
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+    const removeKey = <TKey extends string>(key: TKey) => {
+
+        delete event[key];
+    };
+
+    return {
+        subscribe,
+        dispatch,
```
danielweck, line 80: Suggestion: `dispatch: rx,` (and remove `const dispatch = tx;`)?

PR #1188, opened 2020-09-29
PDF reader

panaC:
**pdf support in thorium reader** 

pdf and lcp pdf files are accepted.

when importing a pdf file, thorium parses the pdf file and generates a .webpub zip file with the cover and the readium publication. Same procedure as the .lcpdf generator.
save under this directory for the main process : https://github.com/edrlab/thorium-reader/tree/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/main/pdf

thorium can import a pdf from an opds feed

the reader window launch the pdfReader host mounting point with a webview and an event bus on IPC.

pdfReader is sandboxed into a chromium webview. It contains the pdfJs library, an eventBus, and the bridge between both api: pdfjs and eventBus UI.

pdfReader logic is saved under : https://github.com/edrlab/thorium-reader/tree/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf/webview
pdfReader host mounting point is saved in this file: https://github.com/edrlab/thorium-reader/blob/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf/driver.ts

All thorium UI is mapped on the [event bus](https://github.com/edrlab/thorium-reader/blob/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf/common/eventBus.ts) ([typing](https://github.com/edrlab/thorium-reader/blob/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf/common/pdfReader.type.ts))
readerConfig doesn't save the configuration for the pdf.
pdf config allows to set: 
- the zoom (scale) : works great but need to tuned the scale value https://github.com/edrlab/thorium-reader/blob/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf/webview/pdfReader.ts#L465-L472
- display (page, column) : the logic need to be implemented in pdfReader. Event bus connected and dispatched in pdfReader (https://github.com/edrlab/thorium-reader/blob/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf/webview/pdfReader.ts#L474-L480
https://github.com/edrlab/thorium-reader/blob/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf/webview/pdfReader.ts#L482-L488
)

pdf in reader link : https://github.com/edrlab/thorium-reader/tree/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf

the webview folder src/renderer/reader/pdf/webview (https://github.com/edrlab/thorium-reader/tree/fa17ec90f5c15091f4b463a930a303610fd28dc4/src/renderer/reader/pdf/webview) contains : 
- index_pdf.ejs : html template file loaded to the webview source url
- index_pdf.ts : pdfReader mounting point loaded to the webview preload script
- pdfReader.ts : pdfReader logic with pdfJs

eventBus API : (typed in src/renderer/reader/pdf/common/pdfReader.type.ts)
- `"page": (pageNumber: number) => any;` : set the page to display
- `"scale": (scale: IPdfPlayerScale) => any` : `IPdfPlayerScale = "fit" | "width" | "50" | "100" | "150" | "200" | "300" | "500";` scale mapped to zoom value in UI.
- `"view": (view: IPdfPlayerView) => any;` : ` "scrolled" | "paginated";`
- `"column": (column: IPdfPlayerColumn) => any;` : `IPdfPlayerColumn = "auto" | "1" | "2";`
- `"search": (searchWord: string) => any;` : search request (not connected to the UI yet)
- `"search-next": () => any`
- `"search-previous": () => any;`
- `"page-next": () => any`
- `"page-previous": () => any`
- `"ready": (toc?: TToc) => any` and `"start": (pdfPath: string) => any` : event dispatched between the reader UI and the pdfReader webview 

cover : takes 7 seconds to screenshot the first page with pdfium in a hidden electron browser window

**what currently works :** 

- pdf one page view with different scale (page fit and page width, ...)
- move forward and backward in the pdf with the arrows.
- move to any page with the footer bar.
- bookmarks a page
- go to page in reader menu
- table of contents mapped to pdf outlines (few pdf implement it)
- pdf metadata information with cover
- text selection with pdfjs plugin
- link annotation with pdfjs annotation layer plugin (support internal link to a pdf page)

**what currently should be improved**

- 2 page view (display menu) (not connected in pdfReader) https://github.com/edrlab/thorium-reader/issues/1190
- scrolled or paginated view (display menu) (not connected in pdfReader) https://github.com/edrlab/thorium-reader/issues/1191
- scale view needs to be improved to fit with the right percentage (currently random values) https://github.com/edrlab/thorium-reader/issues/1192
-- https://github.com/edrlab/thorium-reader/pull/1188#issuecomment-700834600
- link annotations: https://github.com/edrlab/thorium-reader/pull/1188#issuecomment-700836083 https://github.com/edrlab/thorium-reader/issues/1193
- search in pdfJs not connected in pdfReader and not connected to the reader view UI (fixes in the thorium search PR) https://github.com/edrlab/thorium-reader/issues/1194
- accessibility
- cover rendering
- pdf number of pages : extracted into the main process and injected in the readium web publication metadata. 2 issues : 
pdfReader in the webview don't share this information. And with lcpdf this information doesn't exist in the manifest. 
- _Daniel:_ no CRC check for identical PDF publications => duplicates imported in bookshelf
- _Daniel:_ the left/right keyboard arrow keys should invoke the "page turn" command, when the keyboard focus is inside or outside the PDF webview.
- _Daniel:_ extract PDF cover image from encrypted PDF (.lcpdf file import, or imported indirectly via .lcpl)
- _Daniel:_ for LCP-protected PDFs, the clipboard copy interceptor is not implemented in the webview, so the LCP limitation (aka "right") is not respected.


src/renderer/assets/styles/reader-app.css
```diff
@@ -2,6 +2,34 @@
...
+}
+
+#annotation-layer > .linkAnnotation > a {
+  position: absolute;
+  font-size: 1em;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+} */
```
danielweck, line 31: pdfReader.ts is code executed in sandboxed webview, so this “reader” CSS does not reach inside the DOM created by PDF.js (canvas, etc.)

danielweck: alternative solution: inject styles into webview DOM via preload script, or even more easily, via the EJS HTML template

panaC: but of course !

danielweck:
Bug: page "fit" and "width" are inverted in the menu.

danielweck:
Bug: click "more information" for the bottom navigation / scrub bar, then use mouse of previous/next buttons to reach first page, and last page => the offset is not correct. Is this because of zero-based, versus one-based?

danielweck:
TODO: for LCP-protected PDFs, the clipboard copy interceptor is not implemented in the webview, so the LCP limitation (aka "right") is not respected.

danielweck:
TODO: the left/right keyboard arrow keys should invoke the "page turn" command, when the keyboard focus is inside or outside the PDF webview.

danielweck:
TODO: no CRC check for identical PDF publications => duplicates imported in bookshelf

danielweck:
TODO(?): the PDF "Anatomy of an Election" provides an TOC outline. However, when clicking on (sub)chapter headings that reference text fragments inside a page, the correct page is displayed but there is no scrolling-into-view (+ highlight) of the actual text referenced. In my opinion this is an optional feature / stretch goal, we should not focus on this at this early implementation stage.

danielweck:
Feature improvement: when changing zoom level, there is currently a line of code that resets the X coordinate to zero. Instead, the scroll top/left should be reset, to make the top-left corner of the page visible (in other words, changing the zoom currently preserves the scroll offsets, which does not place the viewport in a satisfactory / useful location).

danielweck:
Bug: in page 341 of the "Anatomy of an Election" PDF, there are many hyperlinks which trigger a "hand" mouse cursor, implying they can be clicked. Nothing happens when clicked. I would expect an external web browser to open the clicked links.

danielweck:
TODO: extract PDF cover image from encrypted PDF (`.lcpdf` file import, or imported indirectly via `.lcpl`)

danielweck:
There are some serious issues with the localisation, I am fixing in this PR...

danielweck:
I added the missing PDF section in library / bookshelf listings.

src/renderer/reader/components/Reader.tsx
```diff
@@ -777,7 +831,7 @@ class Reader extends React.Component<IProps, IState> {
             return;
         }
 
-        this.navLeftOrRight_(isPrevious, true);
```
danielweck, line 780: This seems wrong ... the whole point of `onKeyboardSpineNavigationPreviousNext()` is that the navigation should be at the spine level, not at the internal pagination or scrolling level. @panaC do you remember why you introduced this change in logic?

danielweck: Rectified this via https://github.com/edrlab/thorium-reader/pull/1188/commits/60187f714dfe645c98b7f1afc7a533c44f1d403d

...but still would like to understand the rationale for the code modification, just in case there was a utility to it in the context of this PDF PR.

danielweck:
Support for both internal and external links now fixed: https://github.com/edrlab/thorium-reader/pull/1188/commits/f2823368f37933ddbb3accdd8035663202a7f5c2

danielweck:
The keyboard focus outline is broken in the entire reader UI for PDF docs (works okay with EPUB, etc.).

PR #1202, opened 2020-10-08
fixes #1201

### LINKED ISSUE #1201: The window disappears if opened on a secondary monitor in a different position
I tried to look for previousely opened issues but i was unable to find any.
This issue can be reproduced easily with the following steps:

- Open the app, move the window on a secondary screen
- Close the app
- Disconnect the second monitor OR move it via screen settings on another position (if it was above the main monitor, move it on the right, left or bottom)
- Open the app again

The window will not be reachable until the screen is either connected/moved again on the same position or the user settings are deleted manually. 

A similar issue can be found for Github desktop app ( [/desktop/desktop/issues/2107] ).

panaC:
tested on mac and windows. that solved the issue

fixes #1201


panaC:
ready to merge

PR #1248, opened 2020-11-02
accept webpub content-type in supportedFileTypeLinkArray

panaC:


danielweck:
Why "webpub unpacked", isn't that "packed" instead? (`.webpub` file extension)

danielweck:
Sorry, I am being stupid. There is a separate entry for the "packed" WebPub variant:
https://github.com/edrlab/thorium-reader/blob/68f3195117590635495d69faccea86c95370b84a/src/utils/content-type.ts#L27-L30

panaC:
you're right i added both

PR #1272, opened 2020-12-01
Library Simplified Vendor Specs Implementation 

saad-kh:
# Brief

The Library Simplified suite of tools and software relies on the OPDS v1.x specs to serve its catalog of content. 
However to implement some custom behavior it uses its own vendor specifications on top of the OPDS v1.x format.
To be able to consume Library Simplified powered content, Thorium needs to implement some of those specs.

# Use Cases
- Handle Library Simplified authentication requirements.
    - [x] Spec: Implement the Library Simplified SAML Authentication workflow.
    - [x] Additional: Validate that the common authentication behaviors and procedures work well with this workflow.
    - Resources: 
        - [Library Simplfied Authentication-For-OPDS-Extensions](https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions)
- Handle Library Simplified additional data entries.
    - [x] Spec: Implement the parsing, display and behavior of specific Library Simplified data entries.
    - Resources: 
        - [Understanding the Fulfillment State of a Book](https://github.com/NYPL-Simplified/Simplified/wiki/OPDS-For-Library-Patrons#understanding-the-fulfillment-state-of-a-book)
        - [Detailed semantics for the http://opds-spec.org/acquisition/borrow relation](https://github.com/NYPL-Simplified/Simplified/wiki/OPDS-For-Library-Patrons#detailed-semantics-for-the-httpopds-specorgacquisitionborrow-relation)
        - [The http://librarysimplified.org/terms/rel/revoke relation - Abandon a loan or hold](https://github.com/NYPL-Simplified/Simplified/wiki/OPDS-For-Library-Patrons#the-httplibrarysimplifiedorgtermsrelrevoke-relation---abandon-a-loan-or-hold)
        - 
# Dependent Lyrasis Issues

- [PR-1548](https://github.com/NYPL-Simplified/circulation/pull/1548)
- [MR-3354](https://jira.nypl.org/browse/SIMPLY-3354)
- [MR-3355 ](https://jira.nypl.org/browse/SIMPLY-3355)
- [MR-3356](https://jira.nypl.org/browse/SIMPLY-3356)
- [MR-3359](https://jira.nypl.org/browse/SIMPLY-3359)


danielweck:
Hello @saad-kh may I please invite you to change your PR target to the `develop` branch, and to merge the `develop` branch into your working branch (there have been a few significant updates, not least @panaC 's "auth" now-merged PR)
Thanks :)

danielweck:
For reference:

Old Branch:
https://github.com/edrlab/thorium-reader/tree/panac/feature/opds-authentication

Merged PR:
https://github.com/edrlab/thorium-reader/pull/1252

saad-kh:
Hi @danielweck, @panaC 

I've rebased the code to target develop and cleaned the commit history.

I have a question regarding this bit [rel check in handling link](https://github.com/edrlab/thorium-reader/blob/6275ee2f1af5a1a7f88b98b4019c9f433797995e/src/renderer/library/opds/handleLink.ts#L30)

Because we need to use revoke link within the thorium browsing context, I added this condition to not go to an external browser. But it's hard coded for now. Do you have ideas about how to streamline this?

danielweck:
Great Saad, thank you. I have a quick look at the code diff, and saw the ad-hoc `rel` check. On initial impressions, this looks fine to me.

saad-kh:
Good then it makes things simpler.

I have another issue to discuss. 

We use our access token on requests only when we get a 401, but for some requests (like the alternate link of an acquired link), when trying to access them without an access token, they will return the unprotected content instead of the protected content relevant to us.

We can either enable the usage of the access token:
* Whenever possible.
* Use it when we used it before in the session, and thus bringing some session awareness to the reader.
* Detect the requests that might need it depending on the context (alternate link when browsing a protected link)
* Or something else.

What do you think, @danielweck @panaC

danielweck:
My take on the HTTP `401` issue: I think that strictly-speaking Thorium implements a correct / logical auth flow ... but it is clear that in some cases (now, and in the future) we should be able to discover API endpoints / links that require auth data in the HTTP request headers. For example, the OPDS specification folks are discussing such link "relationship" (`rel`), I believe, to avoid the unnecessary and costly `401` hops.
So, we might as well implement a generic strategy for this mechanism, and populate it with concrete use cases such as the ones found in Saad's SAML work. Then we will also implement whatever the OPDS auth spec. comes up with.

panaC:
I do not really know.
We tackle the authentication only after a 401 http return code. In the case or the accessToken is not needed for the server. There are many server implementation with their own logic, sadly.
Maybe we can hard code the behavior in function of whitelist domains, but not the best, Obviously..

Daniel necessarily knows more than me

panaC:
saml server doesn't provide a cookie ? because in that case it works

saad-kh:
Thanks for your answers @panaC  and @danielweck.

It does not seem that they rely on cookies for tracking sessions, and I think that such thing will always be implementation dependent.

I have this simple implementation that use an accessToken first access [Access Token First Request](https://github.com/edrlab/thorium-reader/pull/1272/commits/3e2f90e5bbfd5aeb7ff1f7ddaadb84b8977d9e8a)

What do you think?

saad-kh:
I Marked the PR as Ready to Review as it's essentially working reasonably well against the up to date Circulation Manager (which is not yet deployed, but tested locally)

danielweck:
Hello Pierre, would you mind doing a code review as the changes impact on / integrate with your original design and implementation:
https://github.com/edrlab/thorium-reader/pull/1272/files
Thank you!
(I will do a more generic peer review)

danielweck:
Hello @panaC Pierre :)
Now that the PDF PR is finished (at peer review stage), could you please do a review of Saad's PR here? Thank you.

danielweck:
For our records, external issues that impact this work:
* https://jira.nypl.org/browse/SIMPLY-3354
* https://jira.nypl.org/browse/SIMPLY-3355
* https://jira.nypl.org/browse/SIMPLY-3356
* https://jira.nypl.org/browse/SIMPLY-3359

danielweck:
For reference: http://www.lirtuel.be/api/saml
(vs. http://www.lirtuel.be/api/ezproxy )

src/main/network/http.ts
```diff
@@ -294,35 +294,27 @@ export const httpGetWithAuth =
...
             }
-            return response;
+
+            return httpFetchFormattedResponse(
+                _url,
+                options,
+                enableAuth ? undefined : _callback,
+                ..._arg,
+            );
+
```
danielweck, line 317: This certainly needs reviewing carefully. This is taking us back to the original crude prototype which populated auth headers in HTTP requests before receiving any HTTP 401 responses.

danielweck: @saad-kh is there a link `rel` we can leverage (?) to discover that a given URL is an endpoint that necessitates auth, and which does not produce a 401 response in case of missing auth data.

panaC: I suggest a call asap

saad-kh: Hi Daniel, Pierre,

Here's the kind of link where we need to send an authentication token without waiting for a 401 response.

`<link href="" type="application/atom+xml;type=entry;profile=opds-catalog" rel="alternate">`

danielweck: Ah, so there is nothing "special" in `type` or `rel` that would allow us to reliably implement a workaround ... damn!

danielweck:
Thank you Saad, I have squashed and merged your PR into a staging branch where I will be fixing the localization JSONs, in preparation for the final merge into the `develop` branch.

danielweck:
Follow-up PR https://github.com/edrlab/thorium-reader/pull/1366

PR #1252, opened 2020-11-04
Oauth2 and basic opds2 authentication

panaC:

PR #1275, opened 2020-12-03
fixes #1148 - filter opds1 entryLinks on rel='alternate' - better contentType parsing

### LINKED ISSUE #1148: Thorium Reader is not able to borrow books from an OPDS feed
I faced with an issue when Thorium Reader is not able to borrow books from an OPDS feed.

The OPDS feed I'm using: https://github.com/vbessonov/opds-feed

Steps to reproduce:
1. git clone https://github.com/vbessonov/opds-feed
2. cd opds-feed
3. docker-compose up -d
4. Open Thorium Reader and add a new OPDS feed using `http://localhost:8080` as its URL:
image
5. Open `Test OPDS feed` catalog and try to borrow a book:
image
6. Thorium Reader opens a `borrow` link in a browser and then stops:
image

The books wasn't downloaded and isn't accessible in Thorium Reader:
image

panaC:
- filter alternate entrylinks in opds1

'alternate' rel is it sufficient ?

fixes #1148

### DIFF
--- a/src/utils/content-type.ts
+++ b/src/utils/content-type.ts
@@ -33,3 +33,16 @@ export enum ContentType {
     lcppdf = "application/pdf+lcp",
     pdf = "application/pdf",
 }
+
+export const parseContentType = (RawContentType: string): ContentType | undefined => {
+
+    if (!RawContentType) {
+        return undefined;
+    }
+
+    const contentTypeArray = RawContentType.replace(/\s/g, "").split(";");
+
+    const contentType = contentTypeArray.reduce<ContentType | undefined>(
+        (pv, cv) => pv || Object.values(ContentType).find((v) => v === cv), undefined);
+    return contentType;
+};
--- a/src/main/converter/opds.ts
+++ b/src/main/converter/opds.ts
@@ -320,6 +320,7 @@ export class OpdsFeedViewConverter {
         const entrylinkView = fallback(
             this.convertFilterLinkToView(baseUrl, r2OpdsPublication.Links, {
                 type: "type=entry;profile=opds-catalog",
+                rel: "alternate",
             }),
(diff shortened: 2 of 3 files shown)


panaC:
@danielweck can you merge it ?

src/utils/content-type.ts
```diff
@@ -33,3 +33,16 @@ export enum ContentType {
     lcppdf = "application/pdf+lcp",
     pdf = "application/pdf",
 }
+
+export const parseContentType = (RawContentType: string): ContentType | undefined => {
```
danielweck, line 37: In Thorium, handling / comparison of "content type" values is very inconsistent. There is now this new function `parseContentType()`, but elsewhere in the codebase there is `===`, `array.includes()`, `string.startsWith()`.

I did a quick search for `ContentType.xxx` in the TypeScript source tree to find all usages:

In addition to this PR's `dispatchOpdsLink()` function, there are the `findLink()` and `filterTypeLink()` functions, there is also the code inside `opdsRequest()` (`isPub` `isAuth` and `isFeed`), and in `OpdsService` there is `contentTypeisXml()` + `contentTypeisOpds()`, and the comparison code in `returnPublication()` (`renewLink.Type`), and `importFromLinkService()` with `isLcpFile`, `isEpubFile`, `isAudioBookPacked`, `isAudioBookPackedLcp`, `isHtml`, `isDivinaPacked`, `isPdf`, `isJson`.

So, I do not mind merging this PR, but isn't this an opportunity to harmonize the source code, to remove the inconsistencies?

What do you think?

panaC: Maybe see #1252 : it's extract from it

danielweck: I see. Let's merge this PR then, and the WIP PR will address inconsistencies.

panaC: 👍

PR #1287, opened 2020-12-23
Adding Russian localization

Modeo147:


panaC:
each line of your `src/resources/locales/ru.json` file end with a carriage return and line feed ascii code, but it should ended with line feed only.
May you set your code editor to format your file with only a line feed ?

for example : https://stackoverflow.com/questions/10034585/why-does-n-give-a-new-line-on-windows#:~:text='%5Cn'%20is%20the%20Line%20Feed%20character.&text='%5Cr'%20is%20the%20Carriage,left%20edge%20of%20the%20paper.

panaC:
To complete your translation. You can provide the Russian countries localisation (ISO 3166-1) like this : https://github.com/edrlab/thorium-reader/blob/develop/src/resources/locale-names/enLang.json.

potential resources : 

- https://ru.qaz.wiki/wiki/List_of_ISO_3166_country_codes

- https://github.com/edrlab/thorium-reader/blob/638b62e5fe94965bb413c18554e27ffcd4551973/scripts/readme.md
- https://github.com/edrlab/thorium-reader/blob/638b62e5fe94965bb413c18554e27ffcd4551973/scripts/csvToJson.py

danielweck:
Thank you for your contribution! I have merged your work into an intermediary branch so I can add a few missing integration / plumbing bits.

danielweck:
See: https://github.com/edrlab/thorium-reader/pull/1307

PR #1291, opened 2020-12-31
enhance catalog update from setReduxState

panaC:


danielweck:
this looks great, but could you please give a description of the bug that this fixes? i am wondering whether this fixes a library (mis)behaviour that i observed some time ago, but wasn’t able to reproduce. Thanks :)

panaC:
not a bug fixes in fact, just an evolution

danielweck (approved):

PR #1292, opened 2021-01-04
fix #1273 home screen bookshelf

### LINKED ISSUE #1273: Home screen evolution
In the home screen, we'll find, from top to bottom: 

"Recently added" (the max 5 publications of all types added since less than a month and not yet opened)
"Resume reading EPUB" (max 5 last open publication)
"Resume reading PDF"
"Resume reading Audiobook"
"Resume reading Divina"

and of course the section is absent if empty.

An "All books" link is added at the right of each section, to remind the user that he can go to the page listing all his books. This will later be refined when a selection by type (EPUB, PDF ...) is available in the user catalog screen.

panaC:
fixes #1273


src/main/api/catalog.ts
```diff
@@ -71,64 +71,63 @@ export class CatalogApi implements ICatalogApi {
...
-            {
-                title: __("catalog.entry.continueReading"),
-                totalCount: epubReaded.length,
-                publicationViews: epubReaded,
-            },
             {
                 title: __("catalog.entry.lastAdditions"),
-                totalCount: epubAdded.length,
-                publicationViews: epubAdded,
+                totalCount: allAdded.slice(0, NB_PUB).length,
```
danielweck, line 93: Shouldn't this be `Math.max(NB_PUB, allAdded.length)`?
(same remark with all the other cases in this PR)

danielweck: Sorry, brain fart :(

src/main/api/catalog.ts
```diff
@@ -138,104 +137,215 @@ export class CatalogApi implements ICatalogApi {
...
 
-        const lastAddedPublicationsDocumentArray = lastAddedPublicationsDocument.map(
-            (doc) => this.getLastReadingPublicationId().includes(doc.identifier)
-                ? [doc, undefined]
-                : [undefined, doc],
-        );
+        const pdf = {
+            readed: lastReadedPublicationsView.filter(
+                (view: PublicationView) => {
+                    const r2Publication = viewToR2Pub(view);
```
danielweck, line 217: Pierre, as you remarked before, each call to the `viewToR2Pub()` function is computationally expensive. As it is only necessary for `isPdfFn()`, is there a way we could cache some kind of `isPDF` boolean?

src/main/api/catalog.ts
```diff
@@ -71,64 +71,63 @@ export class CatalogApi implements ICatalogApi {
...
-                title: __("catalog.entry.continueReading"),
-                totalCount: epubReaded.length,
-                publicationViews: epubReaded,
-            },
             {
                 title: __("catalog.entry.lastAdditions"),
-                totalCount: epubAdded.length,
-                publicationViews: epubAdded,
+                totalCount: allAdded.slice(0, NB_PUB).length,
+                publicationViews: allAdded.slice(0, NB_PUB),
```
danielweck, line 94: `allAdded.slice(0, NB_PUB)` has a cost (array allocation + shallow copy, i.e. clone of primitive types, or pointer copies for structured objects).
Therefore, I would advice not invoking `slice()` for both `totalCount` and `publicationViews`.

PS: same remark for all other cases in this PR.

Technical references:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice

Note the difference in NodeJS between `TypedArray#slice()`, `Buffer#slice()` and `TypedArray#subarray()`:

https://nodejs.org/api/buffer.html#buffer_buffers_and_typedarrays

> While TypedArray#slice() creates a copy of part of the TypedArray, Buffer#slice() creates a view over the existing Buffer without copying. This behavior can be surprising, and only exists for legacy compatibility. TypedArray#subarray() can be used to achieve the behavior of Buffer#slice() on both Buffers and other TypedArrays.

danielweck (changes requested):
Laurent's design brief states `5` max publications per horizontal scroll, not `10`.

https://github.com/edrlab/thorium-reader/issues/1273

https://github.com/edrlab/thorium-reader/blob/e1da810b7db13b83a012c8abf6c57b9c7234060c/src/main/api/catalog.ts#L31

src/main/api/catalog.ts
panaC: you're right , need to memoize ..
perhaps. this: https://github.com/ramda/ramda/tree/v0.27.0/source/memoizeWith.js

src/main/api/catalog.ts
panaC: thanks for the resources 👍

danielweck:
> not actually implemented. I suggest to implement this new feature in a new couple of issue/pr. What do you think ?

Pierre, I hope you don't mind, I decided to add this feature in your PR / branch:

https://github.com/edrlab/thorium-reader/pull/1292/commits/11ab9a723f6e276e2e17eb0c2b4c9983a44c8c31

Let me know what you think, and feel free to revert this code change if you disagree (in which case we will create a separate PR to fully meet Laurent's design brief)

panaC:
> > not actually implemented. I suggest to implement this new feature in a new couple of issue/pr. What do you think ?
> 
> Pierre, I hope you don't mind, I decided to add this feature in your PR / branch:
> 
> [11ab9a7](https://github.com/edrlab/thorium-reader/commit/11ab9a723f6e276e2e17eb0c2b4c9983a44c8c31)
> 
> Let me know what you think, and feel free to revert this code change if you disagree (in which case we will create a separate PR to fully meet Laurent's design brief)

No !, I'm very glad

src/main/api/catalog.ts
panaC: https://github.com/edrlab/thorium-reader/pull/1292/commits/6c4b1e1b28eb72c819c081f96e7e9d31ecaf13f1

danielweck: Good idea. `R.identity` for `view` will not work, right? (because the object pointer is different at each invocation, I think)

danielweck: Ah, I have just seen your commit with `PublicationView.identifier` ... great :)

src/main/api/catalog.ts
```diff
@@ -30,20 +31,28 @@ export const CATALOG_CONFIG_ID = "catalog";
 
 const NB_PUB = 10;
 
+const viewToR2Pub = memoizeWith((v) => v.identifier, (view: PublicationView) => {
```
danielweck, line 34: I like the idea of using Ramda's `memoizeWith()`, however this creates an in-memory cache of the entire `R2Publication` object for nearly all the publications in the catalog. Instead, couldn't we just memoize a `isPDF` boolean, as this is all we need `viewToR2Pub()` / `R2Publication` for.

danielweck (changes requested):

src/main/api/catalog.ts
panaC: i'm wonder if we can use this for the whole app ?

danielweck: Well, let's see ... the memoization is really just a lazy-loaded memory map between string identifier and R2Publication instances, but over hundreds/thousands of publications in the catalog this will eventually create significant additional memory allocations / footprint.

src/main/api/catalog.ts
danielweck: Fixed via https://github.com/edrlab/thorium-reader/pull/1292/commits/f3038a774bd3f76383d805e38342fd6350fd2fd7

src/main/api/catalog.ts
danielweck: Addressed via https://github.com/edrlab/thorium-reader/pull/1292/commits/697b7977c22a98da66fb3168907404953546db51

PR #1297, opened 2021-01-07
[FIX 1296] fix the redux persistence by atomic file writing rather than pouchdb

### LINKED ISSUE #1296: save the main redux state in disk rather than in pouchdb
Currently the redux state from main process is saved with a debounce every second into the pouchdb config repository.
It's probably better to load and save it from the file system rather than from the orm.

panaC:
- increase the persistence debounce time from 1s to 30s.

fix #1296

### DIFF
--- a/src/main/di.ts
+++ b/src/main/di.ts
@@ -7,6 +7,7 @@
 
 import "reflect-metadata";
 
+import * as atomically from "atomically";
 import * as debug_ from "debug";
 import { app, BrowserWindow } from "electron";
 import * as fs from "fs";
@@ -287,6 +288,48 @@ const getAllReaderWindowFromDi =
     () =>
         container.getAll<BrowserWindow>("WIN_REGISTRY_READER");
 
+/// redux persistence
+
+const rootReduxPersist = path.join(
+    userDataPath,
+    (_NODE_ENV === "development" || _CONTINUOUS_INTEGRATION_DEPLOY) ? "reduxpersist-dev.json" : "reduxpersist.json",
+);
+
+const getReduxPersistState = async (): Promise<Partial<RootState>> => {
+
+    try {
+        const buffer = await atomically.readFile(rootReduxPersist);
+        return JSON.parse(buffer.toString());
+    } catch (e) {
+
+        debug(e);
+        debug("create a new file:", rootReduxPersist);
+        fs.writeFileSync(rootReduxPersist, Buffer.from(JSON.stringify({})));
+    }
+    return {};
+};
+
+const setReduxPersistState = async (data: Partial<RootState>) => {
+
+    const buffer = Buffer.from(JSON.stringify(data));
+
+    try {
+        await atomically.writeFile(rootReduxPersist, buffer);
+    } catch (error) {
+
+        debug("######");
+        debug("######");
+        debug("ERROR TO ATOMICALLY Write ", rootReduxPersist);
+        debug(error);
+        debug("######");
+        debug("######");
+
+        fs.writeFileSync(rootReduxPersist, buffer);
+    }
+};
+
+///
+
 // local interface to force type return
 interface IGet {
     (s: "store"): Store<RootState>;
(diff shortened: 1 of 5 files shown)


src/main/redux/sagas/persist.ts
```diff
@@ -22,20 +22,22 @@ debug("_");
...
+        publication: nextState.publication,
+        reader: nextState.reader,
+        session: nextState.session,
+    };
 
     debug("start of persist reduxState in disk");
+    await setReduxPersistState(data);
+
+    // TODO : REMOVE THIS !
+    // there are a new redux persistence
```
danielweck, line 36: Can you please add "WIP" to the PR title, and make sure the PR is marked as "Draft". Thanks!

src/main/redux/sagas/persist.ts
```diff
@@ -8,12 +8,12 @@
...
 import { selectTyped } from "readium-desktop/common/redux/sagas/typed-saga";
 import { ConfigRepository } from "readium-desktop/main/db/repository/config";
-import { CONFIGREPOSITORY_REDUX_PERSISTENCE, diMainGet } from "readium-desktop/main/di";
+import { CONFIGREPOSITORY_REDUX_PERSISTENCE, diMainGet, setReduxPersistState } from "readium-desktop/main/di";
 import { winActions } from "readium-desktop/main/redux/actions";
 import { RootState } from "readium-desktop/main/redux/states";
 import { call, debounce  } from "redux-saga/effects";
 
-const DEBOUNCE_TIME = 1000;
+const DEBOUNCE_TIME = 30000;
```
danielweck, line 16: On first impressions, this seems like a very long time (for example, AudioBook playback notifications are throttled at 1s intervals, so there would be many missed reading location updates).
However, pass the first impressions, I am wondering whether the state is immediately saved to persistent storage when closing the app? To me, this is the main scenario that needs to be covered: user reads / listens to (audio)book, quickly close the app => we must make sure all current settings are preserved, especially the current reading location which accounts for the highest notification / event traffic / relatively high frequency ... which is why the database "compact" MUST work reliably (currently does not seem to :(

src/main/redux/sagas/persist.ts
panaC: It is not a WIP PR , but also not a PR to merge now.

src/main/redux/sagas/persist.ts
panaC: indeed the state is saved to persistent storage on the closing process

src/main/di.ts
```diff
@@ -287,6 +288,48 @@ const getAllReaderWindowFromDi =
...
+
+        debug("######");
+        debug("######");
+        debug("ERROR TO ATOMICALLY Write ", rootReduxPersist);
+        debug(error);
+        debug("######");
+        debug("######");
+
+        fs.writeFileSync(rootReduxPersist, buffer);
+    }
```
danielweck, line 328: May I suggest adding a comment about `EXDEV`, as per:

https://github.com/sindresorhus/conf/blob/a96e9d78ac7e675a24ed702f1c21dd46a0b3ae0e/source/index.ts#L456-L468

src/main/redux/sagas/persist.ts
danielweck: So I guess this leaves the possibility of a sudden application crash, in which case the worst case scenario is that 30s worth of state changes can would be missed.

src/main/redux/sagas/persist.ts
danielweck: > It is not a WIP PR

WIP literally means "Work In Progress", and this PR is included in the `v1.6` project with the "In Progress" status ... so this is inconsistent.

> but also not a PR to merge now

In this case, it is definitely a WIP, and should be marked as "Draft".

danielweck: PS: the `TODO : REMOVE THIS !` code comment is what prompted me to remark on the status of the PR (I assume dual DB / FS persistence is not the final logic, right?)

PR #1300, opened 2021-01-08
fix #1299

### LINKED ISSUE #1299: protocol.registerSchemesAsPrivileged should be called before app is ready at initSessions
`protocol.registerSchemesAsPrivileged should be called before app is ready at initSessions`

error happen on a second instance launch and the current instance get the lock

panaC:
fix #1299

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -61,12 +61,12 @@ setLcpNativePluginPath(lcpNativePluginPath);
 //     process.exit();
 // });
 
+// protocol.registerSchemesAsPrivileged should be called before app is ready at initSessions
+initSessions();
+
 export const main = async (flushSession: boolean = false): Promise<Store<RootState>> => {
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -61,13 +61,13 @@ setLcpNativePluginPath(lcpNativePluginPath);
 //     process.exit();
 // });
 
+// protocol.registerSchemesAsPrivileged should be called before app is ready at initSessions
+initSessions();
+
 export const main = async (flushSession: boolean = false): Promise<Store<RootState>> => {
```
danielweck, line 67: Ah yes, indeed! (race condition)

```
                    await Promise.all([
                        main(xxx),
                        app.whenReady(),
                    ]);
```

danielweck (approved):

PR #1308, opened 2021-01-08
[WIP] auto fs backup in db repo base class

panaC:


src/main/db/repository/base.ts
```diff
@@ -80,6 +93,17 @@ export abstract class BaseRepository<D extends Identifiable & Timestampable> {
 
         await this.db.put(dbDoc);
 
+        try {
+            await atomically.writeFile(path.join(dbBackupfilePath, dbDoc._id), JSON.stringify(dbDoc));
+        } catch (e) {
```
danielweck, line 98: May I suggest adding a comment about `EXDEV`, as per:

https://github.com/sindresorhus/conf/blob/a96e9d78ac7e675a24ed702f1c21dd46a0b3ae0e/source/index.ts#L456-L468

danielweck:
For reference: I think this PR is successor or https://github.com/edrlab/thorium-reader/pull/1297

panaC:
What do you think @danielweck ?

panaC:
I think we have to add this to next release. the sooner the better. For the next db migration.

If you accepted this PR Daniel, of course :)

danielweck (changes requested):
I've just fixed the unit tests (minor change, ad-hoc dependency injection container configuration).

I've just checked the generated JSON files, and of course, as we know, the "secrets" are stored in plain text. We need to ensure that LCP passphrases (SHA256) and OPDS authentication tokens + cookie jar are stored more securely.

Issue #1328, opened 2021-01-19
"about Thorium" auto-generated EPUB fails to create on Windows


sukiletxe:
v1.6.0-alpha.1.495798844
Steps:
1. Open Thorium.
2. Press on "More information", just after version number.

Expected: The information should show.
Actual: 3 error dialogs show.

1.

Uncaught Exception:
Error: Invalid package c\Users\sukil\AppData\Local\Programs\EDRLab.ThoriumReader\resources\app.asar
    at createError (electron/js2c/asar.js:126:17)
    at Object.fs.lstat (electron/js2c/asar.js:257:23)
    at electron/js2c/asar.js:294:33
    at processTicksAndRejections (internal/process/task_queues.js:76:11)

2.

Uncaught Exception:
Error: Invalid package c\Users\sukil\AppData\Local\Programs\EDRLab.ThoriumReader\resources\app.asar
    at createError (electron/js2c/asar.js:126:17)
    at Object.fs.lstat (electron/js2c/asar.js:257:23)
    at electron/js2c/asar.js:294:33
    at processTicksAndRejections (internal/process/task_queues.js:76:11)

3.

Uncaught Exception:
Error: Invalid package c\Users\sukil\AppData\Local\Programs\EDRLab.ThoriumReader\resources\app.asar
    at createError (electron/js2c/asar.js:126:17)
    at Object.fs.lstat (electron/js2c/asar.js:257:23)
    at electron/js2c/asar.js:294:33
    at processTicksAndRejections (internal/process/task_queues.js:76:11)

danielweck:
I am observing this error in dev mode:
<img width="590" alt="Capture" src="https://user-images.githubusercontent.com/625201/105030562-e1aa1800-5a4b-11eb-941f-154e4e4084a7.PNG">

danielweck:
Follow-up to https://github.com/edrlab/thorium-reader/issues/1325

danielweck:
I am not able to reproduce the error in MacOS, I suspect Linux is fine too. @panaC could you please check at your end? (make sure to delete Thorium's config folder so that the "about" EPUB is re-generated)

danielweck:
Note the double / duplicate file path concatenation in my screenshot. This is probably just a misuse of `path.join()` or some other filesystem APi in Windows.

danielweck:
@panaC have you been able to reproduce this bug on your Windows machine?

panaC:
i haven't yet any windows machine. I will see next week

[...]

PR #1344, opened 2021-01-28
Update Japanese translation (1.6 version) 

peaceroad:
Hello,
It seems that version 1.6 will be released soon, so I updated the Japanese translation. It also includes a few things that made the previous translation better.

Also, there was en.xhtml in /src/resources/information, so I created ja.xhtml.


peaceroad:
I noticed that "more information" of About Throuim cannot be opened on Windows in the current develop, but the issue was already mentioned in #1328.

danielweck:
Great, thank you very much!

PR #1353, opened 2021-01-31
Update spanish translation

nvdaes:
Please, let me know if the keyboard shortcuts should be also updated in the spanish wiki.


danielweck:
Thank you so much, this is great!

nvdaes:
Thank you for merging.

Enviado desde mi iPhone

PR #1354, opened 2021-02-01
Panac/fix/lcp start date in future

### LINKED ISSUE #1351: LCP, bad message if the start date is in the future
If one creates a license with a start date in the future and side-loads it in Thorium, the message which appears is about an "expired" / "out of date" licence. 

The message should be "License not yet valid" / "Licence par encore valide" (fr).

panaC:
fixes #1351

### DIFF
--- a/src/main/redux/sagas/api/publication/import/importLcplFromFs.ts
+++ b/src/main/redux/sagas/api/publication/import/importLcplFromFs.ts
@@ -61,7 +61,7 @@ export function* importLcplFromFS(
             try {
                 if (r2LCP.Rights.Start) {
                     if (moment(r2LCP.Rights.Start).isAfter(now)) {
-                        res = 11;
+                        res = 1111;
                     }
                 }
                 if (r2LCP.Rights.End) {
--- a/src/main/services/lcp.ts
+++ b/src/main/services/lcp.ts
@@ -566,6 +566,10 @@ export class LcpManager {
                     message = this.translator.translate("publication.licenseOutOfDate");
                     break;
                 }
+                case 1111: {
+                    message = this.translator.translate("publication.licenseNotYetValid");
+                    break;
+                }
                 case 101: {
                     // message = "CERTIFICATE_REVOKED: " + val;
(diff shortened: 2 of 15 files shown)


src/main/redux/sagas/api/publication/import/importLcplFromFs.ts
```diff
@@ -61,7 +61,7 @@ export function* importLcplFromFS(
             try {
                 if (r2LCP.Rights.Start) {
                     if (moment(r2LCP.Rights.Start).isAfter(now)) {
-                        res = 11;
+                        res = 1111;
```
danielweck, line 64: That's fine, but please add a code comment:
```
// Normally: LICENSE_OUT_OF_DATE 11
// Note that this is reserved already: LICENSE_SIGNATURE_DATE_INVALID 111
```

danielweck: I wonder whether we need similar logic here:
https://github.com/edrlab/thorium-reader/blob/cf04170d65959b66cef906f3cbe362d8dbbd73c0/src/main/redux/sagas/publication/openPublication.ts#L102-L104
and:
https://github.com/edrlab/thorium-reader/blob/cf04170d65959b66cef906f3cbe362d8dbbd73c0/src/main/redux/sagas/publication/openPublication.ts#L188-L190
and:
https://github.com/edrlab/thorium-reader/blob/cf04170d65959b66cef906f3cbe362d8dbbd73c0/src/main/api/lcp.ts#L66-L68

danielweck (changes requested):

danielweck:
After Slack discussion, a sufficient fix is to change the French translation of `“licenseOutOfDate”: “License is out of date”` which is currently `License expirée` (which doesn't work for LCP licenses with start date in the future)

danielweck:
Proposal for the French translation: "License inutilisable à cette date" (English: "license out of date")

danielweck:
Based on Laurent's input in the Slack discussion, I propose the following code change: https://github.com/edrlab/thorium-reader/pull/1357

PR #1359, opened 2021-02-03
increase debounce time to persist state

panaC:


danielweck:
3 minutes interval is fine.
For visitors wondering if this means that 3 minutes worth of state updates will be lost when leaving the app (for example, the precise playback position in an audiobook): no, that's not problematic because the state is saved on shutdown:

https://github.com/edrlab/thorium-reader/blob/4c7ec27f44a36b2ab1b2255a1e5289ff09fa3860/src/main/redux/sagas/app.ts#L102-L132

...however this can be problematic when the application crashes unexpectedly.

danielweck (approved):

PR #1360, opened 2021-02-03
opds group self link

### LINKED ISSUE #1295: OPDS UI, Mixed Feeds
Mixed feeds contain navigation feeds + publication feeds, with groups. Only OPDS 2 supports Mixed Feeds. 

The navigation feed and publication feed parts are detailed in #1293 and #1294. We'll focus here on groups. 

The only evolution is that:
- group labels become hyperlinks, using the self link found in the OPDS 2 structure. 
- the number of items found in the structure is displayed on the right of the group label, as it appears in navigation feeds (but with no caret after the count).

panaC:
fixes #1295

### DIFF
--- a/src/main/converter/opds.ts
+++ b/src/main/converter/opds.ts
@@ -158,7 +158,7 @@ export class OpdsFeedViewConverter {
 
         return (subject.Name || subject.Code) ? {
             name: convertMultiLangStringToString(subject.Name || subject.Code),
-            link: this.convertFilterLinkToView(baseUrl, subject.Links || [], {
+            link: this.convertFilterLinksToView(baseUrl, subject.Links || [], {
                 type: [
                     ContentType.AtomXml,
                     ContentType.Opds2,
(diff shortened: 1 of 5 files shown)


src/main/converter/opds.ts
```diff
@@ -35,6 +35,7 @@ import { filterRelLink, filterTypeLink } from "./tools/filterLink";
 import { urlPathResolve } from "./tools/resolveUrl";
 import { TLinkMayBeOpds, TProperties } from "./type/link.type";
 import { ILinkFilter } from "./type/linkFilter.interface";
+import { OPDSProperties } from "r2-opds-js/dist/es6-es2015/src/opds/opds2/opds2-properties";
```
danielweck, line 38: `import { OPDSProperties } from "@r2-opds-js/opds/opds2/opds2-properties";`

;)

danielweck (approved):

PR #1363, opened 2021-02-04
facets + pub view grid

### LINKED ISSUE #1294: OPDS UI, Publication / acquisition feeds
Pure publication (opds 2 terminology) / acquisition (opds 1 terminology) feeds contain publications only.  They may present facets. 

If facets are present, they are represented as a side-panel (like those of the reading view). The panel appears on the right of the screen.  
To make the side panel appear, a "Facets" label is added on the right of the screen, aligned with the breadcrumb, the font and font size are the same as in the breadcrumb. If the screen size is no large enough to have both the  "Facets" and breadcrumbs, the end of the breadcrumbs are hidden. 

The mosaic presentation of the books does not change,  but the list view is not present in publication feeds (the icon is hidden), only "cards" are available. The reason for that is that when a feed is mixed (navigation blocks  + publication groups interlaced) the list view is defective). 

Facets are split in groups. Each group has a title and a set of values with a label, a count and an hyperlink.
The title of a facet group is in bold; values are listed below the group title and are clickable, counts are in narrow chars. 
No caret (this is different from navigation feeds). 
Groups are separated by a 20px space.

panaC:
fixes #1294


src/renderer/library/components/opds/BrowserResult.tsx
```diff
@@ -73,22 +73,34 @@ export class BrowserResult extends React.Component<IProps, undefined> {
...
-                            }
-                            <EntryPublicationList
-                                opdsPublicationView={browserResult.data.publications}
-                                links={browserResult.data.links}
-                                pageInfo={browserResult.data.metadata}
-                            />
+                            <div className={Array.isArray(facetsRender) ? styles.publicationgrid : ""}>
+                                {
+                                    Array.isArray(facetsRender)
+                                        ? <div className={styles.publicationgriditem}>
```
danielweck, line 89: The CSS class `publicationgriditem` is missing from the style definition, is that intentional?
(see `src/renderer/assets/styles/opds.css` where only `publicationgrid` is defined)

danielweck (changes requested):

src/renderer/library/components/opds/BrowserResult.tsx
panaC: anticipation :)

danielweck: Okay, so I will test and merge as-is.

danielweck (approved):

PR #1362, opened 2021-02-04
Finnish translation

terhol-celia:

PR #1364, opened 2021-02-05
WIP Finnish translation

danielweck:
Original PR: #1362


danielweck:
Hello @terhol-celia I finished correcting your Finnish locale PR. Thank you very much!

For you information:

* The JSON was broken in one place (invalid syntax), and the UTF8 Unicode binary wasn't right either.
* The XHTML was detected as ISO-8859-1 but was treated as UTF-8 encoding, which was breaking the text editor as well as the HTML webview renderer (I fixed this as well).

What text editor and operating system did you use to create the XHTML and JSON files?

danielweck:
That's what the XHTML looked like, before the encoding fix:

Screenshot 2021-02-05 at 17 51 49

PR #1370, opened 2021-02-09
misc. code fixes for URL resolve and handling of Windows filepaths

danielweck:
Apart from the percent-encoding URL potential issue, this PR does not fix any obvious user-level bugs. This is mostly a developer code cleanup to make sure we don't get bugs in future code revisions.

* Fixed `.replace("\\", "/")` (replace once) with the correct regexp `.replace(/\\/g, "/")` (replace all)
* Fixed `importFromStringService()` which was setting `baseUrl` to `file://` (missing `baseFileUrl`)
* Fixed string-based "manual" concatenation of `file://` with Windows filepath: backslash produces incorrect URL syntax (by luck, goes undetected because following uses do not trigger URL parsing bugs, as they rely on filesystem API such as `path.join()` which will happily consume the characters that follow `file://` without sanity checks (whereas URL parsing would not correctly handle the backslashes as separators in normative URL syntax). 
* Replaced occurences of `url.resolve(from, to)` (deprecated API https://nodejs.org/api/url.html#url_url_resolve_from_to .. with inconsistent / incorrect handling of `file://` with Windows drives (e.g. `C:/` instead of `/C:/` in the `URL` API ... I added a comment to clarify this slash prefix edge-case)
* Removed unjustified use of `decodeURIComponent()` which can break URL resolving (thrown exception), if the percent character is present without actually being followed by an escape entity reference (e.g. `audio_100%_progress.mp3`). The `encodeURIComponent()` counterpart is typically used to ensure that URL/URI query parameters are escaped appropriately, e.g.: `https://domain.com/path?q=URI_ENCODED_PARAM` where URI_ENCODED_PARAM is `encodeURIComponent("https://domain.org/here?q1=v1&q2=v2")`. I noticed several uses of `decodeURIComponent()` in the codebase, for `link@href` values which shouldn't be escaped / encoded in the first place. Example: `node -e "console.log(decodeURIComponent('audio_100%25_.mp3'))"` => `audio_100%_.mp3` (fine). Other example: `node -e "console.log(decodeURIComponent('audio_100%_.mp3'))"` => `URIError: URI malformed at decodeURIComponent` (crash).
* added various code comments
* added several `debug()` statements at key points in the code flow, to track the packager process


danielweck:
This commit introduced a crash bug: https://github.com/edrlab/thorium-reader/commit/1c03bd7e52fdcbdcd70a2e911c9b49b26f1d30be

danielweck:
Peer code review and Windows + Mac tests finished. Thank you @panaC !

PR #1374, opened 2021-02-11
Replace TSLint with ESLint -- Fixes #625

### LINKED ISSUE #625: Migrate from TSLint to ESLint (for TypeScript) with Prettier rules
TSLint will be deprecated at some point, so our setup should probably migrate to ESLint with Prettier rules.
References:
https://github.com/typescript-eslint/typescript-eslint#what-about-tslint
https://github.com/prettier/eslint-config-prettier/blob/master/%40typescript-eslint.js
https://dev.to/robertcoopercode/using-eslint-and-prettier-in-a-typescript-project-53jb

danielweck:
Fixes https://github.com/edrlab/thorium-reader/issues/625


danielweck:
Hello @panaC I would like to merge this PR.

In my opinion, this is the minimum amount of changes required to integrate ESLint as a starting point.

I am pretty confident this doesn't break anything at runtime (the changes are trivial, mostly compile-time type fixes which works perfectly fine).

Note that I deactivated the entire Prettier code formatting checks, because there are too many indentation (etc.) errors to fix. The fix can easily be automated, but the resulting code diff would be so huge that it would be impossible to manually review, and this would introduce merge conflicts in all the other PRs, and the Thorium forks!

So, until maybe a major Thorium 2.0 release update, I think we have to live with a codebase full of formatting inconsistencies. That being said, we can incrementally clean up the mess, in isolated PRs which affect well-scoped filesets. It is simply a matter of removing `src` from the Prettier "ignore file".

panaC (approved):

PR #1389, opened 2021-02-22
[FIX #1384] reset search url after access to another feed domain

### LINKED ISSUE #1384: OPDS OpenSearch: when failure to discover templated search link (or timeout), previous search is still active!
(previous state is preserved, needs to be reset)

To reproduce: load an OPDS feed that has a valid OpenSearch URL (e.g. Feedbooks), search for a term, then load another OPDS feed that does not have a valid OpenSearch URL (e.g. desLibiris with BOM, see https://github.com/edrlab/thorium-reader/issues/1383 ) => the Feedbooks search is still present, giving the user the impression that they are searching on desLibris!!

panaC:
fix #1384


src/renderer/library/redux/sagas/opds.ts
```diff
@@ -28,6 +28,34 @@ export const SEARCH_TERM = "{searchTerms}";
...
+        const previousLocationPathname = yield* selectTyped(
+            (state: ILibraryRootState) => state.history);
+
+        const { rootFeedIdentifier: previousRootFeedIdentifier } =
+            parseOpdsBrowserRoute(previousLocationPathname[previousLocationPathname.length - 1]?.pathname) || {};
+
+
+        debug("RESET search url: ", previousRootFeedIdentifier, " != ", rootFeedIdentifier);
+
+        if (previousRootFeedIdentifier != rootFeedIdentifier) {
```
danielweck, line 44: Shouldn't the search URL be reset at **every** navigation? (just before a "search" URL is discovered (or not) in the feed) I don't think we should preserve the search URL if we are still on the same "domain" (OPDS root identifier). The "search" functionality is totally dependent on each OPDS feed declaring its available services / URL endpoints (hypermedia model).

danielweck (changes requested):

src/renderer/library/redux/sagas/opds.ts
danielweck: Please let me know what you think, maybe I missed something. Thank you!

PR #1426, opened 2021-03-13
fix #1424: dismiss cookies error

### LINKED ISSUE #1424: Error on the OPDS 2 fulcrum catalog
There is an error when adding to Thorium the OPDS 2.0 Catalog https://www.fulcrum.org/api/opds.

The message is: Cannot read property 'headers' of undefined.

This message is clear as mud. Before correcting the issue (in the fulcrum OPDS feed or in Thorium) we'll have to correct the message.

panaC:


llemeurfr:
by the way there are 4 instances of "tougth" in the fetch.ts file, to be renamed "tough".

PR #1435, opened 2021-04-09
fixes #1376 sanitize the opds description with dompurity

### LINKED ISSUE #1376: OPDS text/html description shows raw markup instead of rendered HTML
unnamed

panaC:
fixes #1376


danielweck (approved):
Tested with various different OPDS feeds.

PR #1437, opened 2021-04-13
feature (divina): new divina player

panaC:
from https://github.com/readium/divina-player-js/tree/panac/develop -> https://github.com/readium/divina-player-js/tree/panac/build


panaC:
NEED TO TEST

PR #1456, opened 2021-05-03
[WIP] json redux state recovery from file (user interface file picker / chooser)

panaC:


panaC:
before testing this PR need to merge this previous PR https://github.com/edrlab/thorium-reader/pull/1451

src/renderer/library/components/settings/RecoverySettings.tsx
```diff
@@ -0,0 +1,168 @@
...
+                </form>
+            </>
+        );
+    }
+
+    private getRecoveryState = async () => {
+
+        try {
+
+            const files = await fsp.readdir(this.configDataFolderPath);
```
danielweck, line 105: May I suggest moving this to the Electron main process, and then instead of `webContents.send("GETPATH", app.getPath("userData"));`, send the data parameter of `this.setState({ recovery });` via IPC.

danielweck (changes requested):

src/renderer/library/components/settings/Settings.tsx
```diff
@@ -40,6 +41,22 @@ class Settings extends React.Component<IProps, undefined> {
...
+                    <br></br>
+                    <br></br>
+                    <br></br>
+                    <br></br>
+                    <br></br>
+                    <br></br>
+                    <br></br>
+                    <br></br>
+
+                    <h1>Expert MODE</h1>
```
danielweck, line 58: TODO: translate

src/renderer/library/components/settings/RecoverySettings.tsx
```diff
@@ -0,0 +1,168 @@
...
+                            <input
+                                id={"radio-" + date}
+                                type="radio"
+                                name="language"
+                                onChange={() => this.setRecoveryFile(date)}
+                                checked={this.state.name === date}
+                            />
+                            <label htmlFor={"radio-" + date}>
+                                {this.state.name === date && <SVG svg={DoneIcon} ariaHidden />}
+                                {date ? new Date(parseInt(date, 10)).toString().split("GMT")[0] : "actual state"}
```
danielweck, line 92: I suspect that the `GMT` string split will break on computers with different locales. Better use `Moment` (already a dependency) or `Date-Fns` for processing timestamps.

src/renderer/library/components/settings/RecoverySettings.tsx
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+    };
+
+    private setRecoveryFile = async (name: string) => {
+
+        try {
+
+            const filename = name ? `state.${name}.json` : "";
+            await fsp.writeFile(this.recoveryFilePath, filename, { encoding: "utf8" });
```
danielweck, line 144: Suggestion: move to Electron main process, use IPC.

PR #1458, opened 2021-05-04
Update AboutThoriumButton.tsx

panaC:


danielweck:
Merging. Note caveat: old revisions are left behind (on filesystem and DB). But very small data cost, not a major concern at this point in time.

danielweck:
Fixes https://github.com/edrlab/thorium-reader/issues/1447

PR #1476, opened 2021-05-26
fixes #1423

### LINKED ISSUE #1423: Layout behaviour doesn't work properly for pdf
Once the Layout is switched on "scrollable" inside a pdf it can't be switched again. Trying to switch to "paginated" isn't working properly. 

How to reproduce it:
- Open a PDF
- Change layour to scrollable
- Try to change it again to paginated. It will be stuck in scrollable mode

panaC:


src/renderer/reader/components/ReaderOptions.tsx
```diff
@@ -581,55 +581,60 @@ export class ReaderOptions extends React.Component<IProps, IState> {
...
+                                        ? this.getButtonClassNamePdf(this.state.pdfView === "paginated")
+                                        : this.getButtonClassName("paged", true)}
+                                >
+                                    <SVG svg={PagineIcon} />
+                                    {__("reader.settings.paginated")}
+                                </label>
+                            </div>
+                        </div>
+                    </section>
+            }
```
danielweck, line 637: Post-merge suggestion: the `<section>` immediately below this one uses a different technique to test for PDF and eliminate from GUI: `hidden={this.props.isPdf}`, I think it would be better to use the same technique across the board (i.e. `hidden` attribute or `isPdf` ternary). I personally prefer the latter, as the `isPDF` state does not change in the reader view instance, so it makes total sense to completely "remove" the unused JSX fragments so that the generated DOM is as small as possible.

PR #1478, opened 2021-05-26
fixes #1449

### LINKED ISSUE #1449: Feature request : Keep TOC panel opened
It will be nice to be able to pin the TOC panel so TOC panel stays opened when clicking on a TOC entry.

Moreover, using a different style on the current TOC entry where the content is currently will be nice too (example : use bold for actual entry vs other entries in non bold).

It is useful when exploring a book to be able to jump fast from different TOC entries.

panaC:


src/renderer/reader/components/ReaderMenu.tsx
```diff
@@ -227,13 +227,13 @@ export class ReaderMenu extends React.Component<IProps, IState> {
                                     isRTL ? styles.rtlDir : " ")
                             }
                             onClick=
-                                {link.Href ? (e) => this.props.handleLinkClick(e, link.Href) : undefined}
+                                {link.Href ? (e) => this.props.handleLinkClick(e, link.Href, false) : undefined}
```
danielweck, line 230: Post-merge comment: I disagree with this breaking UX change. In fact, this PR implementation goes against @llemeurfr 's own suggestion:

> I'd personally prefer to keep the default behavior as-is (one click on a ToC entry or return key means move to the entry and close the panel)

https://github.com/edrlab/thorium-reader/issues/1449#issuecomment-848647409

danielweck: Follow-up PR: https://github.com/edrlab/thorium-reader/pull/1484

PR #1484, opened 2021-05-27
bookmark focus

panaC:
follow up #1449


danielweck:
Merging, to follow PR: https://github.com/edrlab/thorium-reader/pull/1478
...but will "revert" the change of behaviour, and make it consistent across the board (TOC, landmarks, bookmarks, search results)

PR #1496, opened 2021-06-04
[WIP] fixes #1492 (JavaScript Error while trying to open EPUB from another application)

### LINKED ISSUE #1492: JavaScript Error while trying to open EPUB from another application
I was working on a literature review and decided to keep track of the various documents with a mind map using XMind. I figured it would be a neat idea to attach links to my PDFs and EPUBs to the various nodes of my mind map. I thought that when double clicking a link, it should open the attached file. Please note the file is not embedded; only a link to the full path of the file is saved with each node (another reason in favour of not storing EPUBs in your own DB repository). When I double click a link to a PDF, it opens in default PDF viewer without any problem. However, when I double click a link to an EPUB, bang!

image

If I click on OK, the book opens but there is an empty Thorium window behind it that opens too:

image

The EPUB is still readable. The only problem is the loud warning sound, the error message and the "ghost" Thorium window.

I tried creating a shortcut in Windows to the EPUB to see if I could reproduce the problem, but it just opened the EPUB without any error. I then tried creating a link to an EPUB in OneNote and it opened the same EPUB without any error. Why is it not working with XMind?

panaC:
fixes #1492

### DIFF
--- a/src/main/lock.ts
+++ b/src/main/lock.ts
@@ -84,12 +84,16 @@ export function lockInstance() {
             // Someone tried to run a second instance, we should focus our window.
             debug("comandLine", argv, _workingDir);
 
-            const libraryAppWindow = getLibraryWindowFromDi();
-            if (libraryAppWindow) {
-                if (libraryAppWindow.isMinimized()) {
-                    libraryAppWindow.restore();
+            try {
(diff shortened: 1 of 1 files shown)


panaC:
I'm closing the PR
superseded by #1573

PR #1540, opened 2021-09-08
fix #1539

### LINKED ISSUE #1539: Go to feature crashes the app
I’m writing to report that in the Go to Page tool in the Navigation panel, when you press Enter on the keyboard, after entering the number in the "Enter a page number box", the program looks like crashing, turning to blank and you have to close it by pressing the "x" and re-open the file; you need to click the "Go to" rather than pressing "Enter" to navigate to the desired page.
 
I was wondering if that is a bug, or was it designed that way? Lots of times I just pressed Enter “automatically” after entering the number, only to have to close and re-open it.
 
Thanks again for the wonderful tool for reading EPUB!

panaC:


panaC:
https://github.com/edrlab/thorium-reader/commit/27f1588b342e26520dd174986783719899a50f34

panaC:
closeNavPanel doesn't works , why ?

panaC:
with `button` type=`submit` : `shift` + `alt` + `enter` : doesn't work

danielweck:
https://github.com/edrlab/thorium-reader/commit/d6991b095d37fbfe7121ad43f45aa0daa583ff3e#diff-9a0ee397868c7b388a6ee40b24d37f68f10f8585ad8a947c23553a3a52265acdR506

danielweck (approved):
Great fix, thank you!
In summary:
1) the `<button>` inside the `<form>` must be `type="button"` (not `type="submit"`) in order to intercept the `click` event with the `shift` and `alt` key modifiers (this was already correct in the original code)
2) ...BUT: because of the above point, the `input` inside the form de-facto becomes the form's `submit` action, which resulted in:
3) `enter` key triggers a full window reload in development mode (WebPack dev server), and a navigation bug in production
4) `shift` + `enter` key triggers the same but with an external window (detached from the main reader window)
5) ...so, this PR calls `preventDefault()` in the form's `submit` event handler, which disables the `input` action (only the `button` action remains)
6) ...so, this PR also migrates the `keypress` event handler from `input` to the `form` element, so that the `submit` action can be triggered directly in the text input field, without moving focus to the button (note: I tested that the `space` key can still be used in the text field, despite being a trigger when used with the `shift` + `alt` modifiers)
7) in closing: I reviewed the rest of the codebase for instances of `form` with `submit` not `preventDefault`'ed, there are some in the library settings tab but they are not impacted by the "new blank window" bug when pressing `shift` + `enter` due to using radio controls (not text input)

PR #1541, opened 2021-09-09
search broken with epub open from command line

### LINKED ISSUE #1517: Search broken in v1.7.1-alpha when book opened by double-clicking on EPUB in Windows File Explorer
I was running v1.7.1-alpha.1.842238767 since about mid May and did not notice it until today, but the search feature does not return any result, even for words present on the displayed page, when the book is opened by double-clicking on the EPUB in Windows File Explorer. However, if one returns to bookshelf and goes back to very same book, the search feature works again.

Thorium_FJxzjG7PIB

If Thorium is launched by itself, i.e. without double-clicking on an EPUB, and a book from the bookshelf is opened, the search feature works well.

Thorium_m4SFQncEzO

I downloaded v1.7.1-alpha.1.949738250 this morning and it still does not work normally. So I cannot tell at what point in time it stopped working, but it seems the search has been broken for a few versions, at least as far back as v1.7.1-alpha.1.842238767.

panaC:
fix #1517


panaC:
Tested on windows and it's works. It's fixed the issue

panaC:
explanation:  

When a publication is open on double-click with thorium-reader closed. The software starts the main process, launch the library window and doesn't wait the end of library window loading to launch the reader window with the publication wanted.
The library window is hided and only the reader appears. With this process the search is broken on window ( ressources not loaded ). 

I propose to wait the end of library window loading to launch the reader window. same process as a standard user that launch thorium and then open a publication ( search work fine in that case ).

In windows this workaround works well. I build an executable and tested it.

danielweck (approved):
This fix seems very reasonable to me, and safe in terms of the risk of regression bugs. Thank you :)

Issue #274, opened 2019-06-07
Adding a fixed OPDS feed to "Catalogs"


kevinmainairungu:
We see how the OPDS feed is stored after the user inputs the feed, but do not see a good way to add a fixed feed. Would this affect how the opds model is implemented? My thoughts were this would probably lead to a lot of instances being restructured. Would this be a constraint? Could we discuss a recommended implementation? Happy to take if forward, but would appreciate your thoughts on a solution.

llemeurfr:
Good question, as in many situations, an app based on Readium Desktop will at least list some possible OPDS feeds that the user could configure himself. And in some apps, pre-configured OPDS feeds will be offered. 
We should first discuss here the required changes in the software architecture, enabling these 2 use cases.

kevinmainairungu:
@llemeurfr thank you. Talk soon

danielweck:
A couple thoughts:
* Using the new CLI API, in principle Thorium / readium-desktop installations can be provisioned with pre-defined OPDS feed URLs via a post-install script (e.g. as part of a NSIS installer for Windows, for example).
* We can also hard-code a curated selection of OPDS feeds into the application, so that users do not have to manually enter URLs (e.g. Feedbooks, but there are many more candidates, for example https://github.com/readium/r2-streamer-js/blob/develop/docs/opds.md#a-selection-of-public-opds-feeds )

llemeurfr:
@kevinmainairungu the first solution proposed by Daniel seems the most flexible, as it can be tailored to different use cases, still using the same application. What is your opinion on that?

kevinmainairungu:
@llemeurfr I agree. This can be fully modular and can be re-used in different cases with the same application. A post-install script will be much easier than having to hard-code the feed every time.

llemeurfr:
Do you think you could develop this feature with our help? If yes what do you need?

danielweck:
Related issue: https://github.com/readium/readium-desktop/issues/608

panaC:
may i close this issue ?

danielweck:
I am in favour of closing this issue, as well as this one: https://github.com/edrlab/thorium-reader/issues/608
I do not think that we have a consensus to promote a particular set of OPDS feeds directly from within Thorium. Adding OPDS feeds via the UI is easy enough. If vendors need to automatically include a list of feeds in their builds, they can easily do this in their fork, no need to add the overhead of an additional API in Thorium to do this. Just my two cents :)

[...]

PR #1546, opened 2021-09-17
[fix #1536] olds problem details

### LINKED ISSUE #1536: OPDS, handle HTTP Problem Details
The support of HTTP Problem Details should be as generic as possible in Thorium. Here is a first use case, but every HTTP error returned by an OPDS server should be treated as well. 

Use case: library books have a limited count of copies, and when all copies are already used, a new user cannot borrow it.

In practice: an OPDS feed contains acquisition links of type "borrow". A link is activated, but the server wants to abort the action. It does that by returning an HTTP error + a json object detailed in rfc7807.

Several applications handle such HTTP Problem Details documents, including Aldiko and SimplyE. There is explicit mention of this in ODL: https://drafts.opds.io/odl-1.0#54-interacting-with-a-checkout-link

The `title` and `details` of the HTTP Problem Details document should be based on the Accept-Language header of the client request. Therefore a modela box containing this information is displayed to the user. 

More information: see Readium Slack, channel opds, 31/08/2021

panaC:
fix #1536


danielweck:
Hello Pierre, do you consider this PR ready to merge, or draft/WIP?

panaC:
> Hello Pierre, do you consider this PR ready to merge, or draft/WIP?

need Laurent and you review

panaC:
I doesn't yet translate the error page , I'm waiting your approbation

src/renderer/common/redux/sagas/browse.ts
```diff
@@ -18,12 +18,12 @@ import { apiSaga } from "./api";
...
 const debug = debug_(filename_);
 
-type TA = apiActions.result.TAction<TReturnPromiseOrGeneratorType<TApiMethod["opds/browse"]>>;
+type TA = apiActions.result.TAction<TReturnPromiseOrGeneratorType<TApiMethod["browser/browse"]>>;
 
 export function* opdsBrowse(link: string, REQUEST_ID: string) {
 
     debug("opds-browse", link);
-    yield apiSaga("opds/browse", REQUEST_ID, link);
+    yield apiSaga("browser/browse", REQUEST_ID, link);
```
danielweck, line 26: Can you please explain the rationale for the change from `opds/browse` to `browse/browse`? If the intent is to signal that this API will / can work with any HTTP request/response cycle, then I understand, but `browse/browse` looks like a typo rather than an intentional entry point in the API surface. Would `http/browse` be preferable, or is that a misnomer too?

danielweck: ...furthermore, there is a file renaming from `opdsBrowse.ts` to `browse.ts` which signals the generic nature of the API, but the function is still named `opdsBrowse()`:

```
export function* opdsBrowse(link: string, REQUEST_ID: string) {
...
}
```

danielweck (changes requested):

src/renderer/common/redux/sagas/browse.ts
danielweck: ...I also note that the `IBrowserResultView` signals a generic HTTP request/response use-case, yet its object shape / model seems to be tightly-coupled with OPDS:

```
export interface IBrowserResultView {
    opds?: IOpdsResultView;
    problemDetails?: IOpdsProblemDetailsResultView;
}
```

danielweck (changes requested):

llemeurfr:
Re the UI: it should be user oriented, i.e. the textual message is important, the http error type and code are much less important, the details and instance should not be displayed. 

So the message could be: 

**The request has failed** (-> localized string, generic title for all error messages)
You have reached ... library (-> a string localized by the server from the Accept-Language header)
(http 403 - loans_limit_reached) (-> more technical info, smaller font, mostly used by developers)

src/renderer/common/redux/sagas/browse.ts
panaC: thanks 

`browser/browse` replace with `httpbrowser/browse` instead

panaC:
ready to merge

src/common/views/opds.ts
```diff
@@ -144,4 +143,11 @@ export interface IOpdsNavigationLink {
     self: IOpdsLinkView[];
 }
 
-export type THttpGetOpdsResultView = THttpGetResultAfterCallback<IOpdsResultView>;
+// https://datatracker.ietf.org/doc/html/rfc7807#section-3.1
+export interface IOpdsProblemDetailsResultView {
```
danielweck, line 147: Why not `interface IOpdsProblemDetailsResultView extends IProblemDetailsResultView` ... or even `type IOpdsProblemDetailsResultView = IProblemDetailsResultView` given that there is currently no difference?

src/main/api/browser.ts
```diff
@@ -0,0 +1,96 @@
+import { ok } from "assert";
```
danielweck, line 1: Please use `import { ok } from "readium-desktop/common/utils/assert";` (remember, we found critical runtime bugs with the `assert` module!)

src/renderer/common/redux/sagas/browse.ts
```diff
@@ -18,12 +18,12 @@ import { apiSaga } from "./api";
 const filename_ = "readium-desktop:renderer:redux:saga:opds-browse";
```
danielweck, line 18: I think that this TypeScript file should keep its original name `opdsBrowse.ts`, as the exported function is called `opdsBrowse()` and its `debug()` prefix is `readium-desktop:renderer:redux:saga:opds-browse`.

danielweck (changes requested):

src/common/views/opds.ts
panaC: forgot to remove it

src/main/api/browser.ts
panaC: done

src/renderer/common/redux/sagas/browse.ts
panaC: done

panaC:
@danielweck  can you merge it, please, I need it for the #274

PR #1549, opened 2021-09-20
[WIP] [fix #1544] http request with file:// URI

### LINKED ISSUE #1544: OPDS: allow the use of file URIs
In situations where the internet is not available and shared ebooks (especially textbooks) are stored in a network disk drive (i.e. a NAS + Wifi), it is useful to enable the use of OPDS feeds referenced via a file URI, and in which every reference is also a file URI.  

On Windows, file paths are formatted as `file:///c:/path/to/the%20file.txt`. 
On a Mac or Linux box, they are formatted as `file://hostname/path/to/the%20file.txt, or `file:///path/to/the%20file.txt` for localhost.

panaC:
fixes #1544


src/main/network/http.ts
```diff
@@ -453,7 +454,53 @@ export const httpGetWithAuth =
...
+                debug("wrong URL : fallback to httpGetWithAuth");
+                debug(e);
+            }
+
+            if (url?.protocol === "file:") {
+
+                let isFailure = false;
+                let stream: NodeJS.ReadableStream | undefined;
+                try {
+                    stream = createReadStream(url.pathname);
```
danielweck, line 476: Enormous security risk (full disk access).

danielweck (changes requested):

src/main/network/http.ts
danielweck: So, a maliciously-crafted OPDS feed, LCP license, etc. could trigger arbitrary filesystem location access. This wouldn't pass a security audit, even if in practice the recipient of the HTTP response would be limited to some internal Thorium API, because an arbitrary filesystem path could be an operating system hook (not necessarily a downloadable binary blob). In other words, a mounted filesystem should be regarded as a sensitive API surface, and Thorium should implement provisions to secure access (e.g. ask user permissions, limit to known folders, etc.).

panaC: https://code.visualstudio.com/docs/editor/workspace-trust

panaC: should we used the same mechanism ?

PR #1551, opened 2021-09-21
[fix #298] search by author in library

### LINKED ISSUE #298: Allow searching by author
Currently the search is limited to the book title. The name of the author should also be a search token.

panaC:
fixes #298

### DIFF
--- a/src/main/db/repository/publication.ts
+++ b/src/main/db/repository/publication.ts
@@ -213,7 +213,7 @@ export class PublicationRepository  /* extends BaseRepository<PublicationDocumen
     }
 
 
-    public async searchByTitleAndAuthor(title: string): Promise<PublicationDocument[]> {
+    public async searchByTitleAndAuthor(titleOrAuthor: string): Promise<PublicationDocument[]> {
 
         try {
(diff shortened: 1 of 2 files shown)


src/main/db/repository/publication.ts
```diff
@@ -212,7 +213,7 @@ export class PublicationRepository  /* extends BaseRepository<PublicationDocumen
     }
 
 
-    public async searchByTitle(title: string): Promise<PublicationDocument[]> {
+    public async searchByTitleAndAuthor(title: string): Promise<PublicationDocument[]> {
```
danielweck, line 216: Nit picking: could you please rename `title` to a more generic variable name like `str` or `keyword`?

danielweck (changes requested):

src/main/db/repository/publication.ts
danielweck: Fixed via https://github.com/edrlab/thorium-reader/pull/1551/commits/1c90a5b058f77b8e09a621fedbe3b09104679f01

PR #1552, opened 2021-09-23
opds: fix opds search error for calibre-web.

schspa:
Signed-off-by: schspa <schspa@gmail.com>

### DIFF
--- a/src/main/services/opds.ts
+++ b/src/main/services/opds.ts
@@ -151,6 +151,7 @@ export class OpdsService {
                 if (url.search.includes(SEARCH_TERM) ||
                     tryDecodeURIComponent(url.pathname).includes(SEARCH_TERM)) {
 
+                    atomLink.url = atomLink.url.replace("%7B", "{").replace("%7D", "}")
                     debug("parseOpdsSearchUrl (atomLink): ", atomLink.url);
                     return (atomLink.url);
                 }


schspa:
@danielweck I don't know much about typescript or javascript.
But this can fix my problem when searching in opds server deployed by [calibre-web](https://github.com/janeczku/calibre-web)

panaC:
Thanks you to report this. we will investigate

src/main/services/opds.ts
```diff
@@ -151,6 +151,7 @@ export class OpdsService {
                 if (url.search.includes(SEARCH_TERM) ||
                     tryDecodeURIComponent(url.pathname).includes(SEARCH_TERM)) {
 
+                    atomLink.url = atomLink.url.replace("%7B", "{").replace("%7D", "}")
```
danielweck, line 154: The `String.replace()` function with a string function argument replaces only the first occurrence. I think you need to use a regular expression to "replace all":

`atomLink.url = atomLink.url.replace(/%7B/g, "{").replace(/%7D/g, "}")`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

Also, there is a missing trailing semicolon (lint)

danielweck (changes requested):

src/main/services/opds.ts
danielweck: I wonder why such an ad-hoc string replacement is needed. As we haven't come across this problem in our tests, I am concerned about introducing a regression bug. Normally this would be handled by `decodeURIComponent()`. Do you have an example of Atom URL that exemplifies the problematic syntax?

danielweck (changes requested):

schspa:
@danielweck 
Sorry for the late.
this is my calibre-web opds feed
https://pastebin.com/raw/0r2BaXwy

And search atom link is:
`<link type="application/atom+xml" rel="search" title="搜索" href="/opds/search/{searchTerms}" />`
Please check it.

panaC:
https://gistcdn.githack.com/panaC/659a81199e2ddc762315c877dcc90755/raw/715cc5df3a210048d662f8abd26692b1fd0ec974/calibre-web.xml

image

danielweck:
Thank you @schspa for reporting this. Thanks you @panaC for the investigation.
Fixed via https://github.com/edrlab/thorium-reader/commit/1d11eb37de704cdf36fb41d0d855d84a8c8cbdac

PR #1555, opened 2021-09-23
[fix #1554] form-urlencoded

### LINKED ISSUE #1554: OAuth with Password grant sent as JSON instead of URL encoded form
The OAuth with Resource Owner Password OPDS authentication flow is supposed to follow the OAuth RFC6749.

> In order to identify the use of a Resource Owner Password Credentials Grant Authentication Flow (as defined in [RFC6749] in section 4.3. Resource Owner Password Credentials Grant)
> https://drafts.opds.io/authentication-for-opds-1.0.html#346-resource-owner-password-credentials-grant

The RFC6749 states that the payload must be an URL encoded form:

> The client makes a request to the token endpoint by adding the following parameters using the "application/x-www-form-urlencoded" format
> https://datatracker.ietf.org/doc/html/rfc6749#section-4.3.2

But Thorium is sending the payload as a JSON object, which breaks with spec-compliant servers:

https://github.com/edrlab/thorium-reader/blob/3e4aa1018bbd0ae0877daa44a02b3235e2300685/src/main/redux/sagas/auth.ts#L291

panaC:
fixes #1554

### DIFF
--- a/src/main/redux/sagas/auth.ts
+++ b/src/main/redux/sagas/auth.ts
@@ -283,12 +283,12 @@ async function opdsSetAuthCredentials(
                         }
 
                         const headers = new Headers();
-                        headers.set("Content-Type", ContentType.Json);
+                        headers.set("Content-Type", ContentType.FormUrlEncoded);
 
                         const { data: postData } = await httpPost<IOpdsAuthenticationToken>(
                             authenticateUrl,
                             {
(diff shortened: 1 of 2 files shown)


panaC:
tested with opds2-auth-test-server

danielweck (approved):
Great, thank you Pierre! :)

PR #1608, opened 2021-12-21
feat(a11y): "where am i?" information in publication popup, CTRL+i keyboard shortcut with SHIFT modifier key to focus

danielweck:
This is a working baseline with support for audiobook, PDF, Divina, EPUB fixed layout and reflow.
TODO: navigator component to convey preceding heading structure in document (HTML outline) + situation relative to TOC (EPUB navigation document)

### DIFF
--- a/src/common/keyboard.ts
+++ b/src/common/keyboard.ts
@@ -38,6 +38,12 @@ const _defaults_ = Object.freeze({
         shift: false,
         key: "KeyI",
     }),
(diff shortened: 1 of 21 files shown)


danielweck:
Follow-up commit: https://github.com/edrlab/thorium-reader/commit/5907f0a1643ffe0e1bfac0d95fc02bc153a24642
(added missing headings trail leading to current reading location in "where am I" section of publication info popup modal dialog)

PR #1622, opened 2022-01-13
[WIP] removal of SQLite3 and LevelDown native database dependencies

danielweck:
Fixes https://github.com/edrlab/thorium-reader/issues/1619
(see the above issue for details of the rationale for eliminating native DB)


danielweck:
Hello @panaC I tested this PR successfully. This removes the DB code ("absorb" / read-only logic), but this does not delete the DB files on the filesystem. I suggest we implement this in another PR.
Could you please make a backup of your current `node_modules` folder (just in case you need to continue working in the `develop` until this PR is merged), then `rm -rf node_modules` and `npm install` => you will see that `electron-rebuild` does not find any native NodeJS modules / native C/C++ libs to compile via Node-GYP, which is great! (no time-consuming `postinstall`). In principle we could remove the `electron-rebuild` dependency entirely, but let's keep it in for now.
I think this PR is ready to merge, let me know your thoughts. This will bump version 1.9 to 2.0

panaC:
seems very good , thanks Daniel !

panaC (approved):
amazing !

PR #1646, opened 2022-02-26
Add Linux .desktop file keywords to package.json

elcste:
Adding keywords to the Linux .desktop file makes it easier to search and find Thorium without using the application name.

Note: I haven't tried building the Thorium package to confirm this works, but I referred to the electron-builder docs and compared with some other Electron apps' package.json files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -221,7 +221,7 @@
       ],
       "desktop": {
         "MimeType": "application/epub+zip",
-        "Keywords": "Audiobook;Book;Ebook;EPUB;Reader;Readium;Viewer;",
+        "Keywords": "Audiobook;Book;Ebook;EPUB;Reader;Readium;Viewer;"
       },
       "executableName": "thorium",
       "category": "Office"
--- a/package.json
+++ b/package.json
@@ -220,7 +220,8 @@
         "AppImage"
       ],
       "desktop": {
-        "MimeType": "application/epub+zip"
+        "MimeType": "application/epub+zip",
+        "Keywords": "Audiobook;Book;Ebook;EPUB;Reader;Readium;Viewer;",
       },
       "executableName": "thorium",
       "category": "Office"


panaC:
I do not see 'keyword' in the documentation https://www.electron.build/configuration/linux

elcste:
It says 

```
desktop any | “undefined” - The [Desktop file](https://developer.gnome.org/integration-guide/stable/desktop-files.html.en) entries (name to value).
```

which I understand as meaning you can add other desktop file entries under that. See for example this comment: https://github.com/electron-userland/electron-builder/issues/4974#issuecomment-631072452

I am very happy to test the CI build to confirm it works or not if you add it :-)

package.json
```diff
@@ -220,7 +220,8 @@
         "AppImage"
       ],
       "desktop": {
-        "MimeType": "application/epub+zip"
+        "MimeType": "application/epub+zip",
+        "Keywords": "Audiobook;Book;Ebook;EPUB;Reader;Readium;Viewer;"
```
danielweck, line 224: @llemeurfr any other "keyword" suggestions?

Syntax specification:
https://developer.gnome.org/documentation/guidelines/maintainer/integrating.html#desktop-files

Electron Builder doc (Linux-specific section):
https://www.electron.build/configuration/linux
=>
https://github.com/edrlab/thorium-reader/blob/2ad37fa8d4753c308b6d39b3f7c9439e09d93447/package.json#L217-L227

...Windows:
https://www.electron.build/configuration/win
=>
https://github.com/edrlab/thorium-reader/blob/2ad37fa8d4753c308b6d39b3f7c9439e09d93447/package.json#L210-L216

...and Mac:
https://www.electron.build/configuration/mac
=>
https://github.com/edrlab/thorium-reader/blob/2ad37fa8d4753c308b6d39b3f7c9439e09d93447/package.json#L228-L236


Also note the general `fileAssociations` field (keywords there too, but for a different purpose):

https://github.com/edrlab/thorium-reader/blob/2ad37fa8d4753c308b6d39b3f7c9439e09d93447/package.json#L163-L209

danielweck (approved):
Good choice of keywords. Let's merge :)

elcste:
I installed the latest dev version deb package on Ubuntu 22.04, and I can now search "ebook", "epub" etc and find Thorium :100:

danielweck:
That's great. I only occasionally use Linux (I spent most of my time in the MacOS/Darwin console and tools), I didn't realise that this Linux feature existed :)

elcste:
@danielweck That was my background too before switching to Linux at home. The Linux desktop app world can be very byzantine!

PR #1652, opened 2022-03-13
fix(l10n): Updated Spanish translation

nvdaes:


danielweck:
Wonderful, thank you very much Noelia :)

PR #1655, opened 2022-03-16
Updating empty strings and minor correction in Swedish locale

martinpub:
Hi, just keeping pace with the Swedish translation.


danielweck:
Much appreciated, thank you! :)

PR #1668, opened 2022-04-11
Updated Spanish translation

HadrienGardeur:
Based on a review made by Libranda, here's an updated Spanish translation for Thorium.

### DIFF
--- a/src/resources/locales/es.json
+++ b/src/resources/locales/es.json
@@ -69,7 +69,7 @@
         "lang": "Idioma",
         "moreInfo": "Más información",
         "myBooks": "Mis libros",
-        "noPublicationHelpL1": "Suelta una publicación aquí",
+        "noPublicationHelpL1": "Desliza una publicación aquí",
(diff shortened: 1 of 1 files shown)


danielweck:
Thank you very much, I updated the wiki page https://github.com/edrlab/thorium-reader/wiki/Localization-(l10n)-language-translations
Note that @nvdaes (Noelia) is the original contributor of the Spanish translation. I am merging this Pull Request immediately, but I want to make sure that Noelia is aware of the proposed (and now approved) changes.

nvdaes (commented):
Hello: Thanks for this update. I see that this fixes some mistakes made by me: autor instead of my mistake author, se instead of Se, and (lectura en voz alta), where I ommited () symbols.
Other messages are similar and maybe a matter of preference. For router instead of rúter, Fundeu prefers to use "rúter":

https://www.fundeu.es/dudas/palabra-clave/ruter/

Anyway, router is also acceptable.
Kind regards.

PR #1673, opened 2022-04-12
l10n/ga

danielweck:
follow-up to https://github.com/edrlab/thorium-reader/pull/1671


sukiletxe:
Woops, failed to spot this before. `ga` is gaelic, Galician is `gl`. I don't know where the short codes are used, but still.

danielweck:
No problems, I will hotfix this now :)

PR #1693, opened 2022-05-26
React 18 compatibility

danielweck:
This PR works, as far as I can tell without noticeable regressions.
However, I propose that we do **not** merge into the `develop` branch / mainstream releases, because this would force Thorium forks to adapt strict TypeScript typing updates, and it may in fact also invalidate the use of legacy React 16-17 lib dependencies. I would rather encourage forks to upgrade to the latest official Git revision of Thorium, than discourage them to upgrade because of our eagerness to track React's latest updates (which are know to break parts of the ecosystem ... but thankfully without negative side effects as far as I know in Thorium).


danielweck:
The magnitude of changes in the ongoing GUI pull requests (see attached screenshot) is prompting me to revisit the impact of breaking changes from this React 18 PR. I now think that the timing is right to introduce the move away from React 17. This will have repercussions (possibly negative) in Thorium forks, but not nearly as much as some of the GUI refactorings in the pending PRs, so I think we should schedule the React 18 evolution for the next version of Thorium, alongside all the GUI improvements (design, and implementation).

danielweck:
Screenshot 2023-10-27 at 10 37 08

PR #1707, opened 2022-06-09
Translation pt-PT - Update from PR 1684

horus68:
Update pt-PT translation with new strings from #1684


danielweck:
Thank you very much!

PR #1711, opened 2022-06-14
Telemetry

panaC:


src/main/redux/sagas/telemetry.ts
```diff
@@ -0,0 +1,122 @@
...
+
+function* collectAndSave() {
+
+    const version = yield* select((state: RootState) => state.version);
+
+    // fresh install is equal to en language ?
+    const locale = yield* select((state: RootState) => state.i18n.locale);
+
+    let fresh = false;
+    if (_APP_VERSION !== version) {
```
danielweck, line 72: could we please add a comment here to explain that:

yield call(telemetry.collectSaveAndSend);

is invoked before:

 yield put(appActions.initSuccess.build());

...which is essential for the version comparison to work.

thanks :)

src/main/redux/sagas/telemetry.ts
```diff
@@ -0,0 +1,122 @@
...
+        fresh = true;
+    }
+
+    const info = {
+        os_version: osVersion(),
+        locale,
+        timestamp: new Date().toISOString(),
+        fresh,
+        type: "poll", // poll or error emun
+        actual_version: _APP_VERSION,
```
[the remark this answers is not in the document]
panaC: done

src/main/redux/sagas/telemetry.ts
panaC: done

PR #1712, opened 2022-06-19
Update spanish translation

nvdaes:


nvdaes:
Among other additions, this fixes a serious mistake in the translation of "Copied characters". Characters was translated as "personajes" (people or persons) instead of "caracteres" (symbols that have been copied).

danielweck:
Very nice, thank you Noelia!!

nvdaes:
Thank you for mergin, Daniel.

Enviado desde mi iPhone

PR #1803, opened 2022-08-29
set a better message on open-url with a unrecognized content-type from the server

panaC:


danielweck (approved):
I added the missing error string which gets displayed in the toast notification

PR #1806, opened 2022-08-29
fix 1603 cancelling downloads

### LINKED ISSUE #1603: Colour contrast, hover effect: Left and Right buttons
The colour contrast between the left and right buttons and their background is quite low, at 1.8 to 1.
I used the Colour Contrast Analyzer developed by the Paciello Group

The following image shows the left and right buttons. 
image

Additionally there is no hover over effect with the mouse, as there is when the user tabs to it.

panaC:
fixes #1603


danielweck (approved):
tested, it works. I don't really understand the details of this Redux Saga construction, but I can roughly see what is happening  :)

PR #1807, opened 2022-08-30
fixes the array reducing to display author in publicationInfo

### LINKED ISSUE #1586: Book "info"  (in reader) not listing authors correctly
For books with multiple authors, the bookshelf shows multiple authors while the "info" in the reader only shows one author -- and not necessarily the first one in the opf order.

Example:

[author-meta.epub.zip](https://github.com/edrlab/thorium-reader/files/7495934/author-meta.epub.zip) (zippe'd epub)

Bookshelf:

screenshot from bookshelf, in particular authors are given as "B Author, A Author"

Reader => info:

screenshot from reader's info panel, in particular authors are given as "A Author"

panaC:
fixes #1586


danielweck (approved):
Ah yes indeed, the `reduce` was totally ignoring the accumulator!!

PR #1808, opened 2022-08-31
fixes window menu in macos

### LINKED ISSUE #1137: MacOS: Window menu / Menu bar
As a Mac user I find it unusual for the Window menu to be hidden underneath the Application menu.

image

It took me a while to find it and most Mac users would expect it to be in the main menu bar.

image

panaC:
fixes #1137


danielweck (approved):
tested, works.

PR #1819, opened 2022-09-08
fixes go back to history after canceled authentication

### LINKED ISSUE #1405: OPDS authentification dialog: when cancelled, go back to feed root (not 401 page)
When the login form appears, if I select cancel, I'm looping in a 401 error which redirects me to the form.

panaC:
fixes #1405


danielweck (approved):
clever solution :)

PR #1823, opened 2022-09-09
removed unnecessary and broken MathJax dependency in src/package.json

panaC:
(Electron Builder production dependencies)


danielweck:
Merci Pierre :)

PR #1832, opened 2022-10-06
calibre content server with digest authentication

### LINKED ISSUE #1829: Can not connect to calibre server with authentication enabled.
Tried to connect to my calibre server which has enabled authentication. 

I can access the opds data in a web browser with correct credentials. Thorium does not ask me for credentials and the error message says "message: Unauthorized | url: https://xxxxx | type: text/plain;charset=UTF-8 | code: 1000".

panaC:
fixes #1829


src/utils/digest.ts
```diff
@@ -0,0 +1,88 @@
...
+    padding = typeof (padding) === "undefined" || padding === null ? padding = 2 : padding;
+
+    while (hex.length < padding) {
+        hex = "0" + hex;
+    }
+
+    return hex;
+}
+
+export const parseHeaderWWWAuthenticateForDigest = (v: string): { [s: string]: any } => {
```
danielweck, line 21: `[s: string]: any` can be `[s: string]: string`, right?

panaC: yes it is

src/main/services/opds.ts
```diff
@@ -98,14 +109,20 @@ export class OpdsService {
...
+                const isValid = this.wwwAuthenticateIsValid(wwwAuthenticate);
+                if (isValid) {
                     const result: IOpdsResultView = {
                         title: "Unauthorized",
                         publications: [],
                     }; // need to refresh the page
+
+                    const data = this.parseWwwAuthenticate(wwwAuthenticate);
+                    if (!data.type) {
+                        result.title = "Unauthorized (bad WWWAuthenticate type)";
```
danielweck, line 121: Suggestion:
```
result.title = `Unauthorized (unsupported WWWAuthenticate type '${wwwAuthenticate.trim().split(" ")[0]}')`;
```

src/utils/digest.ts
```diff
@@ -0,0 +1,88 @@
...
+    method: string,
+    nonceCount: string,
+}) => {
+
+    nonceCount = nonceCount ? decimalToHex(parseInt(nonceCount, 16) + 1, 8) : decimalToHex(1, 8);
+    const ha1MD5 = createHash("md5").update(`${username}:${realm}:${password}`).digest("hex");
+    const ha1 = algorithm === "MD5-sess" ? createHash("md5").update(`${ha1MD5}:${nonce}:${cnonce}`).digest("hex") : ha1MD5;
+    const ha2 = createHash("md5").update(qop === "auth-int" ? "" : `${method}:${uri}`).digest("hex"); // qop === "auth-int" not supported what is entityBody?
+    const response = createHash("md5").update((qop === "auth" || qop === "auth-int") ? `${ha1}:${nonce}:${nonceCount}:${cnonce}:${qop}:${ha2}` : `${ha1}:${nonce}:${ha2}`).digest("hex");
+    const accessToken = `username="${username}", realm="${realm}", nonce="${nonce}", qop=${qop}, algorithm=${algorithm}, response="${response}", uri="${uri}", nc=${nonceCount}, cnonce="${cnonce}"`;
```
danielweck, line 85: Can we please add a comment to explain why `qop`, `algorithm` and `nc` are "unquoted strings". A link to the specification should suffice:
https://httpwg.org/specs/rfc7616.html#rfc.section.3.4
> For historical reasons, a sender MUST only generate the quoted string syntax values for the following parameters: realm, domain, nonce, opaque, and qop.
> For historical reasons, a sender MUST NOT generate the quoted string syntax values for the following parameters: stale and algorithm.

danielweck, line 85: The key/value pair `username="${username}"` is incorrect if `username` contains characters forbidden in the "quoted string" syntax, or more generally Unicode characters. Instead, `username*=` must be used.
Suggestion: maybe consider passing `userhash=true` and normalise `username` via a hashing function that generates a valid "quoted string"? (not sure about Calibre server support though...)

See:
https://httpwg.org/specs/rfc7616.html#rfc.section.3.4
Unicode example:
> username*=UTF-8''J%C3%A4s%C3%B8n%20Doe

danielweck (changes requested):

src/utils/digest.ts
panaC: there are no implementation of `username*=` or `userhash=` in calibre I think

https://github.com/kovidgoyal/calibre/blob/eb78a761a99ac20a6364f85e12059fec6517d890/src/calibre/srv/tests/auth.py

panaC: they used https://www.rfc-editor.org/rfc/rfc2617

src/main/services/opds.ts
```diff
@@ -162,55 +179,36 @@ export class OpdsService {
...
-        _realm: string,
+        data: IDigestDataParsed & {type: "digest" | "basic"},
         responseUrl: string,
     ) {
 
         const opdsAuthDoc = new OPDSAuthenticationDoc();
 
         opdsAuthDoc.Id = "";
-        opdsAuthDoc.Title = ""; // realm || "basic authenticate"; NOT HUMAN-READABLE!
+        opdsAuthDoc.Title = "Login";
```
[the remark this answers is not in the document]
panaC: https://github.com/edrlab/thorium-reader/blob/411f6312a856469a7f8fac8a728f16b31e052f9f/src/main/redux/sagas/auth.ts#L668

src/utils/digest.ts
danielweck: Resolving this issue as a no-fix, for now. As this is a known implementation caveat, I am adding a code comment to record the information, in case the bug surfaces in the real world in the future.

danielweck: See 21f490f73bccff38a786b172fb584cb58d84a3c3

danielweck: Fixed: 36869c2becabe0eff6961e2e559196f7e99f5161

src/main/services/opds.ts
danielweck: Fixed: 0ea8746e255189d217489af9898603f251da3a29

danielweck (approved):

PR #1840, opened 2022-10-17
fix #1635

### LINKED ISSUE #1635: keyboard characters y and n for yes/no cannot be used to eg. delete a book form My Books?
I cannot use 'y' and 'n' for 'yes' and 'no'?
In My Books view -  I choose the action button next to the cover img and 'Delete'
image
The delete window now opens:
image
If I type 'y' now with my ordinary keyboard, I would think the file was deleted - but this doesn't happen. 
Instead keybord focus is activated and fixed on the close button, it seems. And I have to tab two times before focus sets on the 'yes' activation button and I can delete the file with enter.
Just an observation.

panaC:
fix #1635

### DIFF
--- a/src/renderer/library/components/dialog/DeletePublicationConfirm.tsx
+++ b/src/renderer/library/components/dialog/DeletePublicationConfirm.tsx
@@ -38,12 +38,18 @@ class DeletePublicationConfirm extends React.Component<IProps, undefined> {
         super(props);
 
         this.remove = this.remove.bind(this);
+
+        this.yesButtonRef = React.createRef();
     }
+
+    private yesButtonRef: React.RefObject<HTMLButtonElement>;
 
     public render(): React.ReactElement<{}> {
         if (!this.props.open || !this.props.publicationView) {
             return <></>;
         }
+
+        setTimeout(() => this.yesButtonRef.current.focus(), 0);
 
         const { __, closeDialog } = this.props;
         return (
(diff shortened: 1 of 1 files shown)


danielweck:
quick code review, warning:  the fix should introduce keyboard mnemonics / quick-access single-key shortcuts that activate buttons based on underlined label character (e.g. y or n for yes or no). Naturally, this introduces localisation / internationalisation issues.

the "default keybaord focus in dialog action buttons" is a separate issue, which impacts all dialog components (not just the one changed in this PR)

src/renderer/library/components/dialog/DeletePublicationConfirm.tsx
```diff
@@ -38,13 +38,19 @@ class DeletePublicationConfirm extends React.Component<IProps, undefined> {
...
     }
 
+    private yesButtonRef: React.RefObject<HTMLButtonElement>;
+
     public render(): React.ReactElement<{}> {
         if (!this.props.open || !this.props.publicationView) {
             return <></>;
         }
 
+        setTimeout(() => this.yesButtonRef.current.focus(), 0);
```
danielweck, line 52: This is incorrect React code: first of all, `this.yesButtonRef.current` can be null, but more importantly `render()` only constructs the VDOM and may in fact be called multiple times. There is no guarantee that the actual DOM is ready at that point (i.e. at the `setTimeout()` tick). That's what a functional component's `useEffect()` is for (or class component equivalent).

I am working on a solution that works for all possible dialogs in Thorium (we need a consistent logic for focus management, including returning to the calling site).

danielweck (changes requested):

PR #1847, opened 2022-10-24
publication info description : punctuation and separators should be localized#962

### LINKED ISSUE #962: Publication info popup dialog (metadata), punctuation and separators should be localized
For example, in French, colon separator is between space characters (before and after), whereas currently there is a space only after (which is fine for English locale).

Thorium_publication-metadata-id_1

Thorium_publication-metadata-id_2

panaC:
fix #962


danielweck (changes requested):
This PR introduces regressions due to labels being used elsewhere with the expectation of not containing punctuation (e.g. no `:` suffix). I will check all occurrences (a quick code search reveals that `catalog.publisher` is used in the "data grid" bookshelf.

Issue #1849, opened 2022-10-31
Localisation update, translations for Thorium 2.2


gautierchomel:
Dear contributors, Thorium 2.2 will be released around November 7, it has some new features that require translation. 

Following is an extraction of the changed and added values, you can get a similar but in context view of changes in the reference file en.json with the following command line : `git diff v2.1.0:src/resources/locales/en.json develop:src/resources/locales/en.json`

All features can be tested in the [latest continuous test builds](https://github.com/edrlab/thorium-reader/releases).

## 1 Change: *More info* becomes *Book details*
In the Book information panel (About this book), the *More info* section is now titled *Book details* wich correspond better to it content. 

* [line 75](https://github.com/edrlab/thorium-reader/blob/1d48f88691134b8d0f53bbe5a0f70b7f0d3d6904/src/resources/locales/en.json#L75): "moreInfo": "Book Details",

### edit OPDS feeds
Thorium now allows users to edit OPDS feeds (name and adress). This triggers the followings added fields : 

* [line 112](https://github.com/edrlab/thorium-reader/blob/1d48f88691134b8d0f53bbe5a0f70b7f0d3d6904/src/resources/locales/en.json#L112): "update": "Edit"
+ [line 207](https://github.com/edrlab/thorium-reader/blob/1d48f88691134b8d0f53bbe5a0f70b7f0d3d6904/src/resources/locales/en.json#L207): "name": "Name:",
+ [line 208](https://github.com/edrlab/thorium-reader/blob/1d48f88691134b8d0f53bbe5a0f70b7f0d3d6904/src/resources/locales/en.json#L208): "namePlaceholder": "Name",
+ [line 209](https://github.com/edrlab/thorium-reader/blob/1d48f88691134b8d0f53bbe5a0f70b7f0d3d6904/src/resources/locales/en.json#L209): "title": "Edit an OPDS feed",
+ [line 210](https://github.com/edrlab/thorium-reader/blob/1d48f88691134b8d0f53bbe5a0f70b7f0d3d6904/src/resources/locales/en.json#L210): "updateButton": "Edit",
+ [line 211](https://github.com/edrlab/thorium-reader/blob/1d48f88691134b8d0f53bbe5a0f70b7f0d3d6904/src/resources/locales/en.json#L211): "url": "Link:",
[...]

PR #1851, opened 2022-10-31
Update Lithuanian translation for Thorium 2.2

naglis:
Note: prior to translating I ran the commands from Localization wiki [Tooling](https://github.com/edrlab/thorium-reader/wiki/Localization-(l10n)-language-translations#tooling) section, as a result some strings were removed/added.

Related issue: #1849


danielweck:
follow-up PR https://github.com/edrlab/thorium-reader/pull/1885

PR #1854, opened 2022-10-31
Update spanish translation for release 2.2

nvdaes:
- friendy has been translated as "amigable con", according to this [recommendation at FundéuRAE](https://www.fundeu.es/recomendacion/amigable-con-alternativa-valida-friendly/).
- Conforms to: translated without colon since generally it shouldn't appear after a preposition, in this case "Conforme con", preferred over "Conforme a".
cc: @HadrienGardeur


danielweck:
Follow-up PR https://github.com/edrlab/thorium-reader/pull/1884

PR #1855, opened 2022-11-01
Traditional Chinese translation for v2.2

ThanatosDi:


danielweck:
follow-up PR https://github.com/edrlab/thorium-reader/pull/1888

PR #1856, opened 2022-11-02
fix: Dialog modal

### LINKED ISSUE #1635: keyboard characters y and n for yes/no cannot be used to eg. delete a book form My Books?

panaC:
fixes #1635


panaC:
I added a new localisation key 'dialog.cancel' perhaps I should do re-used an old 'cancel' key instead

danielweck:
bug:  keyboard activation of LCP hint hyperlink is broken (enter key only, not space unlike buttons).
I am working on a fix.

danielweck:
bug: the "secondary" button style puts incorrect visual emphasis on button that is actually not the form / enter key default.
I am pushing a fix shortly.

danielweck:
bug: the "API APP" dialog is completely broken with long result list, and the initial enter key form submit causes a crash in the console due to prematurely submitting via the dialog footer.
I am fixing this.

danielweck:
bug: enter key submit bypasses sanity check of existing text in input fields (I am fixing this)

danielweck:
bug: the opds feed editor allows empty text input field in one or the other (I am fixing this)

PR #1858, opened 2022-11-04
 Add Japanese translation of version 2.2 and fix previous translation

peaceroad:
I added the Japanese translation of version 2.2 and fixed previous translation.
(Thanks for mentioning me.)

// I'm not aware of any readers that support EPUB accessibility meta information.
// With Thorium's support, I expect meta-information to be used over time.


danielweck:
follow up PR https://github.com/edrlab/thorium-reader/pull/1886

PR #1866, opened 2022-11-10
Updated Simplified Chinese translation

cary-rowen:
Sorry for being a little late, I updated with the new translatable strings for Simplified Chinese.


danielweck:
Follow-up PR: https://github.com/edrlab/thorium-reader/pull/1882

PR #1999, opened 2023-09-18
[DEV] REACT18

panaC:
remove 'react-beautiful-dnd' seems not used at all

update react et react-dom to latest version 18.2.0

update react-axe to new version

fixed typing error in react18 due to lack of children type in props by default


danielweck (changes requested):
I think that the scope of the "React18" PR (see original https://github.com/edrlab/thorium-reader/pull/1693 ) should strictly be constrained to the changes necessary for React upgrade. I see that "react-redux" is missing here (updated in the original PR). I also see that Deque Axe is changed unnecessarily, and the DND package removed too. These changes are unrelated to React18.
Rationale: the Thorium forks might not be ready for React18 upgrade, which is why the small original PR was laser-focussed on identifying the code gaps, and potential pitfalls for Thorium forks. Maybe now is the right time to force an upgrade, let's discuss.
But I would like to understand the motivation for a separate React18 PR (additional code changes since commits to the `develop` branch? ... if so, let's align the original PR with `develop`)

panaC:
I'm removing this branch, Daniel original branch merged.

PR #2002, opened 2023-09-21
[DEV] collection of useful react hooks

panaC:


src/renderer/common/hooks/useKeyboardShortcut.ts
```diff
@@ -0,0 +1,23 @@
...
+import { useSelector } from "./useSelector";
+import { ICommonRootState } from "readium-desktop/common/redux/states/commonRootState";
+
+export function useKeyboardShortcut(ListenForKeyUP: boolean, keyboardShortcut: (s: ICommonRootState["keyboard"]["shortcuts"]) => TKeyboardShortcutReadOnly, callback: () => void) {
+
+    const keyboardShortcutState = useSelector((state: ICommonRootState) => state.keyboard.shortcuts);
+    React.useEffect(() => {
+        registerKeyboardListener(ListenForKeyUP, keyboardShortcut(keyboardShortcutState), callback);
+        return () => unregisterKeyboardListener(callback);
+    }, [keyboardShortcutState]);
```
danielweck, line 20: I haven't tested this hook placeholder yet, but I think this will cause unnecessary renders due to the "state" object inside the dependency array being compared via reference identity. See `keyboardShortcutsMatch()` calls in `componentDidUpdate()` class component lifecycle.
I also don't see the `ensureKeyboardListenerIsInstalled()` call equivalent to the `componentDidMount()` class component lifecycle (`React.useEffect()` with _empty dependency array_)
@panaC did you test this hook, or is this work in progress? (the PR is marked as "[DEV]")
Thank you :)

danielweck:
What's the difference with https://github.com/edrlab/thorium-reader/pull/1997 ?

PR #2007, opened 2023-09-25
SASS implementation: base branch of many Dialog refactor branches

arthur-lemeur:


panaC:
This branch is the base of one or (several in future) dialog refactor branches. This PR need to be carefully tested before merge to develop.

panaC:
title of the PR 7/11 : `[DEV][WIP]test implementation Sass sur Settings.tsx ` now this PR is the base branch of other refactor branches, the title need to be changed.

panaC:
I removed sass exec and move sass-loader and node-sass to dev-deps

panaC:
I see that the PR was initially used to test the new sass functionality in the Settings component. Now it's the base branch with sass implementation of other Dialog branches. We need to keep this PR with the minimal change from the develop, just the base sass feature to work with other thorium branches. So I think just keep global css configuration like dark/light theme ?

src/renderer/assets/styles/partials/_screenSizes.scss
danielweck, line 1: Hello, I am wondering what the added value of SASS is given that PostCSS already provides all the required functionality? For example, the "variables" in this file could easily be supported by adding a plugin to the existing PostCSS toolchain, instead of inserting a heavy SASS processor? https://github.com/postcss/postcss-simple-vars
Maybe the other pending PRs make use of complex SASS features that are not available in PostCSS already?
If so, I totally understand that SASS is the way forward :)
Thank you for educating me!
(sorry if I missed something obvious)

arthur-lemeur: In my opinion, Sass offers a non negligible comfort in writing and reading. The usage of variables, as well as the nesting syntax with Sass seems much more intuitive and simple compared to PostCSS, with less code used as well. Moreover, Sass is better known than PostCSS and so is easier to maintain by any front dev. And I think that the fact that Sass has all its functionalities out of the box and you don't need to install a plugin for each functionality is an advantage too. As for this file in particular, it is not used at the moment, I created it during my tests of implementation of Sass but I might not need it at all for now it's there just in case :)

danielweck: Thank you for explaining :)

Note that PostCSS nesting conforms to the CSS specification (admittedly, a draft at this moment in time), whereas SASS does not:
https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-nesting
...however, I completely understand that you (and other devs) may wish to write CSS with your preferred / more-familiar SASS nesting syntax instead, so I respect your choice.
PS:- "nested" is different from "nesting":
https://github.com/edrlab/thorium-reader/blob/develop/postcss.config.js#L18
https://github.com/postcss/postcss-nested

Personally I consider the PostCSS plugins ecosystem as an advantage (for example we can easily add support for SASS-like variables, just by adding the appropriate official PostCSS plugin in the chain) ... but again I do not want to interfere with the wishes of the CSS developers (right now, yourself), so I am happy to introduce SASS in the build system as long as we are fully aware of the redundancy / overlap that this will create with the existing PostCSS toolchain.

PR #2009, opened 2023-09-26
remove switch case in ToastManager return

arthur-lemeur:


src/renderer/common/components/toast/ToastManager.tsx
```diff
@@ -62,15 +63,22 @@ export class ToastManager extends React.Component<IProps, IState> {
...
+                                close={() => this.close(id)}
+                                type={toast.type}
+                                displaySystemNotification={false}
+                                />
+                            );
+                            default:
+                            return <></>;
+                        }
                     }
                     return undefined;
```
panaC, line 83: You should not return `undefined` but an empty jsx tag

PR #2014, opened 2023-09-28
[FEAT] Menu component with Radix-ui popover element

panaC:
cf https://github.com/edrlab/thorium-reader/pull/2013

the return focus function on the menu button (doBackFocusMenuButton) has been removed to keep a small PR.
I plan to add the back focus feature from Radix-ui popover/modal and nest a modal in a popover element to keep focus memory implemented in Radix-ui


panaC:
replace by #2016

PR #2016, opened 2023-09-29
[WIP] PublicationMenu with PublicationInfoDialog + DeletePublicationAlertDialog | publicationOpdsInfo | publicationInfoReader

panaC:


panaC:
new menu with focus lock and Radix&floating-UI : 

image

PublicationInfo with Dialog Radix-UI : 
image

Delete Altert Modal with Radix : 

image

The ESC chain work.

Same thing on Opds section :

image

Need to fix the css dimensions of the publication Modal

src/renderer/assets/styles/reader-app.css
```diff
@@ -86,7 +86,7 @@ body {
   font: 1.1rem "Open Sans", sans-serif;
 }
 
-button {
+.menu_option button {
```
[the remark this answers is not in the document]
danielweck: Reverted: https://github.com/edrlab/thorium-reader/pull/2016/commits/9d620ef4acf2db6e0ba7439f186ed671cbe9f818

danielweck:
TODO:
* reader window, CSS buttons `height` selector / cascade override
* OPDS info dialog: tab stops correct, but not cyclic (unlike non-OPDS pub info dialog)

package.json
```diff
@@ -251,6 +251,10 @@
     }
   },
   "dependencies": {
+    "@radix-ui/colors": "^3.0.0-rc.5",
+    "@radix-ui/react-alert-dialog": "^1.0.5",
```
panaC, line 255: @arthur-lemeur should we keep the radix-colors or should we add our own color scheme ? 

I think we must have a global color css file ? What do you think ?

panaC:
Status: Almost Ready to merge (tested with screen reader and approved in call) .. need to check the css color of the alertDialog (deletePublication)

src/renderer/library/components/dialog/publicationInfos/catalogControls.tsx
```diff
@@ -53,26 +55,26 @@ export class CatalogControls extends React.Component<IProps, undefined> {
...
+                    <SVG svg={ReadBook} ariaHidden />
                     {__("catalog.readBook")}
                 </button>
-                <DeletePublicationConfirm
-                    trigger={(
-                        <button className={stylesButtons.button_transparency}>
-                            <SVG svg={DeleteIcon} ariaHidden />
-                            {__("catalog.deleteBook")}
-                        </button>
+                <button onClick={this.deletePublication} className={stylesButtons.button_transparency}>
```
panaC, line 61: Why delete here the new radix alert component :  `DeletePublicationConfirm` ?

src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
```diff
@@ -107,225 +108,225 @@ const Progression = (props: {
...
-            txtProgression = `${percent}% [${formatTime(Math.round(locatorExt.audioPlaybackInfo.globalTime))} / ${formatTime(Math.round(locatorExt.audioPlaybackInfo.globalDuration))}]`;
-        } else if (isDivina) {
-            // console.log("----- ".repeat(100), divinaNumberOfPages, r2Publication?.Spine?.length);
-            let totalPages = (divinaNumberOfPages && !divinaContinousEqualTrue) ? divinaNumberOfPages : (r2Publication?.Spine?.length ? r2Publication.Spine.length : undefined);
-            if (typeof totalPages === "string") {
-                try {
-                    totalPages = parseInt(totalPages, 10);
-                } catch (_e) {
-                    totalPages = 0;
+            const isAudio = locatorExt.audioPlaybackInfo
```
panaC, line 111: Why change a lot of code by adding a tabulation if the linting already pass and you do not modify these lines ? 

The diff become too heavy, difficult to check change

src/common/models/dialog.ts
```diff
@@ -56,6 +57,9 @@ export interface DialogType {
         feed: IOpdsFeedView;
     };
     [DialogTypeName.ApiappAddForm]: {};
+    [DialogTypeName.DeletePublicationConfirm]: {
```
panaC, line 60: same why re-added the previous modal definition for deletePublicationConfirm

src/renderer/assets/icons/SaveAs-icon.svg
panaC, line 1: These icons are really really small, perhaps is in my 4k screen but, I'm wonder how does it render on a big 720p screen. We need to check on a pretty old screen. What do you think ?

src/renderer/library/components/dialog/publicationInfos/catalogControls.tsx
arthur-lemeur: Yes, sorry it was a mistake, I hadn't realised that.

src/common/models/dialog.ts
arthur-lemeur: Same reason, error from my part. I copied some things from the other branch and those shouldn't have been.

src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
arthur-lemeur: IDE automatic linter is to blame on that one I think

src/renderer/assets/styles/components/buttons.css
```diff
@@ -297,6 +285,62 @@ button {
...
+.button_primary_blue {
+  border: 1px solid var(--color-button-primary);
+  height: 30px;
+  border-radius: 4px;
+  padding: 5px;
+  min-width: 100px;
+  background-color: var(--color-button-primary);
+  color: var(--color-secondary);
+
+  & svg {
```
panaC, line 297: linting error here
```
src/renderer/assets/styles/components/buttons.css
 309:3  ✖  Unexpected duplicate selector ".button_primary_blue svg",      no-duplicate-selectors
           first used at line 297
```

panaC:
image
image
image


a few comments : 
- There are no hover on primary button and the outline is in the same color than the button
- Add tag button is in primary color instead of secondary
- FYI the a11y info, they do not respect the official bullet design
- SVG icons are not reliable between publication info ! Why ?
- the size font is not respected, but it's a dev choice , we need to transmit the current font-size to Mylène.

panaC:
@danielweck @llemeurfr What is your choice for Publication Info Modal size ? 

For example : 
image
image
image
image

The modal change the size in fonction of his content with a min/max size height. 
Do you think must be fixed or not ?

panaC:
last but not least :) 
image

we really need to change dark mode color, but perhaps we need to edit color dark mode directly in develop branch.

panaC:
image


nitpicking many details need to be checked yet :) 
bold title / bullet point / lcp columns / margins

src/renderer/common/components/dialog/publicationInfos/publicationInfoA11y.tsx
```diff
@@ -207,33 +209,33 @@ export class PublicationInfoA11y extends React.Component<IProps, IState> {
...
                     </ul>
-                </details>
-            </div>
-            : <></>}
-        </>
-        : <p>{__("publication.accessibility.noA11y")}</p>;
+                    : <></>}
+                {(a11y_certifiedBy || a11y_certifierCredential || AccessibilityFeature || AccessibilityConformsTo || AccessibilityConformanceReport || AccessibilitySummary) ?
+                    <div className={stylePublication.accessibility_infos_right}>
+                        {/* <details> */}
```
panaC, line 224: @danielweck Just a question : there are no accessibility issue to remove details and summary html elements ?

danielweck: The HTML summary/detail elements provide built-in interaction behaviour and semantics, compared with div, and should be easier to style too. Better avoid generic HTML containers like div when there is a semantically-stronger alternative in the web standard.

panaC:
some points to check : 

- modal width size is not constant until window size is less than its value 
image
image

- we need to update the font size in the figma model to be related to the thorium implementation 

- some margins seems to tight like the distance between tag label and tag items
image

- columns implemented for a11y but not for lcp 

image

- the width of the reader window too small compared to the size of the modal (broken also on the dev branch) 
image
image

- spine items in readerWIndow need some improvement to fit figma design (arrow and margins)
image
image

seems very cool, we are really close to being able to merge into develop :)

panaC:
@arthur-lemeur @llemeurfr @danielweck The question arises whether we should not increase the min width and height of a window 
https://github.com/edrlab/thorium-reader/blob/9dd2e12b3d3f9a675fc77cc2ee42bf5256b56720/src/main/redux/sagas/win/browserWindow/createReaderWindow.ts#L35-L36

https://github.com/edrlab/thorium-reader/blob/9dd2e12b3d3f9a675fc77cc2ee42bf5256b56720/src/main/redux/sagas/win/browserWindow/createLibraryWindow.ts#L46-L47

panaC:
i'm closing it, merged to `ui` master ui branch

PR #2032, opened 2023-10-26
[WIP] Dialog setting

panaC:


src/renderer/library/components/settings/LanguageSettings.tsx
```diff
@@ -29,43 +37,79 @@ interface IBaseProps extends TranslatorProps {
...
-                                    { AvailableLanguages[lang] }
-                                </label>
-                            </div>,
-                        )}
-                    </form>
-                </section>
-            </>
-        );
-    }
+interface AvailableLanguages {
```
panaC, line 57: duplicate from https://github.com/edrlab/thorium-reader/blob/1f0decaa938b5cdf621da84804fdda193aed6bac/src/common/services/translator.ts#L159

src/renderer/library/components/settings/LanguageSettings.tsx
```diff
@@ -29,43 +37,79 @@ interface IBaseProps extends TranslatorProps {
...
 // ReturnType<typeof mapDispatchToProps>
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-interface IProps extends IBaseProps, ReturnType<typeof mapStateToProps>, ReturnType<typeof mapDispatchToProps> {
-}
+// interface IProps extends IBaseProps, ReturnType<typeof mapStateToProps>, ReturnType<typeof mapDispatchToProps> {
+// }
+
+const SelectItem = React.forwardRef(({ children, ...props }: any, forwardedRef) => {
+    return (
+        <Select.Item {...props} id="itemInner" ref={forwardedRef}
```
panaC, line 45: You don't need to forwardRef a sub-component at the top of the tree

src/renderer/library/components/settings/Settings.tsx
```diff
@@ -36,10 +43,48 @@ class Settings extends React.Component<IProps, undefined> {
                 <LibraryLayout
                     title={__("header.settings")}
                 >
```
panaC, line 45: This component is loaded from react-router https://github.com/edrlab/thorium-reader/blob/1f0decaa938b5cdf621da84804fdda193aed6bac/src/renderer/library/routing.ts#L62-L66

I think that the react-router route `/setting` should be disable (the setting component is no longer a page but a modal dialog now). And Setting Dialog must be move to the `common/components/dialog` folder (used in library and reader as well)

danielweck: > I think that the react-router route `/setting` should be disable (the setting component is no longer a page but a modal dialog now).

Nested routes (i.e. relative `settings_modal` path inside another route definition) would work well declaratively, for example `/bookshelf` + nested `/bookshelf/settings_modal` ... but the settings modal popup dialog with Radix is really designed to be displayed anywhere (overlaid on top of an existing route component) where a "call site" button (trigger UI control) can be placed, so I agree that using the router to implement this is a waste of time and effort (unless we really want to use router functionality, for example to open the settings dialog in a specific tab by passing route parameters?)

src/renderer/library/components/settings/ReadingOptions/TextContent.tsx
```diff
@@ -0,0 +1,130 @@
+import fontList, { FONT_ID_DEFAULT, FONT_ID_VOID } from "readium-desktop/utils/fontList";
+import * as stylesReader from "readium-desktop/renderer/assets/styles/reader-app.css";
+
+const textContent= () => {
+    const {__, readerConfig} = this.props;
```
panaC, line 5: - The first letter of the component must be in the upperCase.
- You cannot used `this.props` from a react functional component, `props` must be passed from function parameter and not from function binding `this` this keyword is used in class components instead.

panaC:
replaced by #2039

PR #2082, opened 2024-03-25
Update pt-pt locale

horus68:
Just keeping up with current develop.
Small update to pt-pt locale.


danielweck:
Thank you very much!!

PR #2088, opened 2024-03-28
[Ready to merge] put a book in a "set as finished" mode

panaC:
put a book in a "set as finished state".

From the UI : it is Implemented as a button in the bottom right of the readerFooter when the book is close to finish. 
And a "Readed" icon in the allPublicationVIew (aka Tableau), there are also a button action named "set as finished" in the publication card menu.

This feature use a Priority Queue reducer to pop out id from the "lastReading" queue and push it to the "readingFinished" queue (the new pqueue). When a book is re-read this is the reverse and pop from "readingFInished" and push it to the "readingFinished".


danielweck:
Hello, is this PR still valid or can it be closed?

panaC:
This PR is it ready for a review @arthur-lemeur  ?

Can you add Daniel as reviewer and change the start of the title with "ready to merge" instead of "WIP". 

Thanks :)

src/renderer/library/components/searchResult/AllPublicationPage.tsx
```diff
@@ -167,6 +168,7 @@ export class AllPublicationPage extends React.Component<IProps, IState> {
         ], () => {
             apiAction("publication/findAll")
                 .then((publicationViews) => {
+                    console.log(publicationViews);
```
danielweck, line 171: TODO: remove this console log?

src/renderer/library/components/searchResult/AllPublicationPage.tsx
```diff
@@ -1478,6 +1519,9 @@ export const TableView: React.FC<ITableCellProps_TableView & ITableCellProps_Com
                     label: publicationView.authors ? publicationView.authors.join(", ") : "",
                     authors: publicationView.authors,
                 },
+                colReadingState: { // IColumnValue_Authors
+                    label: publicationView.readingFinished ? `${props.__("publication.read")}` : publicationView.lastReadingLocation ? `${props.__("publication.onGoing")}` : "",
+                },
```
danielweck, line 1524: See https://github.com/edrlab/thorium-reader/issues/2110
I propose "Finished" (instead of "Read" which is both a verb and an adjective so can be confused for an action indicator / affordance instead of a state)
I propose "In Progress" instead of "On Going" (which is incorrect anyway, should be "Ongoing")
I propose "Not Started" instead of empty string so that the column cells are sorted and batched together when the user looks for publications that are not opened yet.

src/renderer/reader/components/ReaderFooter.tsx
```diff
@@ -376,10 +378,10 @@ export class ReaderFooter extends React.Component<IProps, IState> {
                                 } */}
                             </div>
                         }
-                        {/* {isEnding ? <button className={stylesReaderFooter.finishedIcon}
-                             onClick={() => this.props.finishReading()}
+                        {isEnding ? <button className={stylesReaderFooter.finishedIcon}
+                             onClick={() => this.props.finishReading(this.props.publicationView.identifier)}
                              title="Mark as finished"
```
danielweck, line 383: TODO: translate label

danielweck (changes requested):

PR #2105, opened 2024-04-25
Dev/UI arthur update

arthur-lemeur:
minor updates of ui (footer & header in library, responsive) + update tts options location


danielweck (approved):
Great thank you! Next time if you can please create smaller, more atomic Pull Requests that target one specific issue or feature, for example the change of TTS voice selector in this PR (this will make it easier to review the code diff and run tests, also makes it easier to rollback / revert if necessary)

PR #2117, opened 2024-05-02
set readerMenu/Settings dialog as modal in undocked mode

panaC:


danielweck:
As the centered non-docked dialog is now a proper modal, I think we should also make sure that `onPointerDownOutside` and `onInteractOutside` behave accordingly:

```
onPointerDownOutside={(e) => { e.preventDefault(); console.log("settingsModal onPointerDownOutside"); }}
onInteractOutside={(e) => { e.preventDefault(); console.log("SettingsModal onInteractOutside"); }}
```

danielweck:
https://github.com/edrlab/thorium-reader/blob/0a430ace83d339d6a11789979b0dfa86ab690b9a/src/renderer/reader/components/ReaderHeader.tsx#L968-L972

https://github.com/edrlab/thorium-reader/blob/0a430ace83d339d6a11789979b0dfa86ab690b9a/src/renderer/reader/components/ReaderHeader.tsx#L1010-L1014

PR #2124, opened 2024-05-03
fix duplicates in voices array causing key errors

arthur-lemeur:
We had an error in the console regarding two elements having the same key (with the voice fr-FR Thomas which was twice in _orderedVoices). I added a function to check for duplicates to avoid that.


danielweck:
Is the React error/warning message related to `key={`tts${item.id + 1}`}`?

danielweck:
Side note: the _orderVoices `reduce()` seems completely unnecessary now? (just use the result of `sort()` directly?

arthur-lemeur:
I have to admit I don't know why Thomas appeared twice in the array. But yes the error message what related to `key={`tts${item.id + 1}`}` even if in the array the two entries had different IDs. To fix this I thought of checking for duplicates before creating the list but maybe this is just a workaround. I am not familiar enough with SpeechSynthesisVoice to be sure.

danielweck:
Thanks @arthur-lemeur I am going to revise your code as I wrote the original and I need to be more familiar with these changes. I will take over this small PR if that's ok with you, so you can focus on more important CSS stuff :)

danielweck:
fixed via https://github.com/edrlab/thorium-reader/commit/657c5a83e5cf7cb9e763d029834ec7a2e824b6a6

PR #2136, opened 2024-05-06
fix actions header button to h4 not clickable in allBooks tableView

arthur-lemeur:


danielweck:
ready to merge? looks good to me :)

PR #2146, opened 2024-05-06
fix(2144): setLocator is called once by the navigator at startup and

### LINKED ISSUE #2144: readingFinished refresh flag consume too much cpu load on reader and library
https://github.com/edrlab/thorium-reader/commit/08b2c3c49ae976b811b805238c7bb020f6325208#r141722337

panaC:
then at each new position in the publication from the user action or tts

so we have just to trigger readingFinishedRefresh once and not at each new position. So a counter to track the second call is perfect.

Just an issue if the user click on "set as finished" on library with the book open on parallel and navigate in the document after the click. The update event that will be not propagated and not refresh on library side.

The issue is the dupplication of the state needed by apiSubscribe to fetch allPublication by apiRequest in allPublicationPage.tsx

so this is a little big hacky just for sync a readingFinished across the app.

we need to rely on the redux/react synchronisation instead of reinventing the wheel with an api on top of redux

fixes #2144


panaC:
see comment of Daniel : https://github.com/edrlab/thorium-reader/issues/2144#issuecomment-2096486428

panaC:
I'm wondering if we really need a throttle logic to refresh readingFinished, just to absolutely keep synced readingFinished. It will consume a lot of CPU cycle if TTS is enabled.
And even if we put a throttle of 10s we need to release it on unmount.

danielweck:
> It will consume a lot of CPU cycle if TTS is enabled.

not just TTS, also EPUB3 Media Overlays with word-level synchronisation, and I guess audiobooks too (although there is already rate limiting from the navigator, if I remember correctly)

panaC:
I'm merging this PR, we have to find a better solution to do "readingFinished".

It's ok @danielweck ?

danielweck:
ok :) I trust your judgement (you understand this particular React lifecycle / data flow better than I do!)
we sure need to avoid flooding a large React components tree with updates, and the JSON / persistent state too.

Issue #2140, opened 2024-05-06
Annoying "do you want to save the session" user prompt ... what does it do, why as sometimes, not all the time??


danielweck:
<img width="437" alt="Screenshot 2024-05-06 at 11 45 53" src="https://github.com/edrlab/thorium-reader/assets/625201/dc15523b-b217-4c91-a047-972bb95fd18a">

I do not understand this feature, so I suspect non-developer users will struggle too.
This seems to cause unnecessary friction in the experience as it forces users to make an obscure decision at the point of closing the application.
This feels just like the annoying "cookie banners" on the web ... in the end I just hit yes or no, it doesn't matter as long as the dialog closes :)

danielweck:
@panaC I assigned you to this one as I believe you are the "session" expert ;)
If I remember correctly, in Thorium3 each publication maintains its own preferences, right? (colour theme, font size etc.)
in Thorium2 this was an opt-in feature which was already poorly understood ("activate session" in top-level library preferences + "save session" / "reset session" in reader menus)

panaC:
I keep the session mode previously developed in Thorium, each reader window is independent and rely to the same global readerConfigState.

panaC:
The steps : 

1/ add a new publication and open it, borrow the global readerConfigState
2/ set/update the readerConfig -> update the readerConfigState of the publication but **also** update the global readerConfigState
3/ now your publication has "forked" the readerConfigState and do not rely on it if you update it in an another publication
4/ update an another publication and do the change for your publication only
5/ add a new publication in Library and open it, readerConfigState will be to the state of any publication last changed.
6/ you have more than one publication opened and the library window so 3 windows, you want to close Thorium and close the libraryWin, Thorium will ask you if you "do you want to save the session" in a message box, 2 solutions :
- accept and click on yes : On next Thorium open, all 3 windows will be restored and your publications in your last state before the closing
- do not accept and click on no : so Thorium will close and save your 2 reader windows the state is saved and on the next open only library window will be open. Your publication state is saved.

danielweck:
> 6/ you have more than one publication opened and the library window so 3 windows, you want to close Thorium and close the libraryWin, Thorium will ask you if you "do you want to save the session" in a message box, 2 solutions :
> 
>     * accept and click on yes : On next Thorium open, all 3 windows will be restored and your publications in your last state before the closing
> 
[...]

PR #2157, opened 2024-05-07
fix(#2140): keep session message box at closing not preserved, annoying

panaC:


danielweck:
Hello Pierre, I see your code comment that accompanies the code removal:

> // keep session in next startup disabled for the release of Thorium 3.0
> // session still exits between publication state but not saved at closing

Could you please clarify (sorry, I am confused, my fault not yours) what is a "session"? Does this mean that individual publications retain their own user preferences (e.g. font size, paginate vs. scroll etc.) during the application lifecycle, and Thorium automatically makes these settings persistent in the DB/JSON so that when the application is relaunched, all the publications keep their own preferences? What happens to a new publication that was never opened before?

panaC:
> Hello Pierre, I see your code comment that accompanies the code removal:
> 
> > // keep session in next startup disabled for the release of Thorium 3.0
> > // session still exits between publication state but not saved at closing
> 
> Could you please clarify (sorry, I am confused, my fault not yours) what is a "session"? Does this mean that individual publications retain their own user preferences (e.g. font size, paginate vs. scroll etc.) during the application lifecycle, and Thorium automatically makes these settings persistent in the DB/JSON so that when the application is relaunched, all the publications keep their own preferences? What happens to a new publication that was never opened before?

I reply on the issue : https://github.com/edrlab/thorium-reader/issues/2140#issuecomment-2110222367

danielweck:
@panaC So, instead of `value = messageValue.response;` could we pluck the boolean from global persistent application state (user-configurable preference, by default true = reader windows are restored automatically). No annoying confirmation dialog, the UX friction becomes minimal because users will just close their reader windows if they are annoyed by their re-appearance at next launch, or they will find the configuration checkbox or ask on the mailing list / GitHub :) (nobody reads the docs, sorry @gautierchomel ;)

panaC:
I'm closing, replace by #2307

PR #2158, opened 2024-05-07
Software update version check

danielweck:
In this PR branch, the version bump is from alpha to production-ready 3.0.0 (just for testing, will downgrade to alpha before merging).  The popup modal dialog only appears at app startup in dev mode, production builds will only see the more discrete pop out toast at the bottom of the home page (ARIA live "polite" alert). The yes/no answer (dismissal) is not persistent across React components, so the message toast reappears when coming back to the homepage. TODO: maybe make persistent until the app process restarts?  PS: note that Windows Microsoft App Store is excluded, the version is not checked at all.

@llemeurfr could you please check that the UX is adequate? (maybe Arthur/Pierre can demo the build from this branch for you?)

@panaC please check my Redux Saga code, feel free to improve if necessary. Also have a think about making the dismissal persistent during the application lifecycle, until restart.

@arthur-lemeur please check the CSS styles of the notification toast.

Thank you :)


src/renderer/library/components/catalog/GridView.tsx
```diff
@@ -147,17 +148,24 @@ class CatalogGridView extends React.Component<IProps, IState> {
                     )
                 }
                 {
-                    this.state.versionInfo ?
-                    <div className={stylesGlobal.new_version}>
+                    this.state.versionInfo && this.props.newVersionURL && this.props.newVersion ?
```
panaC, line 151: Why this is in GridView Component ?

src/main/redux/sagas/index.ts
```diff
@@ -119,6 +130,105 @@ export function* rootSaga() {
...
+
+                    if (IS_DEV) {
+                        yield call(async () => {
+
+                            const translate = diMainGet("translator").translate;
+                            const res = await dialog.showMessageBox(// browserWindow,
+                                {
+                                type: "question",
+                                buttons: [
+                                    translate("app.session.exit.askBox.button.yes"),
```
panaC, line 212: TODO: change i18n key

panaC:
https://www.electronjs.org/fr/docs/latest/api/process#processwindowsstore-readonly :) clever !

src/main/redux/sagas/index.ts
danielweck: or alternatively, hoist the yes/no generic translated labels to top-level, no need to create "duplicate" entries ;)

src/renderer/library/components/catalog/GridView.tsx
danielweck: not sure, I just reused the existing code. this PR is a good opportunity to move things around if it makes more sense architecturally

src/main/redux/sagas/index.ts
panaC: on est bien d'accord

danielweck: we need a "code mod" script to move this JSON key for all locales

src/renderer/library/components/catalog/GridView.tsx
danielweck: Maybe move this "toast" -looking component into the bottom footer info area that already contains the version number. In the Zed code editor, a text link appears in the bottom bar without yes/no buttons, I can just choose to ignore it, or click on the link and download/activate the new version. I like this UX pattern because it means we are strongly encouraging users to update, but we are not nagging them with a toast notification that obscures the UI underneath. @llemeurfr what do you think?

danielweck: Screenshot 2024-05-10 at 11 08 07

PR #2159, opened 2024-05-07
Fixes some design issue

arthur-lemeur:
New colour variables added and update colours through the app to match the Figma design. 
Update the NoPublicationInfo component.


danielweck:
I confirm that this is now fixed too: https://github.com/edrlab/thorium-reader/issues/2151

panaC:
We need to double check some css design (dropdown menu, primary secondary button, ...) before to merge

Issue #2154, opened 2024-05-06
cross icon inside text input fields has shaded design but rest of Thorium GUI uses flat style


danielweck:
<img width="562" alt="Screenshot 2024-05-06 at 20 19 53" src="https://github.com/edrlab/thorium-reader/assets/625201/029afc77-932b-4715-965e-27002740f13f">

danielweck:
Will be fixed by https://github.com/edrlab/thorium-reader/pull/2161

PR #2161, opened 2024-05-10
fix input search reset button (issue #2154)

arthur-lemeur:
Change from default reset icon in input type search to custom button to allow customable colours


src/renderer/library/components/dialog/ApiappAddForm.tsx
```diff
@@ -120,13 +122,26 @@ const ApiappAddForm = () => {
...
                                 // e.stopPropagation();
                             }
                         }}
                     />
+                    {searchInputRef ?
+                        <button
+                            className={stylesGlobal.button_clear}
+                            type="reset"
+                            onClick={() => {
+                                searchInputRef.current.value = "";
```
danielweck, line 137: Unlikely to occur in the React lifecycle, but I would still recommend to satisfy the TypeScript compiler in a project with stricter null/undefined checks:
`searchInputRef.current?.value = ...`

src/renderer/library/components/dialog/ApiappAddForm.tsx
```diff
@@ -120,13 +122,26 @@ const ApiappAddForm = () => {
...
                                 const v = searchInputRef.current?.value;
-                                if (!v) return ;
+                                if (!v) return;
                                 apiAppSearchAction(v);
                                 // e.preventDefault();
                                 // e.stopPropagation();
                             }
                         }}
                     />
+                    {searchInputRef ?
```
danielweck, line 132: The ref itself is always truthy ... shouldn't this be:
`{ searchInputRef.current ?`

src/renderer/library/components/searchResult/AllPublicationPage.tsx
```diff
@@ -428,6 +430,17 @@ const CellGlobalFilter: React.FC<ITableCellProps_GlobalFilter> = (props) => {
                 }}
                 placeholder={`${props.__("header.searchTitle")}`}
             />
+            {props.focusInputRef?.current?.value ?
+            <button
```
danielweck, line 434: no `type="reset"` on this one?

src/renderer/reader/components/picker/SearchFormPicker.tsx
```diff
@@ -118,6 +120,19 @@ class SearchFormPicker extends React.Component<IProps, IState> {
                             <SVG ariaHidden={true} svg={searchIcon} />
                     }
                 </button>
+                {this.state.inputValue ?
+                <button
+                    className={stylesGlobal.button_clear}
+                    type="reset"
+                    onClick={() => {
+                        this.inputRef.current.value = "";
```
danielweck, line 128: Code guard: `this.inputRef.current?.value =`

src/renderer/reader/components/picker/SearchFormPicker.tsx
```diff
@@ -118,6 +120,19 @@ class SearchFormPicker extends React.Component<IProps, IState> {
...
+                        console.log(this.inputRef.current.value);
```
danielweck, line 129: remove logging. but also: does setting the DOM value directly using a ref trigger `onChange={(e: React.ChangeEvent<HTMLInputElement>) => this.setState({ inputValue: e.target.value })}`? if not, we are breaking the React lifecycle of controlled editable components. need to double check...

src/renderer/library/components/opds/SearchForm.tsx
```diff
@@ -95,11 +98,23 @@ class SearchForm extends React.Component<IProps, undefined> {
...
+                title={__("header.searchTitle")}
+            >
+                <SVG ariaHidden={true} svg={SearchIcon} />
+            </button> */}
+                {this.props.search?.url ?
+                <button
+                    className={stylesGlobal.button_clear}
+                    type="reset"
+                    onClick={() => {
+                        this.inputRef.current.value = null;
```
danielweck, line 111: Code safety and happy compiler: `this.inputRef.current?.value =`
...and I think in this case writing directly into the DOM is okay as there is no change event handler or React state management attached to the input field.

danielweck (changes requested):

src/renderer/library/components/dialog/ApiappAddForm.tsx
arthur-lemeur: TypeScript does not allow this: "La partie gauche d'une expression d'assignation ne doit pas être un accès à une propriété facultative.ts(2779)"

src/renderer/reader/components/picker/SearchFormPicker.tsx
arthur-lemeur: TypeScript does not allow this: "La partie gauche d'une expression d'assignation ne doit pas être un accès à une propriété facultative.ts(2779)"

src/renderer/library/components/opds/SearchForm.tsx
arthur-lemeur: TypeScript does not allow this: "La partie gauche d'une expression d'assignation ne doit pas être un accès à une propriété facultative.ts(2779)"

src/renderer/library/components/dialog/ApiappAddForm.tsx
danielweck: Sorry, yes of course, I meant some code guard like:

`searchInputRef.current && searchInputRef.current.value = "";`

or

`if (searchInputRef.current) searchInputRef.current.value = "";`

danielweck:
Will fix https://github.com/edrlab/thorium-reader/issues/2154

danielweck (changes requested):
The logic associated with the cross / clear button does not work. I tested with a without a screen reader. There is a GUI conflict with the screen reader "search" button (I will post a screenshot). Without a screen reader I am able to break the state of the text input field. More testing is needed to identify how to best implement this feature.

danielweck:
Screenshot 2024-05-13 at 16 52 26

panaC:
I'm closing the PR, we found Arthur and I a solution to resolved it in css only. 
https://github.com/edrlab/thorium-reader/commit/62c568d679be3f401de99c5a4281b597e4470fb4

src/renderer/assets/styles/global.scss
```diff
@@ -41,6 +41,23 @@ body {
 
 input {
   font-family: Nunito, sans-serif;
+
+  &[type="search"]::-webkit-search-cancel-button{
+    appearance: none;
```
danielweck, line 46: Isn't it just possible to style the system-level cross instead of eliminating it? Aren't we over-engineering the solution to this problem? This would probably solve the out of sync state problem with controlled components too (because the system-level cross is equivalent to the user setting the text to empty which is captured by the input's change event, as opposed to manipulating the React state which can cause inconsistencies)

danielweck: ah, great :)

https://github.com/edrlab/thorium-reader/commit/62c568d679be3f401de99c5a4281b597e4470fb4

PR #2194, opened 2024-05-15
fix(GUI): remove horizontal scrollbar in combobox (Fixes #2131)

### LINKED ISSUE #2131: combo box with search, long text in option causes horizontal scroll
low priority, but just in case this can be fixed by adding clipping and ellipsis (+ tooltip / accessible full text)

Screenshot 2024-05-03 at 09 33 41

arthur-lemeur:


danielweck (changes requested):
This looks good but I am adding a tooltip to reveal the full list item string (which is accessible to screen readers even with CSS ellipsis, but truncated visually)

danielweck:
@arthur-lemeur @panaC in my commit https://github.com/edrlab/thorium-reader/pull/2194/commits/3d32930d48732543e3c8fea569d008983d661803 you will see that I replaced TypeScript type coercion with type inference ...well, not quite inference in this case as I still declare the type signatures, more like reusing existing library types instead of hard-coding compatible types that may become incompatible (or incomplete) in future lib releases (here: React ARIA).
I also removed a title tooltip in annotations which was 100% always shorter than the text it was overlaid on! :)
The title tooltip trick with onHover is necessary unfortunately as React ARIA doesn't seem to expose the props to inject DOM attribute ... ref would have worked too, but equally gross so I picked onHover instead as the intent is clearer.
Unfortunately this only works with mouse / pointing devices ... keyboard users without screen reader do not see the tooltip.
It's one of the known issues with ellipsed text.

Issue #2182, opened 2024-05-14
GUI: scroll not dark in opds publication list


panaC:
![image](https://github.com/edrlab/thorium-reader/assets/3198865/adf2e9cb-3530-47f9-a4f4-f5a327271f6d)

PR #2197, opened 2024-05-15
fix(GUI): scrollbar dark mode in opds catalog list (issue #2182)

arthur-lemeur:
Fix issue #2182


danielweck:
@arthur-lemeur as I'm sure you have noticed already, there are several styles that should be factored out for sharing across multiple components, such as scrollbars, dialog modal outline/shadow, etc.
Could these becomes reusable mixins instead of duplicating CSS code?

PR #2209, opened 2024-05-16
fix noPublicationInfo for small screens

### LINKED ISSUE #2213: low contrast text with transparent background in welcome screen (first launch after install)
Screenshot 2024-05-16 at 17 18 49

arthur-lemeur:
Fixes #2213

### DIFF
--- a/src/resources/locales/en.json
+++ b/src/resources/locales/en.json
@@ -81,7 +81,7 @@
         "moreInfo": "Book Details",
         "myBooks": "My books",
         "noPublicationHelpL1": "Drop a publication here",
-        "noPublicationHelpL2": "or use the \"+\" button above.",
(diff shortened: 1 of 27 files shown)


src/resources/locales/en.json
```diff
@@ -81,7 +81,7 @@
         "moreInfo": "Book Details",
         "myBooks": "My books",
         "noPublicationHelpL1": "Drop a publication here",
-        "noPublicationHelpL2": "or use the \"+\" button above.",
+        "noPublicationHelpL2": "or use the \"import publication\" button above.",
```
danielweck, line 84: some thoughts: string interpolation with the existing already-translated label would (1) avoid duplication / risk of divergence during GUI updates, (2) avoid creating a TODO for translators ... in the flip side, this forces us to replace the old "+" with the new interpolated string in all locales.

What do you think?

PR #2235, opened 2024-05-19
refactor scrollbar css into mixin

### LINKED ISSUE #2223: light scrollbar should be dark
I mentioned this in another discussion thread, but it would be better if the common styles of scrollbars, modal outline / drop shadow, etc. were factored-out (mix ins?) in order to deliver a consistent look and feel throughout the application without copy-pasting CSS snippets all over components. I think this is why we are seeing inconsistencies.

Screenshot 2024-05-17 at 18 15 46

arthur-lemeur:
Fixes #2223

### DIFF
--- a/src/renderer/assets/styles/components/allPublicationsPage.scss
+++ b/src/renderer/assets/styles/components/allPublicationsPage.scss
@@ -2,6 +2,9 @@
 =-=-= allPublicationsPage.scss
 */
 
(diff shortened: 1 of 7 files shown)


danielweck (changes requested):
code looks good, but see screenshot

danielweck:
Screenshot 2024-05-20 at 10 48 10

PR #2258, opened 2024-05-20
Fixe focus behavior on button next in wizard

### LINKED ISSUE #2245: wizard: keyboard tabbing into "next" button activates it, causing focus outline on modal perimeter and switch to next tab
Screenshot 2024-05-20 at 09 29 08

arthur-lemeur:
Fixes #2245

### DIFF
--- a/src/renderer/library/components/Wizard.tsx
+++ b/src/renderer/library/components/Wizard.tsx
@@ -63,7 +63,7 @@ export const WizardModal = () => {
 
     const [checked, setChecked] = React.useState(true);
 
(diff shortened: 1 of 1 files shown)


danielweck:
interesting that the focus participates in form submission (which is why, I imagine, the preventDefault is necessary?)
anyway, merging :)

arthur-lemeur:
Capture d’écran 2024-05-20 à 14 07 23

From the radix documentation

As this button is a trigger in a TabList (the only way to put a trigger with this component) and the TabList only contains this trigger, it must be the reason it fires it on focus.. I suppose.

PR #2269, opened 2024-05-20
deprecated onKeyPress is replaced with onKeyUp, onKeyDown is repeatable which is not necessary here so replace with onKeyUp

danielweck:
... also replaced .key === " " with "Space" for consistency with existing code


panaC:
see message : https://github.com/edrlab/thorium-reader/pull/2273#discussion_r1606890364

I can't find the documentation for "Space" `key` value in https://www.w3.org/TR/uievents-key/#named-key-attribute-values

panaC:
see https://www.w3.org/TR/uievents-key/#keys-whitespace

danielweck:
Thank you for your feedback. Tested with and without screen reader (Voice Over on MacOS, all latest versions)

PR #2277, opened 2024-05-20
fix docked popover position and dimmensions when search enable

arthur-lemeur:


danielweck:
tested before and after, works.

PR #2286, opened 2024-05-21
fix focus on title in table mode + outline on table header

arthur-lemeur:


src/renderer/library/components/searchResult/AllPublicationPage.tsx
```diff
@@ -1164,9 +1164,15 @@ const CellTitle: React.FC<ITableCellProps_Column & ITableCellProps_GenericCell &
         dir={pubTitleIsRTL ? "rtl" : undefined}
     >
                 <a
+                    tabIndex={0}
                     className={stylesPublication.cell_bookTitle}
                     title={`${pubTitleStr} (${props.__("catalog.bookInfo")})`}
```
danielweck, line 1169: This string is "About publication" so the associated action does not match (`props.openReader(props.value.publicationViewIdentifier)`).
Solution: remove the `title` attribute

danielweck (changes requested):

PR #2288, opened 2024-05-21
fixes #2270 annotation concurrent edit

### LINKED ISSUE #2270: annotations editor should be unique, no concurrent edits
follow-up to:
https://github.com/edrlab/thorium-reader/issues/2260

note that the keyboard shortcuts editor properly handles mouse-driven actuation of another editor (closes the current editor)

Screenshot 2024-05-20 at 15 02 33

panaC:
In docked mode I cannot apply a focus lock because there are too many mouse possible interaction in the header, focus lock is fighting then

Fixes #2270

### DIFF
--- a/src/main/redux/store/memory.ts
+++ b/src/main/redux/store/memory.ts
@@ -120,7 +120,7 @@ export async function initStore()
         debug("On the first launch runtimeStatePath failed it's an empty file (not created)");
         debug("On the second launch runtimeStatePath is equal to an empty object {}");
         debug("and failed on checkReduxState, reduxState has not be preloaded in runtimeStateFilePath");
-        debug("So the Third launch is good!, Thorium State is stabilize");
+        debug("So the Third launch is good!, Thorium State is stabilize")
         const state = await recoveryReduxState(await runtimeState());
         reduxState = await checkReduxState(state, reduxState);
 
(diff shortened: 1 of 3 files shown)


panaC:
add bookmarkList too

danielweck (changes requested):
something is not right...sometimes when I click on the edit button the focus is redirected to the already opened editor, sometimes a new editor appears where I clicked. I test with or without any pending changes (e.g. inside text input area)

danielweck:
I haven't tested the delete button yet, but we'll need to make sure that the state isn't corrupted when mixing edit/delete interactions.

panaC:
> something is not right...sometimes when I click on the edit button the focus is redirected to the already opened editor, sometimes a new editor appears where I clicked. I test with or without any pending changes (e.g. inside text input area)

can you be more precise on your requesting change : 

On modal mode : focus-lock is applied, so when i click on edit the focus is redirected to textEdit, when i click on an another item, the focus is redirected to this item.
When I click on next/previous/delete(any) the item currently edited is closed

On docked mode: the focus-lock is disable (too many focus-fighting on search/annotation popover) the edition cannot be modal for me (need to discuss about that)
otherwise it's the same / the edited mode is close with next/previous/delete

panaC:
> something is not right...sometimes when I click on the edit button the focus is redirected to the already opened editor, sometimes a new editor appears where I clicked. I test with or without any pending changes (e.g. inside text input area)

Got it , focusLock is broken, so I disabled it in this PR, and I filed an issue.

PR #2291, opened 2024-05-21
fixes #2127 feature "hide" annotations

### LINKED ISSUE #2127: feature request: disable annotations rendering entirely, to remove interference with hyperlinks, media overlays and text to speech etc. (any click interaction)

panaC:
Fixes #2127 

Draft PR

I need your review @danielweck 

I just use a checkbox like a button to temporary force/hard remove highlight in current webview and secondary webview.

Navigation to a next ressource and come back make appear Highlight removed

### DIFF
--- a/src/renderer/reader/components/ReaderMenu.tsx
+++ b/src/renderer/reader/components/ReaderMenu.tsx
@@ -46,7 +46,7 @@ import * as AnnotationIcon from "readium-desktop/renderer/assets/icons/annotatio
 import * as CalendarIcon from "readium-desktop/renderer/assets/icons/calendar-icon.svg";
 // import * as DuplicateIcon from "readium-desktop/renderer/assets/icons/duplicate-icon.svg";
 
-import { LocatorExtended } from "@r2-navigator-js/electron/renderer/index";
+import { highlightsRemove, LocatorExtended } from "@r2-navigator-js/electron/renderer/index";
 import { Link } from "@r2-shared-js/models/publication-link";
 
 import SVG from "readium-desktop/renderer/common/components/SVG";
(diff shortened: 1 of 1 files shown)


danielweck:
Ah, that's not how I imagined the feature implementation :)
I hoped that we could "simply" (nothing is simple!) NOT call the highlights drawing functions (navigator webview APi) while preserving the full application-level state. This would immediately solve the problem of (1) visual distractions of highlights (inline, but also margin indicators which remain intrusive when the user just wants "zen" reading) and (2) interaction interference of "tactile" highlights (again, not just inline but also margin).
The user could still access the list of annotations, even click on them to navigate (no need to draw highlights, as they are non-structural, just painting colours over the text document), and even edit/delete them.
We just need to make it very obvious in the GUI so that the user knows that their current configuration is "hide annotation highlights". I proposed the 3-state "inline", "margin", "hide" but I completely agree with your approach to use a checkbox to simulate. The hard part of this implementation is managing state consistency, then we can improve the UI/UX.

panaC:
What do you think is not to called `push` in `handler.ts` , that's right ? 

https://github.com/edrlab/thorium-reader/blob/dbdc0c749a63191bbd86158e93617666f0b71d2c/src/renderer/reader/redux/sagas/highlight/handler.ts#L33-L50

danielweck:
yes exactly. add a new state that conveys the "do not draw highlights" user choice.

panaC:
just the 3 states is not really a 3 states because inline and hide in coupled to the logic saga but not the "hide" that lay on directly call the api endpoint without highlight state handling.

I have to more thinking on it, but I try to not add an "Énième" boolean state

panaC:
I review the code, I'm thinking on it, I will try on a new PR

PR #2292, opened 2024-05-21
fix responsive in annotations edit style

arthur-lemeur:


danielweck:
I do not know what bug this fixes so i will let the assigned reviewer (@panaC ) comment :)

PR #2294, opened 2024-05-21
change select component to combobox

### LINKED ISSUE #2250: selection/option control causes keyboard focus loss after activation of list item
keyboard tab into the selector, arrow down to pop out and reach item, press spacebar or enter => focus goes to window root (I noticed that other selectors like the font options retain focus)

Screenshot 2024-05-20 at 09 56 45

arthur-lemeur:
Fixes #2250


danielweck:
Screenshot 2024-05-21 at 18 09 18

danielweck (changes requested):
given that I can type into the text input field in order to lookup list items, I would expect this to work. Right now it is quite confusing not to be able to search by list item label. arrow key navigation works though, which is nice. That being said, l mentioned in the issue that RAC demos work fine, so it seems there is something we do in Thorium that breaks the good behaviour of React ARiA Components?

danielweck:
TTS speed select:

Screenshot 2024-05-22 at 13 40 48

danielweck:
the selector GUI control is not keyboard accessible (no tab key landing)

Screenshot 2024-05-22 at 13 47 57

danielweck:
i can click on the selector GUI control with the mouse cursor, but I have to aim specifically for the arrow, if I click on the text, nothing happens. When I unfold the list, I cannot select items with the arrow keys, only the mouse. also note the weird outline styles:
Screenshot 2024-05-22 at 13 47 57
Screenshot 2024-05-22 at 13 48 04
Screenshot 2024-05-22 at 13 49 06

danielweck:
https://github.com/edrlab/thorium-reader/issues/2302

Screenshot 2024-05-21 at 18 08 06
Screenshot 2024-05-21 at 18 08 00

panaC:
https://github.com/edrlab/thorium-reader/issues/2250 fixed with https://github.com/edrlab/thorium-reader/commit/900773169efc769e6de177f7f96db634cf12f27b

PR #2299, opened 2024-05-21
fix vertical scroll on left side wizard + position buttons

### LINKED ISSUE #2290: left side of wizard popup modal scrolls vertically (not just the content on the right hand side)
Screenshot 2024-05-20 at 21 35 05

arthur-lemeur:
Fixes #2290

Due to the length of some texts, the only solution I could find is to change the absolute positioning of the buttons (which placed them always at the bottom right of the modal) to a "default" positioning with a margin of 20px at there top and bottom. 
We still have scroll on some tabs and on other the buttons are a bit high.. This is not perfect but this is the only way the buttons are never over the images no matter the screen size or the text length.

### DIFF
--- a/src/renderer/assets/styles/components/modals.scss
+++ b/src/renderer/assets/styles/components/modals.scss
@@ -218,11 +218,12 @@
   margin-top: 30px!important;
 
   .guidedTour_tab {
-    margin-top: 0;
+    margin-top: 0!important;
 
     h3 {
       font-size: 25px;
       margin-top: 0;
+      margin-bottom: 10px;
     }
 
     p {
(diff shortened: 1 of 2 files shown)


danielweck:
Thank you @arthur-lemeur 
To be clear, the issue I filed was about the left hand side of the wizard scrolling (which this PR fixes, so I merged it). I personally didn't mind the overlaid "next" button in the fixed bottom right corner (in fact I prefer this design to the one introduced in this PR which forces me to scroll in order to reveal)...but I am really nit-picking and we have much more important / urgent features and bugs to work on right now, so no worries :)

PR #2307, opened 2024-05-22
fix(#2140): remove "do you want to save session" at closing

panaC:
replace #2157

### DIFF
--- a/src/common/redux/actions/session/index.ts
+++ b/src/common/redux/actions/session/index.ts
@@ -6,7 +6,9 @@
 // ==LICENSE-END==
 
 import * as enable from "./enable";
+import * as save from "./save";
 
 export {
+    save,
     enable,
 };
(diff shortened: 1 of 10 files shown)


danielweck:
Ok, 

@panaC So, instead of `value = messageValue.response;` could we pluck the boolean from global persistent application state (user-configurable preference, by default true = reader windows are restored automatically). No annoying confirmation dialog, the UX friction becomes minimal because users will just close their reader windows if they are annoyed by their re-appearance at next launch, or they will find the configuration checkbox or ask on the mailing list / GitHub :) (nobody reads the docs, sorry @gautierchomel ;)

danielweck:
Sorry, the message above is a copy-paste from the closed PR (which this one replaces), I sent it by mistake. I was going to post:

Ok, so in Thorium3 there will simply be no restoration of reader windows at all, users will have to manually figure out what publication they were reading when they closed the app. I think that's fine, just wanted to confirm that this is what we are going for in this feature revision.

PR #2334, opened 2024-05-23
fix two fr labels which were too long

llemeurfr:


danielweck:
Please make sure to only make changes to en.json via GitHub, all other locales should be modified via GitLocalize for now (until we fully understand the merge conflict risks)

src/resources/locales/en.json
```diff
@@ -84,7 +84,7 @@
         "noPublicationHelpL2": "or use the \"{{- importTitle }}\" button above.",
         "noPublicationHelpL3": "LCP-protected publications are supported.",
         "noPublicationHelpL4": "Supported extensions:",
-        "numberOfPages": "Number of pages",
```
danielweck, line 87: I think this English change will break https://github.com/edrlab/thorium-reader/blob/3290fd97cb45917d9ca1fd77bbb3a45e9999fcbd/src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx#L476-L486

I assume this change was meant for this particular GUI control?

https://github.com/edrlab/thorium-reader/blob/3290fd97cb45917d9ca1fd77bbb3a45e9999fcbd/src/renderer/library/components/searchResult/AllPublicationPage.tsx#L2098-L2107

danielweck (changes requested):

PR #2345, opened 2024-05-24
fixes #2329 fixes #2344

### LINKED ISSUE #2329: delete single bookmark in list (while in edit mode or not) causes crash (works fine with annotations editor)
Screenshot 2024-05-23 at 14 40 59

### LINKED ISSUE #2344: bookmarks and annotations list pagination logic breaks when deleting items
follow-up to https://github.com/edrlab/thorium-reader/pull/2332#issuecomment-2129258465

panaC:


src/renderer/reader/components/ReaderMenu.tsx
```diff
@@ -613,6 +615,9 @@ const AnnotationList: React.FC<{ r2Publication: R2Publication, dockedMode: boole
         }
     }, [isSearchEnable]);
 
+    if (!r2Publication || !annotationsQueue) {
```
danielweck, line 618: `!annotationsQueue?.length`, I think

src/renderer/reader/components/ReaderMenu.tsx
```diff
@@ -581,6 +578,11 @@ const AnnotationList: React.FC<{ r2Publication: R2Publication, dockedMode: boole
     };
 
     const [pageNumber, setPageNumber] = React.useState(getStartPage);
+    React.useEffect(() => {
+        if (pageNumber > pageTotal) {
+            setPageNumber(pageTotal);
+        }
+    }, [pageTotal]);
```
danielweck, line 585: dependency array, missing `pageNumber`?

src/renderer/reader/components/ReaderMenu.tsx
```diff
@@ -914,6 +915,12 @@ const BookmarkList: React.FC<{ r2Publication: R2Publication, dockedMode: boolean
     const pageTotal =  Math.floor(sortedBookmarks.length / MAX_MATCHES_PER_PAGE) + ((sortedBookmarks.length % MAX_MATCHES_PER_PAGE === 0) ? 0 : 1);
 
     const [pageNumber, setPageNumber] = React.useState(1);
+    React.useEffect(() => {
+        if (pageNumber > pageTotal) {
+            setPageNumber(pageTotal);
+        }
+    }, [pageTotal]);
```
danielweck, line 922: dependency array, missing `pageNumber`?

danielweck (changes requested):

src/renderer/reader/components/ReaderMenu.tsx
panaC: I don't think so, I want to trigger the test on pageTotal change

src/renderer/reader/components/ReaderMenu.tsx
panaC: I don't think so either

panaC:
I note a new bug about the bookmark docked component on refresh when i delete a bookmark and i make a new, the component is not DOM refreshed but logically refresh in local state, doesn't appear in annotation, just bookmark

src/renderer/reader/components/ReaderMenu.tsx
panaC: removed

danielweck: falsy `annotationsQueue` from IReaderRootState[reader.annotation] results in empty JSX, but empty array too, no?

src/renderer/reader/components/ReaderMenu.tsx
danielweck: doesn't this break rules of hooks?

> The array of dependencies is not passed as arguments to the effect function. Conceptually, though, that’s what they represent: every value referenced inside the effect function should also appear in the dependencies array. In the future, a sufficiently advanced compiler could create this array automatically

https://legacy.reactjs.org/docs/hooks-reference.html

danielweck (changes requested):

panaC:
https://github.com/edrlab/thorium-reader/pull/2345#discussion_r1613388858


no not at all, use here as subscriber, 

if pageNumber is passed as a dependency this allows you to have a "garde fou" on pageNumber

src/renderer/reader/components/ReaderMenu.tsx
danielweck: > make sure the array includes all values from the component scope (such as props and state) that change over time and that are used by the effect. Otherwise, your code will reference stale values from previous renders.

https://legacy.reactjs.org/docs/hooks-reference.html

panaC: `pageTotal` n'est ni un état ni un props 🙃

danielweck: any value within the closure scope (i.e. render function, given that hooks are always top-level) that participates to computations inside the hook's closure should be declared explicitly as a dependency (in fact a linter would probably yell at us for not following this rule).
but I don't fully understand what the logic is here so you probably have something in mind that justifies firing the effect for only a subset of conputation dependencies?

danielweck: ok, let's merge this PR, I trust you judgement regarding the runtime logic.

PR #2361, opened 2024-05-27
fix key enter to not submit gotopage form + css

### LINKED ISSUE #2342: page "number" selector (EPUB authored page list):  custom select/option instead of native, space doesn't select from list (types character), and enter selects but also submits (premature)

arthur-lemeur:
Fixes #2342


panaC (approved):
I let Daniel review

PR #2368, opened 2024-05-27
reader reduxState saved in registry and session at each reduxState update

### LINKED ISSUE #2289: The reader State is saved at closing, Annoying when annotation on multiple readerWin same publication
How to reproduce :

Open 2 **same** publications concurrently
make some annotation on a books, it's replicated on the other one.

image

Close a publication : 
make a last annotation at the end of the document (see capture, blue highlight) 

image

open the same publication in Library
The annotation is not keeped on the newly opened publication.

image

close the left publication in first and then the right or inversely
and you will see that the last **same** (uuid) publication closed, keep the state to the registry

image

The state is saved/persisted at reader closing only.
It's annoying for a regular user with multiple same publication opened. Or maybe an issue when you do not remember that the same publication is already opened but hided by the windows manager (tray,dock,...)

panaC:
fix #2289

### DIFF
--- a/src/main/redux/sagas/reader.ts
+++ b/src/main/redux/sagas/reader.ts
@@ -316,9 +316,25 @@ function* readerSetReduxState(action: readerActions.setReduxState.TAction) {
 
     const { winId, reduxState } = action.payload;
 
-    const pubId = yield* selectTyped(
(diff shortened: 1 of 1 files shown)


danielweck:
Feel free to merge this PR. As discussed, I think the risk of massive performance regression is minimal whereas the benefits of synchronized state is priceless :)

Issue #2219, opened 2024-05-17
Localisation update, translations for Thorium 3.0


llemeurfr:
Dear contributors, Thorium 3.0 will be released around May 20. The new version proposes a major update of the User Interface; several existing labels have been shortened and new features (mainly annotations) require new labels.

Localization is now eased by the use of GitLocalized, a simple translation tool. As usual, the English locale is the reference. 

Please go to https://gitlocalize.com/repo/9626 and Sign in with Github. Send us you Github ID so that we can add you as "translator". 

Choose your preferred language and open the json file in the interface. On the left side you'll see the English labels, on the right translated labels. Please check the existing translations. News labels will be blank in right cells, please fill them. There is a "comments" feature if a discussion is required on certain terms.  

When it is done, please Create a Review Request (top yellow button).   

# Thorium Localization (l10n) language translations

Information and discussions about localization and language translation process can be found in the [Thorium wiki l10n dedicated page](https://github.com/edrlab/thorium-reader/wiki/Localization-(l10n)-language-translations)

Actual known contributors are :

@horus68
@milliet
@martinpub
@Modeo147
@naglis
@jacoporinaldi
@peaceroad
@panaC
@nvdaes
@HadrienGardeur
@tofi86
@terhol-celia
@maitreya2019
@cary-rowen
@ThanatosDi
@georgegach
@sukiletxe

ThanatosDi:
I already Sign in with Github
https://gitlocalize.com/users/ThanatosDi

naglis:
https://gitlocalize.com/users/naglis

danielweck:
@naglis @ThanatosDi you are now team "translators" :)
Welcome and thank you

martinpub:
Thanks for the localisation call! For this round, @jonaslil will update the Swedish locale.

milliet:
@milliet

I'll make it ASAP,

Cheers
Pedro

Em sex., 17 de mai. de 2024, 04:28, L. Le Meur ***@***.***>
escreveu:

izorba:
https://gr.linkedin.com/in/zorbaioanna
Hi! I have start working, but I am not in the translators list
Can you help me?
Thanks!

nvdaes:
Hello: You can add me for Spanish.
My GitHub ID is: 15809252

danielweck:
@nvdaes @izorba @jonaslil you are now team "translators" :)
Welcome and thank you


@milliet first you need to log into GitLocalize :)

[...]

PR #2386, opened 2024-05-27
Revised Finnish translation

jkinnunen:


danielweck:
@gautierchomel could you please check the status of Finnish on GitLocalize? I don't mind merging this "direct" PR but normally everything goes through GitLocalize in order to avoid data "corruption" problems.
@jkinnunen see https://github.com/edrlab/thorium-reader/issues/2219

gautierchomel:
@jkinnunen thank you for your proposal. As Daniel commented we have adopted a gitlocalize workflow (see #2219) and a Finnish translation was  submitted via that canal, thanks to @MiiaKirsi and @KirsiYl. Maybe it is the same submission but I am not sure. Do not hesitate to provide us with context in the case it's a different work.
If there are notable differences between both works, may we invite you three to discuss together?

danielweck:
I merged this GitLocalize PR https://github.com/edrlab/thorium-reader/pull/2390 feel free to send some more :)

PR #2418, opened 2024-05-30
try to fix focus main in GoToLocator annotations/bookmark

### LINKED ISSUE #2384: need consistent behaviour: annotations, bookmarks, landmarks, search, TOC hyperlinks ... navigation ("goto locator") does not always focus keyboard into content area
To be verified: I think this works with bookmarks, landmarks, search, TOC (except the CTRL-SHIFT-N bug https://github.com/edrlab/thorium-reader/issues/2379 )

panaC:
Fixes #2384


src/renderer/reader/components/ReaderMenu.tsx
```diff
@@ -469,7 +469,9 @@ const AnnotationCard: React.FC<{ timestamp: number, annotation: IAnnotationState
                     e.preventDefault();
                     const closeNavAnnotation = !dockedMode && !(e.shiftKey && e.altKey);
                     goToLocator(annotation.locatorExtended.locator, closeNavAnnotation);
-                    dispatch(readerLocalActionAnnotations.focus.build(annotation));
+                    
+                    // not used anymore, TODO remove the action
+                    // dispatch(readerLocalActionAnnotations.focus.build(annotation));
```
danielweck, line 474: Ah, I think this is still needed, there is no focus steal, just highlighting and "scrolling into view" the currently active annotation. I need to test and verify my assumptions :)

danielweck (changes requested):

PR #2445, opened 2024-06-11
breadcrumb responsive

panaC:


arthur-lemeur (approved):

src/renderer/library/components/layout/BreadCrumb.tsx
```diff
@@ -5,141 +5,192 @@
...
-});
+    // I think it's not needed to run it inside an 'effect' because there are already a check of dependencies mutation
+    // React.useEffect(() => {
+    const spaceLeftWidth = Math.floor(spanLeft.current?.clientWidth || -1);
+    // const spaceLeftWidth = spanLeft.current?.clientWidth || -1;
+
+    // console.log(spaceLeftWidth, spanLeft.current?.clientWidth, containerWidth, between.length, firstOne.name.length, lastOne.name, lastOneFromBreadcrumbData.name);
+
+    if (between.length) {
+        if (lastOne.name !== lastOneFromBreadcrumbData.name || lastOne.path !== lastOneFromBreadcrumbData.path) {
```
panaC, line 128: If I follow your logic, we can imitates the useEffect dependencies check by using the `object.is` property instead of  triple equal comparaison. What is the rational to not use the react.useEffect to do this ? Why react do not provide a mechanism to automatically do this instead of manually check, do you have an idea ? I'm thinking on it

danielweck: this check is because useState doesn't shallow-compare the object keys, just referential identity comparison.

danielweck (changes requested):

package.json
```diff
@@ -255,6 +255,7 @@
     "@radix-ui/react-radio-group": "^1.1.3",
     "@radix-ui/react-tabs": "^1.0.4",
     "@radix-ui/react-tooltip": "^1.0.7",
+    "@react-hook/resize-observer": "^2.0.1",
```
panaC, line 258: https://github.com/radix-ui/primitives/blob/main/packages/react/use-size/src/useSize.tsx

PR #2460, opened 2024-06-18
fix behavior of bookmarks button in reader

### LINKED ISSUE #2340: misleading "bookmark suppression" icon in some cases
If I'm on a screen where there are already several bookmarks and I press the bookmark button without having selected any text, nothing happens, even though overing on the bookmark button displayed an icon indicating that a bookmark would be deleted. If I select text beforehand, it adds a bookmark (whereas the bookmark button indicated a delete action). 
So the icon displayed on mouse hover in this case (bookmarks already existing in the screen) is misleading. 

Compare this to v2.4: the behaviour was the same, but there was no change to the icon on mouseover; there was no information, not a misleading information. 

Note: the "bookmark suppression" icon is useful if there is only one bookmark on the screen : in this case this bookmark will be deleted by hitting the button.

arthur-lemeur:
Fixes #2340


panaC (changes requested):
https://github.com/edrlab/thorium-reader/blob/develop/src/renderer/assets/icons/add-alone.svg
and 
https://github.com/edrlab/thorium-reader/blob/develop/src/renderer/assets/icons/BookmarkRemove-icon.svg

The icon must be the bookmark icon without the cross but with a plus instead :)

arthur-lemeur:
There is no such icon in the Thorium icon library at the moment

PR #2485, opened 2024-07-02
fix delete annotation popover in docked mode

### LINKED ISSUE #2480: [GUI] delete popover just flash on screen, i cannot click on it when the icon is at the bottom right of the window
image

arthur-lemeur:
Fixes #2480


panaC:
ah ! Many thanks Arthur to found the issue, the boolean `hideWhenDetached` should not have been checked, see radix doc : `Whether to hide the content when the trigger becomes fully occluded.`

panaC:
There are some bug in the radix popover lib because the trigger is not fully occluded but however it's hided : 
see this example : 

image

Then : i just scroll a bit 
image

PR #2501, opened 2024-07-31
fix tabIndex navigation arrows

### LINKED ISSUE #2495: Navigating feeds using keyboard arrows
From a user: "In the OPDS Thorium catalog version 3, I cannot move pages using the keyboard (Next button), if I use the mouse click it can work fine."

It is true that for many people, using the left / right arrow key to browse "My books" or an OPDS catalog would be useful.

arthur-lemeur:
Fixes #2495

@llemeurfr, about "I cannot move pages using the keyboard (Next button)" : 
Yes you can do it but it is binded to specific keyboard shortcuts which are by default "Ctrl, SHIFT, ArrowLeft" and "Ctrl, SHIFT, ArrowRight" ("keyboardShortcuts.NavigatePreviousOPDSPage" / "keyboardShortcuts.NavigateNextOPDSPage"

### DIFF
--- a/src/renderer/library/components/opds/PageNavigation.tsx
+++ b/src/renderer/library/components/opds/PageNavigation.tsx
@@ -97,6 +97,7 @@ class PageNavigation extends React.Component<IProps, undefined> {
                     <button
                         className={stylesPublication.allBooks_header_pagination_arrow}
                         aria-label={`${__("opds.firstPage")}`}
+                        tabIndex={-1}
                         disabled={!pageLinks?.first[0]?.url || pageInfo?.currentPage === 0}
                     >
                         {pageLinks?.first[0]?.url ?
@@ -117,6 +118,7 @@ class PageNavigation extends React.Component<IProps, undefined> {
                             transform: "rotate(180deg)",
                         }}
                         aria-label={`${__("opds.previous")}`}
+                        tabIndex={-1}
                         disabled={!pageLinks?.previous[0]?.url || pageInfo?.currentPage === 0}>
                         {
                             pageLinks?.previous[0]?.url ?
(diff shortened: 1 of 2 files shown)


llemeurfr:
I just tested on the Blbliomedia feed (https://bm.ebibliomedia.ch/v1/home.opds2), which has no "previous page" information: 
Shift + Ctrl + ArrowLeft leads to the **NEXT** page, like Shift + Ctrl + ArrowRight.
To be decided it we consider it a bug (but in this case should Shift + Ctrl + ArrowLeft be inactive or display an error message?) or a feature.

panaC:
> I just tested on the Blbliomedia feed (https://bm.ebibliomedia.ch/v1/home.opds2), which has no "previous page" information: Shift + Ctrl + ArrowLeft leads to the **NEXT** page, like Shift + Ctrl + ArrowRight. To be decided it we consider it a bug (but in this case should Shift + Ctrl + ArrowLeft be inactive or display an error message?) or a feature.

https://github.com/edrlab/thorium-reader/issues/2554

src/renderer/library/components/opds/PageNavigation.tsx
```diff
@@ -97,6 +97,7 @@ class PageNavigation extends React.Component<IProps, undefined> {
                     <button
```
panaC, line 97: We need to remove the button tag, the React Router Link component render an anchor html element and not need to be surrouned by a button tag. The button element doesn't even allow to trigger the button action with space keydown/keyup.

panaC:
I'm closing this PR. Arthur works on a new global OPDS issue PR.

src/renderer/library/components/opds/PageNavigation.tsx
panaC: Okay an anchor wrapped inside a button allows to disable the link anchor.

PR #2504, opened 2024-08-02
fix readerSettings ranges to match options-values

### LINKED ISSUE #2500: Version 3 Line Spacing
My apologies for a duplicate of my post at thorium-reader-doc, but was not sure where to post it:

Thorium 3 has a change in functionality from Version 2 that renders Version 3 not usable for me.

In the Line Spacing options, the increments have changed to only 0.5, 1, 1.5, etc. This does not offer the fine control for Line Spacing that Version 2 did. In Version 2 you could change the Line Spacing in smaller increments. 1.13, 1.25, 1.35, 1.5. As it turns out, a Line Spacing of 1.35 was ideal for me and the ebooks I own, and that Line Spacing option is now gone. In Version 3, Line Spacing "1" is too tight, and the next one up "1.5" is too wide.

Would it be possible to re-enable the Line Spacing options from Version 2 in Version 3?

Everything else about Version 3 is an improvement over Version 2, and I would love to switch, but not with the regression in Line Spacing options. This change alone has made me revert to Version 2 for the time being.

arthur-lemeur:
Fixes #2500


src/renderer/reader/components/ReaderSettings.tsx
```diff
@@ -390,63 +390,65 @@ interface ITable {
 
 const Slider = ({ value, option, set }: { value: string, option: ITable, set: (a: Pick<ReaderConfig, "pageMargins" | "wordSpacing" | "letterSpacing" | "paraSpacing" | "lineHeight">) => void }) => {
     const [currentSliderValue, setCurrentSliderValue] = React.useState(option.defaultValue);
+    const [currentIndex, setCurrentIndex] = React.useState(() => optionsValues[option.parameter].findIndex((el) => el === option.defaultValue));
```
danielweck, line 393: This would fail at compile time with TypeScript `strictNullChecks` (or `strict`) ... which is not yet activated in this project unfortunately.
I would recommend `optionsValues[option.parameter]` ==> `(optionsValues[option.parameter] || [])` and also `.findIndex((el) => el === option.defaultValue)` ==> `.findIndex((el) => el === option.defaultValue) || 0`

src/renderer/reader/components/ReaderSettings.tsx
```diff
@@ -390,63 +390,65 @@ interface ITable {
...
 
     React.useEffect(() => {
-        setCurrentSliderValue(value.replace(/rem/g, ""));
-    }, [value]);
+        setCurrentSliderValue(value);
+        const newIndex = optionsValues[option.parameter].findIndex((el) => el === value);
```
danielweck, line 397: same remark as above regarding sanity-checking `optionsValues[option.parameter]` for `undefined`.

src/renderer/reader/components/ReaderSettings.tsx
```diff
@@ -390,63 +390,65 @@ interface ITable {
...
+        setCurrentIndex(newIndex);
+    }, [value, option.parameter]);
+
+    const updateValue = (index: number) => {
+        const newValue = optionsValues[option.parameter][index];
```
danielweck, line 402: Same sanity-checking remark.

src/renderer/reader/components/ReaderSettings.tsx
```diff
@@ -390,63 +390,65 @@ interface ITable {
...
-        } else {
-            newStepValue = Number(currentSliderValue.replace(/rem/g, "")) + step;
-        }
-        const clampedValue = Math.min(Math.max(newStepValue, option.min), option.max);
-        const valueToString = clampedValue.toFixed(2);
-        setCurrentSliderValue(valueToString);
-        set({ [option.parameter]: valueToString + (option.rem ? "rem" : "") } as Pick<ReaderConfig, "pageMargins" | "wordSpacing" | "letterSpacing" | "paraSpacing" | "lineHeight">);
+        setCurrentIndex((prevIndex) => {
+            const newIndex = direction === "out" ? prevIndex - 1 : prevIndex + 1;
+            if (newIndex >= 0 && newIndex < optionsValues[option.parameter].length) {
```
danielweck, line 411: Same sanity-check workaround here ... unless we can "guarantee" through the TypeScript type system that `option.parameter` is within bounds of the `optionsValues` array range.

danielweck (changes requested):

src/renderer/reader/components/ReaderSettings.tsx
arthur-lemeur: Changes done

src/renderer/reader/components/ReaderSettings.tsx
arthur-lemeur: Changes done

src/renderer/reader/components/ReaderSettings.tsx
arthur-lemeur: As `parameter: keyof AdjustableSettingsStrings` and ```
const optionsValues: AdjustableSettingsStrings = {
    fontSize,
    pageMargins,
    wordSpacing,
    letterSpacing,
    paraSpacing,
    lineHeight,
};
```
option.parameter is guaranteed to be in optionValues array range

PR #2510, opened 2024-08-16
fix: oauth2 implicit grant

### LINKED ISSUE #2506: OAuth 2.0 Implicit Authentication querystring parameters missing
## Summary 

My app primarily supports the OAuth 2.0 Authorization Code with PKCE Flow but I am also implementing OAuth 2.0 Implicit flow in order to provide compatability with OPDS Clients.  The OPDS Authentication document returned when accessing my Bookshelf API is:

```json
{
    "id": "https://localhost:7063/auth.json",
    "title": "Bookshelf",
    "authentication": [
        {
            "type": "http://opds-spec.org/auth/oauth/implicit",
            "links": [
                {
                    "href": "https://localhost:7136/connect/authorize",
                    "rel": "authenticate"
                }
            ]
        }
    ]
}
```

And the response header has the Content-Type `application/opds-authentication+json`.

This is very similar to the example provided in the [Authentication for OPDS 1.0](https://drafts.opds.io/authentication-for-opds-1.0#233-example), minus the help links.

When accessing an OPDS feed source that requires authentication, Thorium Reader receives this OPDS authentication document. Stepping through source with the debugger, I can see that Thorium Reader correctly detects the Implicit OAuth flow as supported. It then opens the browser but does not append the required query string values `client_id` or `response_type`. I expect OPDS Clients to use the required OPDS OAuth Client Id `http://opds-spec.org/auth/client` when authenticating and the Implicit Flow should request a `token`.

When I embed these values into the URL in the opds authentication document, Thorium Reader opens the link successfully. I don't think this is the correct behavior: 

* The example in the documentation does not have these query string parameters
* Including these query string parameters in the URL will force all OPDS Clients to use the same `client_id`, even though the [specification states](https://drafts.opds.io/authentication-for-opds-1.0#342-a-shared-client-identifier) that "Specific OPDS clients **may** also be registered by each Catalog, in order to be further identified". Hard coding the `client_id` in the Authentication Document URL will prevent this.

## Versions Tested

I have tested this using current stable release `v3.0.0` and current alpha release `v3.1.0-alpha.1`.

## Actual Behaviour

Thorium opens the URL provided in the authentication document exactly as provided and does not append the required query string parameters: `response_type` or `client_id`.

## Expected Behaviour

Thorium opens a browser window to the link provided in the authentication document, adding the parameters `response_type` and `client_id` at a minimum, but ideally with a generated parameter `state` to prevent CSRF attacks on the login page.

Example of expected URL to be opened: `https://localhost:7136/connect/authorize?client_id=http%3A%2F%2Fopds-spec.org%2Fauth%2Fclient&response_type=token&state=41ysjf5ts5vw1lt90sbgk4e0g7nd89iw`

panaC:
add parameter to the authentification web url :
- clientId with the document id
- response_type to token
- state to prevent CSRF attack but not enable for the moment

Fixes #2506


src/main/redux/sagas/auth.ts
```diff
@@ -452,14 +453,38 @@ async function opdsSetAuthCredentials(
...
+            try {
+                const browserUrlParsed = new URL(auth.links?.authenticate?.url);
+
+                // The unique identifier used in the callback URI must have the same value as the identifier provided in the Authentication Document.
+                // https://drafts.opds.io/authentication-for-opds-1.0#342-a-shared-client-identifier:~:text=The%20unique%20identifier%20used%20in%20the%20callback%20URI%20must%20have%20the%20same%20value%20as%20the%20identifier%20provided%20in%20the%20Authentication%20Document.
+
+                // see also the callback response : it must use the id query parameter to indicate the Authentication Document identifier
+                // https://drafts.opds.io/authentication-for-opds-1.0#342-a-shared-client-identifier:~:text=it%20must%20use%20the%20id%20query%20parameter%20to%20indicate%20the%20Authentication%20Document%20identifier
+
+                browserUrlParsed.searchParams.set("client_id", encodeURIComponent_RFC3986(auth?.id || "http://opds-spec.org/auth/client"));
```
Dunnymeister, line 476: I think using the `auth?.id` here is incorrect. The `id` specified in the OPDS Authentication Document is the unique identifier of the document itself, not the id for the client to use when authenticating.

src/main/redux/sagas/auth.ts
```diff
@@ -780,6 +805,17 @@ function parseRequestFromCustomProtocol(req: Electron.ProtocolRequest)
                     data[key] = value;
                 }
 
+                if (data.state !== getImplicitNonceForImplicitAuthentication()) {
+                    debug("received state parameter is not equal to the nonce sent", "value=", data.state);
+                    debug("the state verification is IGNORED and bypassed to ensure a legacy compatibility with all OPDS OAUTH2 server");
+                    // TODO: improve this and enable the state verification
```
Dunnymeister, line 811: I think a possibility for enabling backwards-compatible behaviour would be to only ignore the `state` nonce if it is missing in the response from the OAuth server. Something like:

```js
// https://auth0.com/docs/secure/attack-protection/state-parameters
if (data.state) {
    if (data.state === getImplicitNonceForImplicitAuthentication()) {
        debug("State parameter was included in callback response and is VERIFIED and CORRECT");
    }
    else {
        // Todo: Reject the response, user is potentially compromised
        debug("received state parameter is not equal to the nonce sent", "value=", data.state);
    }
}
else {
    debug("State verification missing in response - verification checks IGNORED and bypassed to ensure legacy compatibility with all OPDS OAUTH2 server");
}
```

Dunnymeister: Thinking about it a little more, this probably doesn't provide much additional security. If a user's connection is compromised, a malicious actor could just omit the `state` parameter in the crafted response and the check would be bypassed.

src/main/redux/sagas/auth.ts
Dunnymeister: From the documentation:

> Example of an OPDS callback containing an Access Token
> 
> `opds://authorize/?id=http%3A%2F%2Fexample.org%2Fauth.json&access_token=9b3dc428-df5f-4bd2-9f0d-72497cbf8464&token_type=bearer`

Using the [example](https://drafts.opds.io/authentication-for-opds-1.0#233-example) as a guide, the OAuth Server should returns an `opds://authorize` redirect after successful authentication should include a querystring parameter with the opds document id in it.

panaC:
I'm merging now, let's continue the discussion in several separate issues. Many Thanks @mpdunlop

PR #2511, opened 2024-08-19
Changes to OAuth 2.0 Implicit Grant Flow handling

mpdunlop:
Made the following changes on top of @panaC's branch:

* Not using the OPDS Authentication Document's Id as the client_id, using "http://opds-spec.org/auth/client" in all cases
* OPDS Authentication Document Id is validated against the id returned in the callback_url (validation skipped if missing for backwards compatability)
* Basic OPDS Authentication Document validation
* Debug messaging improvements to provide developers with context about why requests are rejected
* Documentation of OAuth parameters and any workarounds for OPDS Authentication/OAuth specification conflicts

### DIFF
--- a/src/main/redux/sagas/auth.ts
+++ b/src/main/redux/sagas/auth.ts
@@ -576,15 +576,11 @@ function opdsAuthDocConverter(doc: OPDSAuthenticationDoc, baseUrl: string): IOPD
 
     // https://drafts.opds.io/authentication-for-opds-1.0#231-core-properties
     if (!doc.Id || doc.Id === "") {
-        // Todo: Return typed error so user-friendly message can be displayed
-        debug("OPDS Authentication Document missing required `id` property.");
-        return undefined;
+        debug("OPDS Authentication Document missing required `id` property. IGNORED and bypassed to ensure legacy compatibility with misconfigured OPDS OAuth 2.0 servers");
     }
 
     if (!doc.Title || doc.Title === "") {
-        // Todo: Return typed error so user-friendly message can be displayed
-        debug("OPDS Authentication Document missing required `title` property.");
-        return undefined;
+        debug("OPDS Authentication Document missing required `title` property. IGNORED and bypassed to ensure legacy compatibility with misconfigured OPDS OAuth 2.0 servers");
     }
 
     if (!doc.Authentication) {
--- a/src/main/redux/sagas/auth.ts
+++ b/src/main/redux/sagas/auth.ts
@@ -453,35 +453,46 @@ async function opdsSetAuthCredentials(
     return [, new Error("")];
 }
 
-let _implicitNonce = "";
-const setAndGetImplicitNonceForImplicitAuthentication = () => (_implicitNonce = nanoid(16), _implicitNonce);
-const getImplicitNonceForImplicitAuthentication = () => _implicitNonce;
+const _implicitAuthData = { authenticationDocumentId: "", nonce: "" };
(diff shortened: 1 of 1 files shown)


src/main/redux/sagas/auth.ts
```diff
@@ -558,16 +568,28 @@ interface IOPDSAuthDocParsed {
...
-
     if (!doc || !(doc instanceof OPDSAuthenticationDoc)) {
+        // Todo: Return typed error so user-friendly message can be displayed
+        debug("OPDS Authentication Document is not an instance of the model");
+        return undefined;
+    }
 
-        debug("opds authentication doc is not an instance of the model");
+    // https://drafts.opds.io/authentication-for-opds-1.0#231-core-properties
+    if (!doc.Id || doc.Id === "") {
```
panaC, line 578: Authentication ID must be set https://drafts.opds.io/authentication-for-opds-1.0#233-example:~:text=The%20Authentication%20Document%20must%20contain%20the%20following%20keys%3A

but the opds parser is mostly not opinionated and allow to parse badly formatted opds format by making assumptions. Currently the Id is not checked, I wonder if this will invalidate misconfigured opds servers unnecessarily.

@danielweck What is your opinion ?

src/main/redux/sagas/auth.ts
```diff
@@ -453,35 +453,46 @@ async function opdsSetAuthCredentials(
...
+                
+                // Record the OPDS Authentication Document's Id for later verification
+                const implicitAuthData = getImplicitAuthData();
+                implicitAuthData.authenticationDocumentId = auth.id;
 
-                browserUrlParsed.searchParams.set("client_id", encodeURIComponent_RFC3986(auth?.id || "http://opds-spec.org/auth/client"));
+                // client_id: Required. 
+                // Any Authentication Provider that supports OPDS Authentication 1.0 and exposes an Authentication Flow based on OAuth
+                // must identify all OPDS clients using the client_id "http://opds-spec.org/auth/client"
+                browserUrlParsed.searchParams.set("client_id", "http://opds-spec.org/auth/client");
```
panaC, line 480: I don't know if it's good practice to hardcode the `client_id`. I understand that this can't be the same value as the authentication ID as I was able to code it in the first PR.
I think it's not a big deal and should work as there is only one "client" registered like Thorium.

What do you think @HadrienGardeur ? When you wrote the opds auth specification, did you think about this?
I poke @danielweck  also.

src/main/redux/sagas/auth.ts
mpdunlop: Updated PR to ensure backwards compatibility with already misconfigured OPDS Authentication Providers.

src/main/redux/sagas/auth.ts
HadrienGardeur: Normally, an Authorization Server should have a unique `client_id` for every client app interacting with them.

That said, to allow public clients to interact with Authorization Servers, Authentication for OPDS introduces a shared client identifier along with a shared redirect URI. It's expected that OPDS clients will use this shared `client_id` as the default/fallback whenever they encounter an Authentication Document using OAuth. 

This is something that we'll have to revisit very soon though since we're adding support for Authorization Code with PKCE to a future revision of Authentication for OPDS.

mpdunlop: @panaC This response indicates that OPDS clients should use the shared `client_id` defined in the Authentication for OPDS spec. Is it ok to proceed like this?

panaC: seems good for me

mpdunlop:
Just checking back in on this. Is there anything else you need from me to help move it forward?

PR #2513, opened 2024-08-19
Speech TTS voice selector

panaC:


panaC:
The tts voice is attached to the publication. 
when first opening a publication, the tts voice is not set. The tts voice is updated only when opening the tts control panel not before. So the tts voice is not set on first reading until the user opens the control panel, the voice is in a undefined state.

when the tts voice has been updated in the control panel, the voice is attached and persisted to the publication. 

default Voice : 
image

default Voice accessible EPUB3 : 
image


About the Localisation : the languages labels and regions labels in both combobox is translated with [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames).


TODO : 

- need to merge the readium-speech [PR](https://github.com/readium/speech/pull/7) 
- update the npm package link to 'readium-speech' when pushed to NPM registry
- need to improve the css margin of the ttsVoice component in control panel
- need to discuss about the tts voice persistence across publication.

panaC:
This PR is ready to review
- npm git package added from https://github.com/readium/speech
- readium-speech instance launched from ReaderHeader
- ...

danielweck:
nit-pick: see the GUI gap:

Screenshot 2024-11-27 at 15 26 10

danielweck:
I tested with Wasteland, the UX is a net improvement over Thorium v3.0.0 (I was able to easily select and test English, German, Italian, Greek, etc. languages).

panaC:
@danielweck I will separate the src and build folders with git branches as I did on your recommendation on edrlab/pdf.js#master2 vs edrlab/pdf.js#build

panaC:
I'm plan to merge this PR today, seems to be enough mature to be merged asap, we have just to keep updated the readium/speech repo.

PR #2553, opened 2024-09-19
proposal fix: to remove annoying saveFilePicker tittle when export annotations lists

panaC:
… macOs


panaC:
On the Chromium implementation, and specially in MacOS when the title is visible to the user. There is a Warning message  : "Warning: This site can see edits you make".
This Warning Message is very annoying, In thorium we run like a native application, and moreover we do not read the file chosen by the user, we just overwrite and dump the annotations content generated.
The implementation in chromium is [here]( https://github.com/electron/electron/blob/0ac8ff6439a5448d5f883ef6656a3e9be753a7e2/shell/browser/file_system_access/file_system_access_permission_context.cc#L800), it's  a new API endpoint, probably always Work In Progress, And not implemented everywhere on each browser.

Chromium issue : https://issues.chromium.org/issues/40061219
Source code : https://github.com/electron/electron/blob/0ac8ff6439a5448d5f883ef6656a3e9be753a7e2/shell/browser/file_system_access/file_system_access_permission_context.cc#L800
https://issues.chromium.org/issues/40061219#:~:text=This%20change%20adds%20a%20title%20for%20the%20file%20picker%20when%0Athe%20showSaveFilePicker%20is%20used.%20This%20title%20should%20be%0Aupdated%20again%20if%20we%20decide%20to%20separate%20read/write%20access.

> This change adds a title for the file picker when
the showSaveFilePicker is used. This title should be
updated again if we decide to separate read/write access. "

For me, I think we cannot choose to separate read or write access from the API (https://developer.mozilla.org/en-US/docs/Web/API/Window/showSaveFilePicker)

I'm trying to switch with this PR on a simpler method to just create a download link with the blob content, electron catch the download blob file and ask to user when he want to save it.

PR #2557, opened 2024-09-20
Fix(annotation): annotationEdit initialise to the last color, drawType & force hex color comparaison to upper case

### LINKED ISSUE #2548: [Bug]: annotation edit popover not initialize with the color and drawType in readerConfig
image

Color and Highlight has always the same default initial state.

panaC:
Fixes #2548

### DIFF
--- a/src/common/rgb.ts
+++ b/src/common/rgb.ts
@@ -12,12 +12,12 @@ export function rgbToHex(color: { red: number; green: number; blue: number }): s
     const redHex = Math.min(255, Math.max(0, red)).toString(16).padStart(2, "0");
     const greenHex = Math.min(255, Math.max(0, green)).toString(16).padStart(2, "0");
     const blueHex = Math.min(255, Math.max(0, blue)).toString(16).padStart(2, "0");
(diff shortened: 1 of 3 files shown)


danielweck:
Just in case Thorium receives a colour code from an external source, note that the RGB HEX syntax isn't necessarily `#RRGGBB`, see:

https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color

In other words, the `hexToRgb()` regular expression will break with all other RGB HEX syntaxes. This might not be a problem right now if the function only consumes strings from Thorium itself, where the RGB HEX syntax is only `#RRGGBB` :)

danielweck:
PS: in the highlights engine I just use `rgb(rrr, ggg, bbb)` instead of `#RRGGBB` as this doesn't need any format conversion, I can just pass the raw numbers from the IColor struct.
That being said, the RGB HEX syntax is of course easier to store in persistent state of during import/export stages, which I guess is why the conversion takes place?

panaC:
Indeed ! Thanks you for your precision.

PR #2585, opened 2024-10-02
windows touch selection quick hacking patch

### LINKED ISSUE #2473: Annotations should be compatible with touchscreens
From an early user of Th3: 

I am playing with Thorium 3 for Windows on a touchscreen "tablet" (specifically, it's a Surface Pro 9, but the behavior I describe below reproduces on my Dell laptop with a touch screen).

If I select text with my *mouse*, I can add annotations.  If I select text with my finger or my stylus and then try to add an annotation, I get a "No selection" error (even though text is very clearly selected).

panaC:
Fixes #2473


danielweck:
Fixed via https://github.com/readium/r2-navigator-js/commit/c17df631b743600e3e9124c889914292d26b4e85

PR #2590, opened 2024-10-03
Dev/feat autocomplete tags

### LINKED ISSUE #1853: Feature: autofill tags with existing tags
Use case: 

As a reader with a large number of books I want to organize them with tags but because I add books thru differents days, months or years, I don't always remember how exactly was lettered my tag.

It results that I may add wrong tags (exemple: "Science Fiction" instead of "Science & Fiction") and lost books until I figure out and correct the wrong tag.

To prevent such mistakes I would like Thorium to help me autocomplete while I write Tags with proposal of existing tags.

arthur-lemeur:
Fixes #1853


panaC:
src/renderer/common/components/dialog/publicationInfos/tag/TagPredict.tsx is not used at all !

You need to cleanup the PR before reviewing :)

PR #2592, opened 2024-10-05
chore(dev): rationalize translator for the 3 processes and store the locale value to redux only / remove inversify in renderer process / in class component split translator context and component refresh when locale updated / remove unused old class component 

### LINKED ISSUE #2593: initialize default app language with the preferred systems language
related to https://github.com/edrlab/thorium-reader/pull/2490

panaC:
Fixes #2593 
Closes #2490


danielweck:
huge change set :)
what does this fix?

PR #2609, opened 2024-10-14
up annotations menu UI

arthur-lemeur:


panaC (changes requested):
2 issues : 

- on Tab cycling : setting icon is hit first, maybe because the setting icon button (previously a summary details) is not on a same line that AnnotationList header component and simulated with a negative css position.
- In filter options : creator summary-details element should not be tabable when not enabled.

panaC:
The develop merge is too complex for me, I don't own your change and I risk to break your code. And moreover we have to deal with the setting button not on the same annotations options line.

panaC (approved):

PR #2610, opened 2024-10-14
up readerSettings Preset tab UI + translation

arthur-lemeur:


panaC (changes requested):
Capture d’écran du 2024-10-15 14-27-27
Capture d’écran du 2024-10-15 14-27-51


reset buttons not aligned 


Also I notice that when I click on the button the blue focus is not released, I don't know if my click is captured or not, no blue outline change.

PR #2611, opened 2024-10-14
annotations import modal new design and some bug fixes 

panaC:


src/resources/locales/en.json
```diff
@@ -124,6 +124,16 @@
     "update": "Edit"
   },
   "dialog": {
+    "annotations": {
+      "descAuthor": "of {{- author}}",
```
naglis, line 128: I am not a native English speaker, but perhaps a more natural variant here would be `by {{- author}}`? I.e.:
_42 note(s) from John Smith will be associated with Moby-Dick by  Herman Melville_
vs.
_42 note(s) from John Smith will be associated with Moby-Dick of  Herman Melville_

panaC: @danielweck

panaC: it seems you are right https://english.lingolia.com/en/vocabulary/confusing-words/of-from-by#:~:text=to%20introduce%20the%20subject%20(agent)%20of%20something%2C%20often%20in%20a%20passive%20sentence 

I'm far to be a native english speaker, so I let daniel to answer you :)

PR #2640, opened 2024-11-05
Fixes #2602 (Electron BrowserWindow WebContents LifeCycle Sanity Check)

### LINKED ISSUE #2602: [crash] browserWindow win.webContents when destroyed
[win.webContents](https://github.com/edrlab/thorium-reader/blob/ff1f46bf4ddf03ced453c78ee6aff596228c6d33/src/main/redux/sagas/app.ts#L87)

danielweck:


danielweck:
Hello @panaC in the code diff you will see:i

1) the obvious `isDestroyed` code fencing / sanity checks
2) `yield take(winActions.library|reader.openSucess.build)` replaced with `yield take(winActions.library|reader.openSucess.ID)` (I can't believe this passed compilation! no idea what the runtime behaviour was/is??)
3) `"did-finish-load"` ALWAYS used, not just in `DEV` mode otherwise we can pretty much guarantee differences with the production builds which we do not test as much as the development builds. Consequently I commented the `yield put(winActions.library.openSucess...` immediately after the `loadUrl`, and also to improve code legibility I positioned the `"did-finish-load"` event handler BEFORE the `loadUrl` (consistent for both reader and library windows)

panaC:
`take` support the `toString` with `StringableActionCreator` 

https://github.com/edrlab/thorium-reader/blob/9d2e3e840425b9d35292f036167148c0f1812f9f/src/main/redux/actions/win/reader/openSucess.ts#L30 

https://github.com/redux-saga/redux-saga/blob/9ec63efbffb576a5d2c8b4bbf9df8f020693fdad/packages/types/types/index.d.ts#L22

danielweck:
ah thanks! phew! :)

that being said, this fact is obfuscated and inconsistent with the many other places in Thorium's codebase where take() uses the ID directly instead of via build().toString() 
...let's do a quick code scan to see where else this is used.

panaC (approved):
many thanks Daniel

panaC:
> ah thanks! phew! :)
> 
> that being said, this fact is obfuscated and inconsistent with the many other places in Thorium's codebase where take() uses the ID directly instead of via build().toString() ...let's do a quick code scan to see where else this is used.

One day we will switch to the latest [rtk](https://redux-toolkit.js.org/) api  ! :)  I hope :)

danielweck:
ah, I verified "npm run build:prod" before merging!
do you get this error in the develop branch??

panaC:
I make this error, to prove action type checking is now not verified

danielweck:
I see. In the meantime I was looking at the codebase again, it would be better to be consistent the other way around: only use `.build`, not `.ID` :)

danielweck:
Follow-up: https://github.com/edrlab/thorium-reader/commit/79a62f3184fc66eb39a38fc6dd98c42e3f51903f

PR #2649, opened 2024-11-10
fix(i18n): translate "Go To Page" when reader menu is docked

naglis:
Currently, when the reader menu window is docked, the "Go To Page" label in the combobox is not using translations, while the label is translated if the menu is not docked.

Running 941fb36c30974b43eee1493aed326f0c49d26acf in Lithuanian language:
Screenshot displays the issue of the "Go To Page" label not being translated into Lithuanian language in the docked reader menu in Thorium Reader while reading a publication


danielweck:
well-spotted, thank you! :)

PR #2651, opened 2024-11-11
fix(i18n): remove untranslated title attribute from General settings tab

naglis:
This value is not translated and is being displayed as the tooltip when hovering on any element in the "General" settings tab:

Screenshot displaying the tooltip with the untranslated word "General" while hovering on an element in the "General" settings tab while Thorium is set to use the Lithuanian language

As [the other tabs do not have](https://github.com/edrlab/thorium-reader/blob/1e026d2efee7987c7ce9459fb77a7a7799d07849/src/renderer/library/components/settings/Settings.tsx#L357-L363) a `title` attribute, I've elected to remove it instead of introducing the `title` attribute to other tabs.


danielweck:
Thank you for spotting this error and for submitting a Pull Request, much appreciated :)

Issue #2536, opened 2024-09-11
Update Display of accessibility metadata following Guide 2.0


gautierchomel:
User Experience Guide for accessibility metadata 2.0 first public draft is available. The set of documents includes:

- [Principles document](https://w3c.github.io/publ-a11y/UX-Guide-Metadata/2.0/principles/) 
- [ONIX metadata Techniques](https://w3c.github.io/publ-a11y/UX-Guide-Metadata/2.0/techniques/onix-metadata/index.html)  
- [EPUB Accessibility metadata Techniques](https://w3c.github.io/publ-a11y/UX-Guide-Metadata/2.0/techniques/epub-metadata/index.html)

Thorium pioneered the display of such information in a reading system, but the way it is presented is now outdated. 

We should implement this significant update so it can be used to test and provide feedback.

gautierchomel:
For implementation testing: 

- [Exploded EPUB ](https://github.com/edrlab/thorium-reader-epub-tests/tree/main/Rare_EPUB_test_suit/a11y_metadata)
- [Packed EPUB](https://github.com/edrlab/thorium-reader-epub-tests/blob/main/Rare_EPUB_test_suit/a11y_metadata.epub)
- [OPDS](https://rawcdn.githack.com/edrlab/thorium-reader-epub-tests/8a9fc9f1eaecdd65726eed348191b21a79097cf6/EPUB_Feeds_for_testing_purpose/opds-rare-epub-suit.json)

gautierchomel:
Thorium should use Compact sentences with no titles for categories and keep things as simple as a list. 

Example:

- Appearance can be modified
- Not fully readable in read-aloud and braille
- Navigation by table of contents, index, landmarks, and page list
- This publication meets minimum accessibility standards

gautierchomel:
The guide is still being refined by editors. 

An [exemple processor](https://daisy.github.io/a11y-meta-viewer/) is available from a [dedicated daisy repo](https://github.com/daisy/a11y-meta-viewer)

gautierchomel:
IMHO our implementation should differ to avoid titles with only one element and include `detail` element for more information. Something like 

<dl id="result">
  <dt>Reading modes</dt>
    <dd>Appearance can be modified</dd>
    <dd>Readable in read aloud or dynamic braille</dd>
    <dd>Prerecorded audio</dd>
  <dt>Navigation</dt>
    <dd>
        Navigation by table of contents, index, supports page navigation, and
        headings
[...]

PR #2716, opened 2024-12-13
modifies text content of accessibility information keys

gautierchomel:
Very partial update implementation of display guide 2.0. See #2536


llemeurfr:
@gautierchomel do you insist having long a11y descriptions in Thorium? We only have limited space in the Information modal box, not a large HTML page, as some retailers propose on their websites. 
You're the a11y expert, therefore you will choose, but my warning is that with the long descriptions, the Information box will be bloated with a11y information (when it exists for the book).

gautierchomel:
current available space seems large enough
image

gautierchomel:
This is how it looks on the shorter possible screen (first screen shot is a standard ebook while the second one is the alla informations testing ebook). In both cases, it does not adds a line. To me it's a better use of the space (less white space on the right).
image
image

gautierchomel:
In any case, it a a very partial implementation of the new guide, only shifting the existing words to ship something in the Thorium 3.1 instead of remaining with the old wording. 

Full guide implementation will need more work as it comes with variants values for each keys (Readable in read aloud or dynamic braille; May not be fully readable in read aloud or dynamic braille; Not fully readable in read aloud or dynamic braille; No information is available).

Issue #2168, opened 2024-05-13
"about" info shipped with Thorium (offline accessible, unlike online doc HTML/EPUB/OPDS)


danielweck:
Currently XHTML not supported by GitLocalize, but Markdown would work fine (TODO: make "about" info much shorter, display in modal popup instead of loading full EPUB):

https://github.com/edrlab/thorium-reader/tree/develop/src/resources/information

This is for post-3.0 release

gautierchomel:
I can make a proposal if it helps.

gautierchomel:
Thorium 3.1 gets ride of the about / info page packed with Thorium and proposes a direct link to the [online thorium about webpage](https://thorium.edrlab.org/en/about/).

danielweck:
dine.

gautierchomel:
Reopening as we still have no offline information panel.

danielweck:
How does this fit with the custom profile feature? CC @llemeurfr and @panaC

PR #2717, opened 2024-12-13
Proposal for Information panel

gautierchomel:
Old localised information embedded as EPUB is not used anymore. Instead, the About link directly brings the user to the About page of the Thorium website, leaving him with a new context to understand. See #2168 for discussion context. 

This proposal is to adopt the usual "about panel" available in apps: a simple popup with contextualised links to what the user may be searching for (documentation, support or info).  

It only provides `en.html` to localise. The visualisation panel for this information has to be implemented.


gautierchomel:
This is a very draft proposal to follow up discussions at https://github.com/edrlab/thorium-reader/issues/2168

PR #2722, opened 2024-12-18
Adds "(online)" mention to the link that will change user context by opening the browser 

gautierchomel:
That's an alternate quick fix for #2717 while we continue discussions about the necessity of having an in-app about panels for short contextual information.


gautierchomel:
Note that this PR inherited the previous branch commit about the already merged bookshelf label. Not sure how to avoid that without having to recreate a new branch.

naglis:
Perhaps another option could be to use a globe icon next to the text to indicate that is it a web resource? E.g. "About Thorium 🌎". It could use one of the globe icons that are already used in Thorium: https://github.com/edrlab/thorium-reader/blob/ded9817fcd660bf4c46defd5eb447e0df66d3057/src/renderer/assets/icons/globe-icon.svg or https://github.com/edrlab/thorium-reader/blob/ded9817fcd660bf4c46defd5eb447e0df66d3057/src/renderer/assets/icons/globe-icon-bold.svg

Although it would still require the extra "(online)" text for screen reader users?

PR #2723, opened 2024-12-18
add query params lang and app version to documentation link

arthur-lemeur:


src/renderer/library/components/catalog/AboutThoriumButton.tsx
```diff
@@ -104,7 +104,7 @@ class AboutThoriumButton extends React.Component<IProps, IState> {
                     <p>{`v${_APP_VERSION}`}</p>
                     <a href="" onClick={async (ev) => {
                                 ev.preventDefault(); // necessary because href="", CSS must also ensure hyperlink visited style
-                                await shell.openExternal("https://thorium.edrlab.org/"); // TODO: not link to the english thorium hyperlink
+                                await shell.openExternal(`https://thorium.edrlab.org/?lang=${this.props.locale}&v=${_APP_VERSION}`);
```
danielweck, line 107: although not strictly necessary here as we can make reliable assumptions about the syntax of the query params values, I strongly suggest - as a matter of good programming practice - to escape the query param values with encodeURIComponent()

danielweck (changes requested):

PR #2732, opened 2024-12-19
fix(opds): download a "sample" or "open access" publication behind an authwall

panaC:
See: https://github.com/edrlab/thorium-reader/issues/2512


panaC:
superseded by https://github.com/edrlab/thorium-reader/pull/2733

Issue #2660, opened 2024-11-15
i18n: Add support for plural forms


naglis:
i18next, the library used by Thorium for translations, supports [plurals](https://www.i18next.com/translation-function/plurals), but it is currently not used in Thorium.

I think it would be nice to display the correctly pluralized strings instead of appending all the possible variations at the end (e.g. _42 match(es)_).

IIUC, there are a few strings that would benefit from plurals in Thorium:

- [`reader.picker.search.founds`](https://github.com/edrlab/thorium-reader/blob/3eef469fe71592c5c590aca2386b7e654bb36222/src/resources/locales/en.json#L429)
- [`dialog.annotations.descList`](https://github.com/edrlab/thorium-reader/blob/3eef469fe71592c5c590aca2386b7e654bb36222/src/resources/locales/en.json#L127)
- [`dialog.annotations.descNewer`](https://github.com/edrlab/thorium-reader/blob/3eef469fe71592c5c590aca2386b7e654bb36222/src/resources/locales/en.json#L128)
- [`dialog.annotations.descOlder`](https://github.com/edrlab/thorium-reader/blob/3eef469fe71592c5c590aca2386b7e654bb36222/src/resources/locales/en.json#L129)
- [`dialog.deleteAnnotationsText`](https://github.com/edrlab/thorium-reader/blob/3eef469fe71592c5c590aca2386b7e654bb36222/src/resources/locales/en.json#L137)

I have experimented with adding support for it (i18next JSON [v3](https://github.com/naglis/thorium-reader/tree/plural-v3), [v4](https://github.com/naglis/thorium-reader/tree/plural-v4)), it seems to work nicely. Please note that in order for i18next to apply plurals, the number placeholder must be named `count`.

There are some issues/unanswered questions:

- Not sure which i18next JSON version to use. I have found v4 plural suffixes a bit easier to understand ([here is a JSFiddle](https://jsfiddle.net/6bpxsgd4) which displays the plural suffixes for a given language/i18next JSON version). It might also make sense to bite the bullet and switch to v4 before adding plurals so as not to have to migrate the keys afterwards. I am not sure if there are any issues (unrelated to plurals) from switching from v3 -> v4, though.
- `i18next-json-sync` (the CLI tool to extract translatable strings) [does not support plural keys](https://github.com/jwbay/i18next-json-sync/issues/22) (it simply removes them), although there are alternatives, e.g. https://github.com/felixmosh/i18next-locales-sync.
[...]

PR #2743, opened 2025-01-01
i18n: enable i18next plurals

naglis:
Addresses: #2660

Use the `count` placeholder name for strings which might have different translations depending on the number of results, which enables nicer messages, e.g. "1 match" instead of "1 match(es)".

`count` is the reserved placeholder for plural forms used in [i18next](https://www.i18next.com/translation-function/plurals).

Translations can use different suffixes to provide the right messages for all the plural forms applicable to that language. To find out which suffixes should be used for a particular language, [this JSFiddle](https://jsfiddle.net/6bpxsgd4) can be used.

NOTE: GitLocalize most likely will not support the extra plural forms in different languages, so in order to add/modify them, the translator will have to make PRs. Also, `i18next-json-sync` (the CLI tool to extract translatable strings) [does not support plural keys](https://github.com/jwbay/i18next-json-sync/issues/22) (it simply removes them). See #2660 for more information.


danielweck:
thank you for keeping this up to date :)
we'll be discussing soon.

danielweck:
Fixes https://github.com/edrlab/thorium-reader/issues/2660

PR #2766, opened 2025-02-01
Fix typo in translation

naglis:


danielweck:
thank you so much for spotting this, AND creating a PR :)

PR #2791, opened 2025-02-13
[PARKED] feature: Image Description With AI

panaC:
Based on https://github.com/edrlab/thorium-reader/pull/new/feat/image-click


panaC:
`MISTRAL_API_KEY=xxxxx npm run start:dev`

image
image

src/renderer/library/components/settings/Settings.tsx
```diff
@@ -235,122 +237,175 @@ const SaveCreatorSettings: React.FC<{}> = () => {
...
+            )}
+        </div>
+    );
+};
+
+const ApiKeysList = () => {
+    const [__] = useTranslator();
+
+    const apiKeys = useSelector((state: ICommonRootState) => state.apiKeys);
+    const [keys, setKeys] = React.useState(apiKeys);
```
panaC, line 377: Why do you need to store the reference to the array keys in the useState ? 

`addNewKey` is used only once when they are one or zero key stored, but why ? 
Then, why do you need to save the entire reference of the apiKeys from the redux store to append an empty apiKey object. You can just use a local boolean to show a non initialised ApiKeyComponent when the button "add new key" is clicked. 

What do you think, you overcomplicated the component I guess.

src/renderer/library/components/settings/Settings.tsx
```diff
@@ -235,122 +237,175 @@ const SaveCreatorSettings: React.FC<{}> = () => {
     );
 };
 
-const ApiKeys = () => {
+const ApiKeyComponent = ({ k, index }: {k: IApiKey, index: number}) => {
```
panaC, line 240: pass the array index in component props is generally not a good idea, you should copy the api key/provider/(id?) tuple instead

src/renderer/library/components/settings/Settings.tsx
```diff
@@ -235,122 +237,175 @@ const SaveCreatorSettings: React.FC<{}> = () => {
...
-
+    const apiKeys = useSelector((state: ICommonRootState) => state.apiKeys);
 
-    const addNewKey = () => {
-        setKeys([...keys, { provider: "", key: "", submitted: false}]);
-        setInputType([...inputType, "password"]);
-    };
+    const [newProvider, setNewProvider] = React.useState("");
+    const [newKey, setNewKey] = React.useState("");
+    const [inputType, setInputType] = React.useState("password");
```
panaC, line 248: I think you do not need to store locally the "password" input type, at worst a boolean

panaC: I'm wondering if the logic need to save locally a component state for this feature, I think you can get the type state from the HTMLInputElement dom reference, I think

panaC:
@arthur-lemeur can you merge master and resolve these minor? conflicts ?

src/common/redux/actions/api_key/index.ts
```diff
@@ -0,0 +1,15 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END==
+
+// import * as enable from "./enable";
+import * as set from "./setKey";
+import * as removeKey from "./removeKey";
+
+export {
+    set,
```
panaC, line 13: you need to have a stability with the action naming, note `set` and `removeKey` difference

src/common/redux/actions/api_key/setKey.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { Action } from "readium-desktop/common/models/redux";
+import { IApiKey } from "../../states/api_key";
+
+export const ID = "API_KEY_SET";
+
+export interface Payload extends IApiKey {
+}
+
+export function build(key: string, provider: string, submitted: boolean): Action<typeof ID, Payload> {
```
panaC, line 16: I guess `submitted` is a local component UI state and not a global state needed to be saved to the redux state, What do you think ?

src/common/redux/actions/api_key/removeKey.ts
```diff
@@ -0,0 +1,27 @@
...
+import { IApiKey } from "../../states/api_key";
+
+export const ID = "API_KEY_REMOVE";
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+// interface IPayload extends Pick<IAnnotationState, "uuid"> {
+interface IPayload extends IApiKey {
+}
+
+export function build(param: IApiKey):
```
panaC, line 18: I think we just need the api provider name identifier in parameter not the entire IApiKey interface, but why not

src/common/redux/reducers/api_key.ts
```diff
@@ -0,0 +1,40 @@
...
+                ...state,
+                {
+                    provider: action.payload.provider,
+                    key: action.payload.key,
+                    submitted: action.payload.submitted,
+                },
+            ];
+        case apiKeysActions.removeKey.ID:
+            return state.filter(
+                key => key.provider !== action.payload.provider || key.key !== action.payload.key,
```
panaC, line 32: what about the same key for several api provider ?

src/main/redux/middleware/persistence.ts
```diff
@@ -44,6 +44,7 @@ export const reduxPersistMiddleware: Middleware
                     settings: prevState.settings,
                     creator: prevState.creator,
                     annotationImportQueue: prevState.annotationImportQueue,
+                    apiKeys: prevState.apiKeys,
```
panaC, line 47: we should not persist apiKeys to the redux state but instead in a separate encrypted json file, I think

src/common/redux/states/api_key.ts
```diff
@@ -0,0 +1,14 @@
+// ==LICENSE-BEGIN==
+// Copyright 2017 European Digital Reading Lab. All rights reserved.
+// Licensed to the Readium Foundation under one or more contributor license agreements.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
+// ==LICENSE-END==
+
+export interface IApiKey {
+    key: string,
+    provider: string,
```
panaC, line 10: `provider` is not a string but an enumeration of provider name define as an apiKeys identifier. One an unique key per provider name i presume ?

src/common/redux/states/api_key.ts
```diff
@@ -0,0 +1,14 @@
...
+    submitted: boolean,
```
panaC, line 11: again, why the boolean "submitted" ?

src/common/redux/reducers/api_key.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const initialState: IApiKeysArray = [];
+
+function apiKeysReducer_(
+    state = initialState,
+    action: apiKeysActions.set.TAction | apiKeysActions.removeKey.TAction,
+): IApiKeysArray {
+
+    switch (action.type) {
+        case apiKeysActions.set.ID:
```
panaC, line 21: I guess that this is more an add/update that an infinite push to an array

panaC: And in this case, remove is not needed, it's more a reset feature for the key of the API AI Provider than a total remove, api providers is known and fixed

src/renderer/library/components/settings/Settings.tsx
arthur-lemeur: I changed the useState to a boolean, but I do not only need to get the type of the input I need to edit it from "password" to "text" and vice versa when clicking the eye button, hence the useState. But if you have a better way to do it I'll take it.

src/renderer/library/components/settings/Settings.tsx
```diff
@@ -235,122 +237,175 @@ const SaveCreatorSettings: React.FC<{}> = () => {
...
 
-const ApiKeys = () => {
+const ApiKeyComponent = ({ k, index }: {k: IApiKey, index: number}) => {
     const [__] = useTranslator();
+    const dispatch = useDispatch();
 
-    const [keys, setKeys] = React.useState([]);
-    const [inputType, setInputType] = React.useState(["password"]);
-
+    const apiKeys = useSelector((state: ICommonRootState) => state.apiKeys);
```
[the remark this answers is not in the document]
arthur-lemeur: I need the apiKeys redux state to know which provider are already stored, if any, in order to know which provider I should allow the user to choose from.

src/renderer/library/components/settings/Settings.tsx
panaC:
```html

<script>

  addEventListener("DOMContentLoaded", () => {

    const input = document.getElementById("in");
    const button = document.getElementById("but");

    button.addEventListener("click", () => {

      if (input.type === "text") {
        input.type = "password";
      } else {
        input.type = "text";
      }
    });

  });
</script>

<input id="in" value="hello world" type="text"/>
<button id="but">Click ME</button>

```

src/renderer/library/components/settings/Settings.tsx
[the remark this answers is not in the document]
panaC: keep it simple, once created we should not allow to change the provider name identifier across the keyEdit component, the edit/remove action should be only on one specific provider at once.

src/renderer/library/components/settings/Settings.tsx
arthur-lemeur: I dont see how this is better than a useState ? Isn't the purpose of the useState to do exactly that in React ?

``    const [seePassword, setSeePassword] = React.useState(false);
return (
<input
                                type={seePassword ? "text" : "password"}
                                name="api-key"
                                className="R2_CSS_CLASS__FORCE_NO_FOCUS_OUTLINE"
                                title={k.provider}
                                value={newKey}
                                onChange={(e) => setNewKey(e.target.value)}
                                required
                                ref={ref}
                            />
                            <button
                                type="button"
                                onClick={(e) => {
                                    e.preventDefault();
                                    setSeePassword(!seePassword);
                                }}
                            >
                                <SVG ariaHidden svg={seePassword ? EyeClosedIcon : EyeOpenIcon} style={{ width: "20px" }} />
                            </button>
``

panaC: But it's works don't worry.

panaC:
```html
<script src="https://unpkg.com/@babel/standalone/babel.min.js"></script>
<script type="text/babel" data-type="module" defer> 
import * as React from 'https://esm.sh/react';
import { createRoot } from 'https://esm.sh/react-dom/client';

let _ref;
const Password1 = () => {
    const [seePassword, setSeePassword] = React.useState(true);
    const inputRef = React.useRef();
    return <>
        <input
            type={seePassword ? "text" : "password"}
            value={"Hello World"}
            ref={(ref) => {
              if (!_ref) {
                _ref = ref
              }
              console.log("refChanged?", ref !== _ref);
            }}
        />
        <button
            type="button"
            onClick={() => {
              setSeePassword(!seePassword);
            }}
        >
        { seePassword ? "HIDE" : "SHOW" }
        </button>
    </>
}

const Password2 = () => {
    const inputRef = React.useRef();
    return <>
        <input
            type="text"
            value="Hello World"
            ref={inputRef}
        />
        <button
            type="button"
            onClick={() => {
              inputRef.current.type = inputRef.current.type === "password" ? "test" : "password";
            }}
        >
        SVG 
        </button>
    </>
}
const app = document.getElementById("app");
createRoot(app).render(<><Password1/><br/><Password2/></>);
</script>
<div id="app"/> 

```

Sorry my previous message is mostly wrong, the react VDOM doesn't 'remount' the HTMLInputElement as the type attribute changed, It's miss-understanding of the react internal work. Obviously the ref Input element is the same. The element is not replaced by a fresh new InputElement with the reflection of the type attribute. The VDOM update just copy the react type attribute to the Document Object Model Element type attribute. 

In the example `Password1` , we can see that the ref attribute doesn't changed, but for each new rendering loop, the ref lambda function is called twice, due to the react rendering update method.

What I wanted to explain to you if I may, is the unnecessary react rendering loop when you just want to update an internal state/attribute of a mounted HTMLElement (cf `Password2`). Now I agree with you, the more "react" way is the first example `Password1` that reflect the internal HTMLInputElement in a local React state, and allows to update the rendering children value of the button, for example, when state change ([React data binding](https://www.geeksforgeeks.org/reactjs-data-binding/)).

danielweck:
Nit-picks:

* Lack of padding in multiline text input area (the one with white background and black text)
* In dark mode, text is invisible inside single-line text input field (the one with black background should contain white text)

Screenshot 2025-04-18 at 10 08 33

danielweck:
> Hello Team, My plan is to split this PR is 2 parts : image viewer part and AI part (back/front/API-Keys) under a feature flag.

That's great, good luck!
This will fix https://github.com/edrlab/thorium-reader/issues/2976

danielweck:
maybe useful resource?
https://github.com/sul-rds/avila/blob/main/prompts/menuez_qwen2.5-VL.yaml
from:
https://www.choice360.org/libtech-insight/creating-alt-text-for-digital-collections-with-ai-a-case-study/

danielweck:
This PR/branch contains hard-to-resolve merge conflicts, can this PR now be closed as obsolete or is it still relevant? (the PR code diff will remain in the GitHub records so we will be able to check back on the proposed changes, if necessary ... I just want to make sure that stale PRs with significant technical debt and no activity do not "pollute" the list of pending pull requests)

PR #2823, opened 2025-02-26
TTS voices : persist default voice per language / refactor GUI 

### LINKED ISSUE #2811: TTS voice selection per language code (GUI checkboxes), pass to navigator (handle multiple preferences)
Suggestion: emit toast notification

panaC:
Fixes #2811


danielweck:
I reviewed the PR code...I am wondering about BCP47 language code comparisons, shouldn't the logic make sure to prioritise accurate complete matches, then fallback to prefix only?

For example, assuming `selectedLanguage.code` can be a full `en-US` or `en-UK` code, or a partial `en` code:

```js
const selectedVoice = defaultVoices.find(({language}) => language.split("-")[0].toLowerCase() === selectedLanguage.code);
```

===>

```js
const selectedVoice = defaultVoices.find(({language}) => language.toLowerCase() === selectedLanguage.code) || defaultVoices.find(({language}) => language.split("-")[0].toLowerCase() === selectedLanguage.code);
```

I ran a search for `split("-")` to find all ocurrences of language code comparisons, but I don't fully understand the possible values of the operands so I am not sure if the "prioritize-full + fallback-to-partial" logic is necessary.

danielweck:
There's a GUI refresh bug (multiple TTS pause + play sequences, ultra-repeated?) when selecting language and/or voice.
This doesn't occur when selecting speech rate / playback speed, or highlights style / colours, etc.

danielweck (changes requested):
GUI refresh bug

danielweck:
Other than the GUI refresh bug, I think this works. I tested The Waste Land on MacOS with different English and German voices (Greek doesn't work because they use a different language code, and on my computer I only have a single voice for Italian)

panaC:
> I reviewed the PR code...I am wondering about BCP47 language code comparisons, shouldn't the logic make sure to prioritise accurate complete matches, then fallback to prefix only?
> 
> For example, assuming `selectedLanguage.code` can be a full `en-US` or `en-UK` code, or a partial `en` code:
> 
> ```js
> const selectedVoice = defaultVoices.find(({language}) => language.split("-")[0].toLowerCase() === selectedLanguage.code);
> ```
> 
> ===>
> 
> ```js
> const selectedVoice =
>        defaultVoices.find(({language}) => language.toLowerCase() === selectedLanguage.code)
>     || defaultVoices.find(({language}) => language.split("-")[0].toLowerCase() === selectedLanguage.code);
> ```
> 
> I ran a search for `split("-")` to find all ocurrences of language code comparisons, but I don't fully understand the possible values of the operands so I am not sure if the "prioritize-full + fallback-to-partial" logic is necessary.


we can do a full comparaison by security, but

`selectedLanguage.code` can only be of type lang: 'en','fr' (2 or 3 first character before the hyphen and lowercase) or of type 'US','FR' (2 last character after the hyphen and uppercase) 

cf: https://github.com/readium/speech/blob/0f17fc25e2956de5e6b1174fc9813c36fd8ff623/src/voices.ts#L249

danielweck:
try changing the TTS voice while playing ... the GUI flickers massively.

this doesn't happen when changing the speech rate, the highlights etc

panaC:
> try changing the TTS voice while playing ... the GUI flickers massively.
> 
> this doesn't happen when changing the speech rate, the highlights etc

For me (macbook m1), there is the same (huge!?) GUI flickering no matter what I change like speech rate or voice. I will investigate

danielweck:
I am seeing a duplicate Spanish voice on Mac, could you please test at your end?

Screenshot 2025-03-03 at 16 49 14

panaC:
I'm merging now

panaC:
> The problem with the voice selector is that this often kills the TTS (stops playing, impossible to resume)

It seems to me that the issue is still present, let's continue to track the issue on a dedicated issue

danielweck:
Duplicate TTS voice fixed in readium-speech, thank you :)

Screenshot 2025-03-04 at 10 07 35

danielweck:
I posted a new issue for the TTS configuration panel flickering: https://github.com/edrlab/thorium-reader/issues/2839

PR #2827, opened 2025-03-01
l10n: enable Arabic language

naglis:
This enables Arabic language use in Thorium. The translation was done by <span>@</span>salah1974 in https://github.com/edrlab/thorium-reader/pull/2429.

I've used the localized name for Arabic from [the Wikipedia entry](https://en.wikipedia.org/wiki/Arabic). I'd appreciate if someone could double check if that is the correct word to use.

It also updates the language list in the README to include Arabic and the recently added Czech.

Also mentioning @gautierchomel to update the [Localization wiki](https://github.com/edrlab/thorium-reader/wiki/Localization-(l10n)-language-translations).


danielweck:
Thank you very much. I propose we merge this so it becomes testable in Thorium from the automated builds.

PR #2828, opened 2025-03-01
SHIFT CTRL B add bookmark with user input'ed label (prefilled with selected text, if any)

danielweck:


danielweck:
@panaC feel free to close this PR if you are working "from scratch" instead of continuing the code changes in this drafty one :)

llemeurfr:
Is it a bookmark we want to create or an annotation?

danielweck:
bookmark, nothing different, other than the ability to avoid the auto-numbering by editing the label immediately after creation.

panaC:
> @panaC by the way I find the auto-numbering very confusing because not "stable", i.e. numbers not strictly associated with particular bookmark at time of creation, but dynamically allocated to indicate the order based on global progression of the locator (percentage). So as a user if I create a bookmark first at 50%, it gets auto label "bookmark 1" ... unfortunately if I create the next label at 25%, now my old bookmark which I mentally memorised as number 1 now is labelled "bookmark 2" .... super confusing (I'm used to it now in the sense that I expect the odd labelling scheme, and i just don't rely on it for memorizing).

I think so, but how to fix that ? index identifier is the most accessible identifier, percent progression is per resources and resource name identifier is mostly not very user friendly (chapter1.xhtml,index1.html,...)

danielweck:
> 
> I think so, but how to fix that ? index identifier is the most accessible identifier, percent progression is per resources and resource name identifier is mostly not very user friendly (chapter1.xhtml,index1.html,...)

no problem with index, that's fine.
the problem is that a given bookmark's label changes based on its position in the ordered list. once created, the label should be fixed.

panaC:
> > I think so, but how to fix that ? index identifier is the most accessible identifier, percent progression is per resources and resource name identifier is mostly not very user friendly (chapter1.xhtml,index1.html,...)
> 
> no problem with index, that's fine. the problem is that a given bookmark's label changes based on its position in the ordered list. once created, the label should be fixed.

I see thanks

PR #2851, opened 2025-03-06
fix tts options popover responsive for small screens

arthur-lemeur:


src/renderer/assets/styles/components/readerHeader.scss
```diff
@@ -236,6 +236,15 @@
     /* gap: 20px; */
     color: var(--color-primary);
 
+    @media only screen and (width <= 1000px) {
+        flex-direction: column;
+        max-height: 450px;
+
+        details {
+            display: none;
```
danielweck, line 244: I disagree, this fonctionality was introduced for users who depend on a specific TTS highlight style to make their reading experience accessible. These users may have small / narrow displays (or may decide to resize the window, or dock the reader window on left/right sides of a large display). There is still screen real estate available on the horizontal axis, let's not hide the preview panel, let's resize it to fit.

danielweck (changes requested):
I am posting screenshots to demonstrate the CSS breakpoint

danielweck:
Screenshot 2025-03-06 at 09 28 48
Screenshot 2025-03-06 at 09 28 55

PR #2872, opened 2025-03-11
fix: bookmark auto-numbering and initialise/migrate created/modifed/creator bookmark data state 

### LINKED ISSUE #2860: bookmark auto-numbering is very confusing, tied to GUI list item order, not actual data model
For example, when deleting "Bookmark 1", it seems to re-appear because the numbering is dynamic for the displayed GUI list, it is not intrinsic to the bookmark data model itself.

Discussion:
https://github.com/edrlab/thorium-reader/issues/2829#issuecomment-2704696235

panaC:
fixes #2860

### DIFF
--- a/src/renderer/reader/index_reader.ts
+++ b/src/renderer/reader/index_reader.ts
@@ -81,7 +81,7 @@ ipcRenderer.on(readerIpc.CHANNEL,
                 data.payload.annotationTagsIndex = pushTags({}, annotationTagsList);
 
                 const bookmarkList = data.payload.reader.bookmark || [];
-                let bookmarkIndex = 0;
+                let bookmarkIndex = data.payload.reader?.config?.bookmark_totalcount || 0;
                 for (const [created, bookmark] of bookmarkList) {
                     if (!bookmark.created && created > bookmark.modified) {
                         bookmark.created = bookmark.modified;
--- a/src/renderer/reader/index_reader.ts
+++ b/src/renderer/reader/index_reader.ts
@@ -80,9 +80,20 @@ ipcRenderer.on(readerIpc.CHANNEL,
                 }
                 data.payload.annotationTagsIndex = pushTags({}, annotationTagsList);
 
+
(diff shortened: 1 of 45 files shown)


src/renderer/reader/index_reader.ts
```diff
@@ -78,6 +80,31 @@ ipcRenderer.on(readerIpc.CHANNEL,
...
+                for (const [created, bookmark] of bookmarkList) {
+                    bookmarkIndex = bookmarkIndex + 1;
+                    if (!bookmark.created && created > bookmark.modified) {
+                        bookmark.created = bookmark.modified;
+                    }
+                    if (!bookmark.created) {
+                        bookmark.created = created;
+                    }
+                    if (!bookmark.name) {
+                        bookmark.name = `bookmark_${bookmarkIndex}`; // TODO? Probably translation issue !?
```
panaC, line 104: @danielweck do you have an idea to deal with bookmark no name migration ? In this PR we force the name of the bookmark to not rely of the bookmark index list position for the naming in UI if no name is available.

panaC:
@danielweck I'm merging this PR, feel free to open any issues that you want :)

panaC:
quick space hit of the bookmark button : 
image

src/renderer/reader/index_reader.ts
danielweck: I think the affordance that renders empty bookmark names (like empty annotation descriptions) must make it clear that this is empty. No need for a placeholder, let alone an arbitrary number bound to the list position (which can vary depending on filtering, ordering, etc.)

panaC: seems fine, just this in this case : 
```
bookmark.name = `${bookmarkIndex}`;
```

panaC: https://github.com/edrlab/thorium-reader/commit/6307e4ab4406b2af8ece81a3b19af64f173386ac

PR #2874, opened 2025-03-12
feat: SHIFT CTRL B keyboard shortcut edit and add a bookmark

### LINKED ISSUE #2859: new bookmark feature: SHIFT CTRL B keyboard shortcut, creation + immediate label edit
We could bind a key modifier like SHIFT on the "bookmark" button click, or ALT-SHIFT which seems to perform better cross-platform (we already use it for TTS / Media Overlays play toggle).

Discussion:
https://github.com/edrlab/thorium-reader/issues/2829#issuecomment-2704696235

panaC:
Fixes #2859


panaC:
ready to merge

panaC:
Uploading Screen Recording 2025-03-13 at 16.20.39.mov…

panaC:
https://github.com/user-attachments/assets/c661a3f5-5583-4ac7-ae8d-1bec1c84c427

panaC:
> here or in a separate PR: we need to make sure that when clicking on a bookmark in the margin, the paginated list goes to the correct page and scroll to the correct offset and highlights / outlines the selected bookmark. right now I have to manually search for the bookmark I want to edit. same problem with annotations: can't really tell which list item has visual emphasis / keyboard focus.

I think it's solved now, you should try

panaC:
> here or in a separate PR: we need to make sure that when clicking on a bookmark in the margin, the paginated list goes to the correct page and scroll to the correct offset and highlights / outlines the selected bookmark. right now I have to manually search for the bookmark I want to edit.
same problem with annotations: can't really tell which list item has visual emphasis / keyboard focus.

except when the bookmark list is paginated, this will be the work of my next PR indeed

PR #2884, opened 2025-03-17
add link in toast to see pub in the library

arthur-lemeur:


src/renderer/library/components/searchResult/AllPublicationPage.tsx
```diff
@@ -397,6 +397,17 @@ const CellGlobalFilter: React.FC<ITableCellProps_GlobalFilter> = (props) => {
...
+    const location = useLocation();
+    const queryParams = new URLSearchParams(location.search);
+    const searchValue = queryParams.get("value") || "";
+
+    React.useEffect(() => {
+        if (props.focusInputRef?.current) {
+            props.focusInputRef.current.value = searchValue;
+            onInputChange(searchValue || undefined);
+        }
+    }, [searchValue, onInputChange, props.focusInputRef]);
```
danielweck, line 409: in the dependency array, I think `props.focusInputRef` should be `props.focusInputRef?.current?.value` because the React "ref" is likely stable but its "current" value is what changes.

src/renderer/common/components/toast/Toast.tsx
```diff
@@ -131,6 +135,8 @@ export class Toast extends React.Component<IProps, IState> {
                 break;
         }
 
+        console.log(this.props);
```
danielweck, line 138: TODO: remove console log

src/resources/locales/en.json
```diff
@@ -194,6 +194,7 @@
         "import": {
             "alreadyImport": "[{{- title}}] was already imported.",
             "fail": "Importing [{{- path}}] failed: [{{- err}}]",
+            "seeInLibrary": "See in the library.",
```
danielweck, line 197: TODO: "Show in library"

src/renderer/common/components/toast/Toast.tsx
```diff
@@ -185,7 +191,46 @@ export class Toast extends React.Component<IProps, IState> {
...
+                }>{this.props.publicationTitle ?
+                    <span>
+                        {this.props.message}
+                        <br/>
+                        <Link
+                            style={{textDecoration: "underline", fontStyle:"italic", fontWeight: "bold", color: "var(--color-primary)"}}
+                            to={{
+                                ...location,
+                                pathname: "/library",
+                                search: `?focus=search&value=${this.props.publicationTitle}`,
```
danielweck, line 203: This can match multiple publications, I wonder if we could be more specific? (uuid?)

danielweck (changes requested):

panaC:
image

When importing notes from the reader , the toast is displayed with "go to library" instead of focus to the reader window

panaC:
There are an another issue when importing notes/bookmark with the toast 
I'm investigate 
image

PR #2888, opened 2025-03-19
fix: fix floating-promises errors

akx:
While viewing this code, I noticed there were a handful of places where promises were left floating – for instance, in the LCP passphrase save code, the code never actually waited for the write to go through (or whether it failed), which could be problematic.

I noticed [the ESlint rule for the issue](https://typescript-eslint.io/rules/no-floating-promises/) had been turned off, so I turned it on and fixed the issues – either by `await`ing where possible, or `void`ing the promises, so as to mark them "yes, we mean to ignore the return value of this promise".

### DIFF
--- a/src/main/cli/index.ts
+++ b/src/main/cli/index.ts
@@ -228,7 +228,7 @@ const yargsInit = () =>
                 if (openPublicationRequestedBool) {
 
                     // flush session because user ask to read a publication
-                    flushSession();
+                    await flushSession();
 
                     // pathArgv can be an url with deepLinkInvocation in windows
                     // https://github.com/oikonomopo/electron-deep-linking-mac-win
@@ -316,7 +316,7 @@ export function commandLineMainEntry(
     debug("processArgv", processArgv, "arg", argFormated);
 
     try {
-        y.parse(argFormated);
+        void y.parse(argFormated);
     } catch (e) {
         debug("YARGS ERROR !!!!!", e);
     }
--- a/src/main/services/device.ts
+++ b/src/main/services/device.ts
@@ -65,7 +65,7 @@ export class DeviceIdManager implements IDeviceIDManager {
             const deviceConfig: DeviceConfig = {
                 device_id: deviceId,
             };
-            this.saveDeviceConfigJson(deviceConfig);
+            await this.saveDeviceConfigJson(deviceConfig);
             debug(deviceConfig);
         }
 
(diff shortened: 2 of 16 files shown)


src/main/services/device.ts
```diff
@@ -65,7 +65,7 @@ export class DeviceIdManager implements IDeviceIDManager {
             const deviceConfig: DeviceConfig = {
                 device_id: deviceId,
             };
-            this.saveDeviceConfigJson(deviceConfig);
+            await this.saveDeviceConfigJson(deviceConfig);
```
panaC, line 68: the promise is not returned from the function https://github.com/edrlab/thorium-reader/blob/8069ae9d3b6c52036304d79cb35bb0886428031e/src/main/services/device.ts#L107

src/main/cli/index.ts
```diff
@@ -218,7 +218,7 @@ const yargsInit = () =>
                 if (openPublicationRequestedBool) {
 
                     // flush session because user ask to read a publication
-                    flushSession();
+                    await flushSession();
```
panaC, line 221: `flushSession` doesn't need to be `async` i guess

https://github.com/edrlab/thorium-reader/blob/8069ae9d3b6c52036304d79cb35bb0886428031e/src/main/tools/flushSession.ts#L11

akx: Making it non-async feels out of scope for this particular PR.

However, it being async doesn't matter – it acts like a synchronous function since there are no continuation points (`await`s) within.

danielweck:
Generally-speaking, non-awaited Promises are "code smell" and there should be ESLint rules to check this.

Regardless of whether or not a Promise/async-function has a return value, it should be awaited unless of course the Promise itself is fed into a subsequent process that is designed to consume it.

...there is also the whole other issue of try/catch error handling on rejected Promises, function coloring and drilling error states up into the call chain (which in the case of Thorium involves Redux Saga yield'ed functions).

akx:
> Generally-speaking, non-awaited Promises are "code smell" and there should be ESLint rules to check this.

Indeed. As mentioned in the PR description, the rule had been turned off (in 8cf571e69239bba6009f03b47e1c72924a934144, 2023, to be precise). It's now on.

> Regardless of whether or not a Promise/async-function has a return value, it should be awaited unless of course the Promise itself is fed into a subsequent process that is designed to consume it.

There can be merit to awaiting the promise within a function instead of returning the unawaited promise (namely, more localized stack traces should the promise be rejected).

akx:
@danielweck Rebased on current `develop`. There were 6 new issues I fixed.

danielweck:
This PR / external branch contains very good suggestions but also contains obsolete changes given the evolutions of the codebase since this was last proposed. I am closing the PR but I am recording an issue to reference it, and to act on "floating promises" linting errors!

danielweck:
https://github.com/edrlab/thorium-reader/issues/3670

PR #2907, opened 2025-03-24
add link in toast to see pub in the library [PR #2] !!

panaC:


danielweck (changes requested):
Screenshot 2025-04-16 at 09 58 43

danielweck:
Hello @panaC this reader window crash is reproducible 100%:

1) open Thorium (library window)
2) open EPUB
3) come back to library
4) optionally open a second EPUB (not necessary, but matches my screenshot below)
5) drag and drop an EPUB on the library window (it doesn't matter if new publication or already imported)
6) observe the crash in the displayed reader window(s)

Screenshot 2025-04-25 at 16 08 06

danielweck:
Problem: router `Link` doesn't exist in reader context. That's ok, we need to rethink the use of Toasts for this functionality, better implement a dedicated "import result" UX that gives the user plenty of time to decide which publication to link to after bulk import, or import from OPDS feed.

PR #2938, opened 2025-04-17
Keyboard Shortcut GUI Searchable and Localized

arthur-lemeur:


src/renderer/library/components/settings/KeyboardSettings.tsx
```diff
@@ -137,6 +144,21 @@ class KeyboardSettings extends React.Component<IProps, IState> {
             passive: false,
             capture: true,
         });
+
+        const userAgent = window.navigator.userAgent;
```
panaC, line 148: Are you confident enough with this OS detection ?

src/renderer/library/components/settings/KeyboardSettings.tsx
```diff
@@ -146,6 +168,69 @@ class KeyboardSettings extends React.Component<IProps, IState> {
     public render(): React.ReactElement<{}> {
         const { __ } = this.props;
 
+        const isSearchEmpty = !this.state.searchItem || this.state.searchItem.trim() === "";
+
+        const cleanNames = {
```
panaC, line 173: you need to type the `cleanNames` object with : https://github.com/edrlab/thorium-reader/blob/00cc65a19f9e3801e2da312e569bd7e1f62e3b10/src/common/keyboard.ts#L400C13-L400C32

gautierchomel (commented):
Looking at the localisation keys, we should use the same as for the button titles, so we do not duplicate localisation work. 
Alternatively, when it's not possible, we should make sure to prefix the key with something like "shortcut" or "keyboard" to help localizers understand where it goes.

src/resources/locales/en.json
```diff
@@ -564,6 +564,63 @@
             "editUserJson": "Edit user file...",
             "keyboardShortcuts": "Keyboard shortcuts",
             "loadUserJson": "Load from user file",
+            "name": {
```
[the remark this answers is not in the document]
panaC: I agree with Gautier, we should have something like that : 

```json
keyboard: {
  list: {
    AddBookmarkWithLabel: {
      description: "...",
      name: "",
    }
  }
}
```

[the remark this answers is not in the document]
panaC: `name` can be replaced with "label" maybe ? IDK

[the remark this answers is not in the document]
panaC: or maybe the same structure as a11y localization: `compact` and `descriptive`. It's a good naming I think

panaC:
@arthur-lemeur you should resolve conflict with the develop branch, I'm merged 2 PR today

src/renderer/library/components/settings/KeyboardSettings.tsx
```diff
@@ -389,11 +389,87 @@ class KeyboardSettings extends React.Component<IProps, IState> {
...
+                            }
+                        }
+                    }
+                    return css;
+                };
+
+                const cssStyles = getCssStyles();
+
+                const completeHtml = `
+        <!DOCTYPE html>
```
panaC, line 438: @arthur-lemeur  Is it really necessary to add this HTML export feature ? , I do not think this is what this PR is addressed. Moreover This is the job of the documentation!, there are no real benefits to maintain an html export. From my point of view.
If you really want to propose this feature, please create a new PR dedicated to this PR !

panaC: Okay, I do not think exporting an html file to the local filesystem is the good way to do it. Why do not integrate it in the keyboard shortcut GUI presentation like a view/edit mode ? We have to keep it simple ! At the end it's a codebase that we have to maintain.

gautierchomel: We considered it easier to export an HTML that already exists in the GUI than a Print/PDF generator and template.

panaC: A file need to be versioned, that means non-sens in this case when any keyboard shortcut can quickly be updated. What we can do instead it's to increase the user experience for the keyboard shortcut FROM thorium. 
Think that a file need to be localised, it's OS centric, can have many copy, can be stored in many place in the disk, and finally the user need to deal with it (open and load the html file in the browser), just for the purpose of a better keyboard shortcut presentation.

panaC: We have to look from the electron API, perhaps we can try to open a new BrowserWindow attached to is parent.

panaC: I reverted the commit and push to a new draft PR https://github.com/edrlab/thorium-reader/pull/2949 

Thanks @gautierchomel for your discussion, Please let's continue the discussion in the PR.

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
...
+                "AudioPlayPause": {
+                    "description": "in reader window, calls the \"play\" or \"pause\" toggle commands for audio (TTS or EPUB3 Media Overlays readaloud).",
+                    "name": "Play/Pause Audio"
+                },
+                "AudioPrevious": {
+                    "description": "in reader window, calls the \"previous\" command for audio (TTS or EPUB3 Media Overlays readaloud, or \"rewind\" for audiobooks).",
+                    "name": "Previous Audio"
+                },
+                "AudioPreviousAlt": {
+                    "description": "in reader window, calls the \"previous\" command for audio (TTS or EPUB3 Media Overlays readaloud, or \"rewind\" for audiobooks). Alternative shortcut key.",
```
panaC, line 891: This is not really Alternative way, I think. `Alt` shortcut for audio previous/next enable escape boolean in `ttsNext` / `ttsPrevious` r2-nav function. see : https://github.com/readium/r2-navigator-js/blob/9724ef3d4523e956e5e616d91f9f10872b4c02a7/src/electron/renderer/webview/readaloud.ts#L434

@danielweck Maybe can you improve this shortcut description ?

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
             "disclaimer": "Shortcuts were defined from QWERTY keyboard layout.",
             "editUserJson": "Edit user file...",
             "keyboardShortcuts": "Keyboard shortcuts",
+            "list": {
+                "AddBookmarkWithLabel": {
+                    "description": "",
+                    "name": "Add Bookmark with Label"
+                },
+                "AnnotationsCreate": {
+                    "description": "When a text is selected, it will be marked as annotation and the annotation dialog box will open to allow additional text and select of color and stroke.",
```
panaC, line 863: Perhaps we need to improve this long description for the `annotationCreate`, what does it means : "It will be marked as annotation` !?

panaC: The annotation (aka Note) will be saved after annotation popup submitted

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
...
+                "FocusReaderNavigationTOC": {
+                    "description": "in reader window, moves the keyboard focus to the navigation panel, but more specifically into the table of contents, and if possible at the detected heading / closest to the reading location.",
+                    "name": "Focus on Reader Navigation TOC"
+                },
+                "FocusReaderSettings": {
+                    "description": "in reader window, moves the keyboard focus to the settings panel (user preferences for displaying publication content).",
+                    "name": "Focus on Reader Settings"
+                },
+                "FocusSearch": {
+                    "description": "Available from the library and the reading winfos. In library window, moves the keyboard focus to the search text input of local bookshelf or OPDS feed. In reader window, moves the keyboard focus to the publication-wide search text input.",
```
panaC, line 951: typo: `winfos`

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
...
+                "NavigateNextLibraryPage": {
+                    "description": "In the library view, display the next set of books.",
+                    "name": "Go to Next Library Page"
+                },
+                "NavigateNextLibraryPageAlt": {
+                    "description": "In the library view, display the next set of books (alternative keyboard shortcut).",
+                    "name": "Go to Next Library Page (Alternative)"
+                },
+                "NavigateNextOPDSPage": {
+                    "description": "in library window for Catalogues (OPD) feeds, activates the next pagination links.",
```
panaC, line 979: typo: `opd`

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
...
+                    "name": "Go to Next OPDS Page"
+                },
+                "NavigateNextOPDSPageAlt": {
+                    "description": "in library window for Catalogues (OPD) feeds, activates the next pagination links (alternative keyboard shortcut).",
```
panaC, line 983: typo: `opd`

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
...
+                "NavigatePreviousHistory": {
+                    "description": "in reader window, go back or forward in the navigation history: Beginning or End of the publication.",
+                    "name": "Go to Previous History"
+                },
+                "NavigatePreviousLibraryPage": {
+                    "description": "In the library view, display the previous set of books.",
+                    "name": "Go to Previous Library Page"
+                },
+                "NavigatePreviousLibraryPageAlt": {
+                    "description": "In the library view, display the previous set of books (alternative keyboard shortcut).",
```
panaC, line 1011: In my memory : For the navigation previous/next `Alt` is not an `alternative` way but a quick way to access first and last page

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
...
+                    "name": "Go to Previous Library Page (Alternative)"
+                },
+                "NavigatePreviousOPDSPage": {
+                    "description": "in library window for Catalogues (OPD) feeds, activates the previous pagination links.",
```
panaC, line 1015: typo: `opd`

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
...
+                    "name": "Go to Previous OPDS Page"
+                },
+                "NavigatePreviousOPDSPageAlt": {
+                    "description": "in library window for Catalogues (OPD) feeds, activates the previous pagination links (alternative keyboard shortcut).",
```
panaC, line 1019: typo: `opd`

panaC: same as library page navigation, It is a previous/next history and a go to first/last history link for the alternative version

src/resources/locales/en.json
```diff
@@ -854,9 +854,229 @@
...
+                "NavigateToBegin": {
+                    "description": "in reader window, goes to the beginning of the publication.",
+                    "name": "Go to Beginning"
+                },
+                "NavigateToEnd": {
+                    "description": "in reader window, goes to the end of the publication.",
+                    "name": "Go to End"
+                },
+                "OpenReaderInfo": {
+                    "description": "in reader window, publication info popup modal dialog (may be used in future Thorium version for similar feature in library window).",
```
panaC, line 1039: `may be used in future Thorium version for similar feature in library window` wtf !? :)

panaC:
@gautierchomel @danielweck @arthur-lemeur We have to finish the keyboard shortcut localization work before to merge the PR.

src/resources/locales/en.json
gautierchomel: fixed

src/resources/locales/en.json
gautierchomel: fixed

src/resources/locales/en.json
gautierchomel: fixed

PR #2944, opened 2025-04-22
fix(notes): import notes

panaC:
- remove annotationImportQueue FIFO queue reducer
- replace with INoteState -> readiumAnnotation.import.target and then converted to locatorExtended on renderer reader start.


src/main/redux/sagas/note.ts
```diff
@@ -176,9 +211,9 @@ function* importAnnotationSet(action: annotationActions.importAnnotationSet.TAct
                 continue;
             }
 
-            const annotationParsed: INotePreParsingState = {
+            const annotationParsed: INoteState = {
                 uuid,
-                target: incommingAnnotation.target,
+                index: -1, // TODO !!!!
```
panaC, line 216: need fixing in another PR

PR #2959, opened 2025-04-30
fix: resourceCache optimisation (Fixes #2840)

### LINKED ISSUE #2840: publication document cache is eager, should be lazy? (significant performance penalty and memory usage when opening a large EPUB)
In the console, I see 300 occurences of `requestPublicationData ISearchDocument`, one for each spine item (they are LCP-decrypted by the way, but the runtime cost is mostly DOM parsing).

panaC:
Fixes #2840


src/common/redux/states/renderer/search.ts
```diff
@@ -5,7 +5,7 @@
 // that can be found in the LICENSE file exposed on Github (readium) in the project repository.
 // ==LICENSE-END==
 
-import { ISearchResult } from "readium-desktop/utils/search/search";
+import { ISearchResult } from "readium-desktop/renderer/reader/redux/sagas/search/search";
```
panaC, line 8: Not great at all :( 

I'm quickly fixing that !

danielweck: Our build process now better checks for when a JS bundle links to code it shouldn't contain, so this case would have been caught if this file was imported/required in the library/pdf/main bundles (directly or transitively). However we are not yet capable of ensuring at build time that the "src/common" code doesn't reference non-common code (such as reader or library-specific code). TODO

https://github.com/edrlab/thorium-reader/commit/9951e48ddd5c59ba4a2d7c9923cb4dba80b2a812#diff-1cf8fd631d850ded4f1d2b46f6718567fadaab504363c3268e0b087fc836ff98

PR #2965, opened 2025-05-13
PDF Print

panaC:
PDF print features : 
- Print Popover from ReaderHeader.tsx
- thumbnail request to pdf.js webview process with IPC BusEvent
- thumbnail caching under a string[] state in Reader.tsx
- thumbnail virtualised with Tanstack Virtual
- multiple page ranges support with thumbnail rendering
- LCP print information in publicationInfo dialog and at the bottom of the print popover
- LCP print (limit) support
- Print command dispatched to main process with LCP check, then reply to the mainWindow sender with the print page range and finally send print request to the pdf.js webview process. 
- Print request (shortcut) is disabled in pdf.js edrlab build, only thorium can request the print.
- Ctrl-P keyboard shortcut

Need to be fixed/updated from the UI/UX point of view : 
- Popover should be a modal dialog : it's still possible once the popover is open to navigate in the pdf.js viewer. pdf.js page navigation during thumbnail rendering, breaks it. Once broken there are no possibility to re-request the thumbnail rendering to the thorium process. A dialog modal should mitigate this issue, and I don't think it's necessary to let the user navigate to the pdf whereas the print popover/dialog is opened.


panaC:
https://github.com/user-attachments/assets/3c66b88d-5665-451f-ac31-7bab8096845f

PR #2994, opened 2025-06-10
feat: Add automatic text selection annotation popup

LFM097384:
- Create new TextSelectionPopup component that automatically shows when text is selected
- Integrate popup with existing AnnotationEdit component for annotation creation
- Add component to Reader.tsx to enable automatic annotation popup functionality
- Position popup near selected text using DOM selection coordinates
- Connect to Redux state for selection info and annotation configuration

### DIFF
--- a/src/renderer/reader/components/TextSelectionPopup.tsx
+++ b/src/renderer/reader/components/TextSelectionPopup.tsx
@@ -1,0 +1,189 @@
+// ==LICENSE-BEGIN==
+// Copyright 2017 European Digital Reading Lab. All rights reserved.
+// Licensed to the Readium Foundation under one or more contributor license agreements.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file exposed on Github (readium) in the project repository.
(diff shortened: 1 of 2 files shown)


src/renderer/reader/components/TextSelectionPopup.tsx
```diff
@@ -0,0 +1,189 @@
...
+        } else if (!hasSelection && hadSelection) {
+            // Selection was cleared
+            console.log('Hiding popup - selection cleared');
+            this.previousSelectionText = null;
+            this.hidePopup();
+        }
+    }    private showPopup = () => {
+        console.log('showPopup called');
+        
+        // Try to get position from current DOM selection first
```
panaC, line 72: You cannot get the current DOM position of the selection from the reader, host of the navigator webview: 

see this: https://github.com/edrlab/r2-navigator-js/blob/48eea2f8bc8f00f666b32506c32c2bbdfb8f9bda/src/electron/renderer/dom.ts#L295

PR #3018, opened 2025-06-24
up print menu according to figma design and up labels

### LINKED ISSUE #3012: [CSS] PDF Print Move from Popover To Dialog modal Cf Figma
Image

arthur-lemeur:
Fixes #3012


panaC (changes requested):
dialog modal size too narrow on my machine : 

image

image

The cssSelector name seems to not be changed, see : `popover_dialog_reader` whereas it's a modal dialog

panaC (changes requested):
Look, bottom margin on the dialog modal

image

the bottom section on each cover seems clipped for me, maybe you can add a thin margin 

Screenshot 2025-06-26 at 13 22 15

need to apply a css text-overflow rule and max-width on lcp section.
Need to fallback on a compressed view like "1 to 999" for example

src/renderer/reader/components/Print.tsx
```diff
@@ -146,6 +146,25 @@ export const PrintContainer = ({ pdfPageRange, pdfThumbnailImageCacheArray }: {
         overscan: 3,
     });
 
+    function formatPrints(prints: number[]) {
```
panaC, line 149: You cannot do that, [1-50] was an example with fifty pages "contigu". We need an algorithm that compress the page array of "contigu" pages.

PR #3027, opened 2025-06-26
Do not mark Traditional Chinese (`zh-Hant` and `zh-TW`) as a RTL language

### LINKED ISSUE #1987: Incorrect rendering of fixed-layout documents with certain RTL languages
For fixed layout documents, under RTL page progression order, if XHTML pages of a document contain an `xml:lang` property with a value that causes [this function](https://github.com/edrlab/thorium-reader/blob/74d5b31c10d8d21ecb98eae8edecfd187c19b713/src/renderer/common/language-string.ts#L36) to return true, the viewport rendering of the document only renders a sliver of the right side (decreasing in width as you zoom out) of the page.

This issue can be encountered naturally/"in the wild" with EPUB3 fixed-layout comics in traditional Chinese (or otherwise have been tagged as such) that readily pass `epubcheck` against EPUB 3.3.

Tested under Thorium version 2.3.0 on Windows, Arch Linux, and Ubuntu (fresh installs in a VM).

Demonstration with a minimally reproducing epub (for sake of example, each page's `xml:lang` property was set to `zh-TW`):
---
### Cover page (or any page with `rendition:page-spread-center`):
Thorium-cover-zh-TW
### A landscape spread of 2 pages:
Thorium-spread-zh-TW

Comparison with correct rendering (either with no set `xml:lang` or `xml:lang` set to something that doesn't count as RTL):
---
### Cover page (or any page with `rendition:page-spread-center`):
Thorium-cover-en-US
### A landscape spread of 2 pages:
Thorium-spread-en-US

---

The relevant EPUB files are linked here:
[Sample-EPUBs.zip](https://github.com/edrlab/thorium-reader/files/12448534/Sample-EPUBs.zip)

initramfs:
Thorium currently marks the IETF BCP 47 tags of `zh-Hant` and `zh-TW` as RTL scripts for some reason when those scripts are decided not.

[This w3.org article](https://www.w3.org/International/questions/qa-scripts) lists the scripts written in RTL, where Chinese, Japanese, and other east Asian logographic languages are not present. Except for a carve-out in the in the "Historical scripts" section:

> In addition to the usual candidates for right-to-left text direction (Arabic, Hebrew, N'Ko, Syriac, Thaana, etc.), a number of other writing systems that are nowadays written left-to-right (LTR) were once written in a right-to-left (RTL) direction. These writing systems include **Chinese, Japanese**, Egyptian hieroglyphs, Tifinagh, and Old Norse runes. See [RTL rendering of LTR scripts](https://www.w3.org/International/questions/qa-ltr-scripts-in-rtl) for a discussion of how to produce RTL text for these scripts.

_(emphasis mine)_

Documents using these scripts should have their [dir](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/dir) tag set as `rtl` and should be rendered accordingly.

Expanding more on the [dir](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/dir) tag, the `auto` value has the following description:

> `auto`, which lets the user agent decide. It uses a basic algorithm as it parses the characters inside the element until it finds a character with a strong directionality, then applies that directionality to the whole element.

If we inspect an arbitrary character such as [鐵](https://util.unicode.org/UnicodeJsps/character.jsp?a=9435) (Traditional Chinese, `zh-Hant`), we can see the `Bidi_Class` is `Left_To_Right`.

Indeed if we look at similar characters: [铁](https://util.unicode.org/UnicodeJsps/character.jsp?a=94C1) (Simplified Chinese, `zh`) and [鉄](https://util.unicode.org/UnicodeJsps/character.jsp?a=9244) (Japanese, `ja`) we see the same `Bidi_Class` of `Left_To_Right`. Essentially there should be no treatment difference between those 3 scripts.

As I partially explained in https://github.com/edrlab/thorium-reader/issues/1987#issuecomment-2233586863, the "RTL" nature is only true of columns of characters (governed by the CSS property [writing-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode)). In essence, the aforementioned east Asian scripts (as well as others like Hangul) are all considered LTR scripts that can have a `writing-mode` property of [horizontal-tb](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode#horizontal-tb) where they render much like English **or** [vertical-rl](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode#vertical-rl) where they render vertically in columns that go right-to-left (**but HTML `dir="ltr"`**).

In other words, the "script" itself has an intrinsic direction of LTR (defined essentially as the writing direction when rendered horizontally) but two `writing-mode`/rendering axes of `horizontal-tb` and `vertical-rl`.

Setting `writing-mode` to `vertical-rl` **and** `dir` to `rtl` yields incorrect results where the text is bottom-aligned and the full stop 『。』 is placed before the sentence.

---

This PR removes `zh-Hant` and `zh-TW` from the list of languages considered to be RTL which fixes rendering for Traditional Chinese documents.

Note: this PR **does not** fix #1987 since that is a bug specific to RTL image display. By correctly classifying Traditional Chinese as LTR, #1987 no longer triggers for Traditional Chinese documents but will still trigger for Arabic/Hebrew documents.

### DIFF
--- a/src/common/language-string.ts
+++ b/src/common/language-string.ts
@@ -104,7 +104,5 @@ export function convertMultiLangStringToLangString(items: string | IStringMap | 
 export function langStringIsRTL(lang: string): boolean {
     return lang === "ar" || lang.startsWith("ar-") ||
         lang === "he" || lang.startsWith("he-") ||
-        lang === "fa" || lang.startsWith("fa-") ||
-        lang === "zh-Hant" ||
-        lang === "zh-TW";
+        lang === "fa" || lang.startsWith("fa-");
 }
--- a/src/renderer/reader/components/ReaderMenu.tsx
+++ b/src/renderer/reader/components/ReaderMenu.tsx
@@ -135,19 +135,15 @@ const isRTL = (r2Publication: R2Publication) => (_link: ILink) => {
                 [r2Publication.Metadata.Language]) :
             [] as string[];
         isRTL = lang.reduce<boolean>((pv, cv) => {
-            const rtlExcludingJapanese = typeof cv === "string" ?
-                // we test for Arabic and Hebrew,
-                // in order to exclude Japanese Vertical Writing Mode which is also RTL!
+            const rtlLang = typeof cv === "string" ?
                 // see langStringIsRTL()
                 (
                     cv === "ar" || cv.startsWith("ar-") ||
                     cv === "he" || cv.startsWith("he-") ||
-                    cv === "fa" || cv.startsWith("fa-") ||
-                    cv === "zh-Hant" ||
-                    cv === "zh-TW"
+                    cv === "fa" || cv.startsWith("fa-")
                 ) :
                 false;
-            return pv || rtlExcludingJapanese;
+            return pv || rtlLang;
         }, false);
     }
     return isRTL;


danielweck:
Thank you for reporting this error.

@jaypanoz I remember adopting the guidance for "page progression direction" from ReadiumCSS, which I must have misinterpreted (apologies):

* https://github.com/readium/css/blob/0728ade3893ba9c07ae6cd64798e5ded7b37d777/docs/CSS16-internationalization.md?plain=1#L90-L91
* https://github.com/readium/css/blob/0728ade3893ba9c07ae6cd64798e5ded7b37d777/docs/CSS03-injection_and_pagination.md?plain=1#L243-L246
* accessorily: https://github.com/readium/css/blob/0728ade3893ba9c07ae6cd64798e5ded7b37d777/css/src/modules/ReadiumCSS-base.css#L188-L209

danielweck:
Related:

* https://github.com/edrlab/r2-navigator-js/pull/88
* https://github.com/edrlab/r2-shared-js/pull/60

danielweck:
Related: https://github.com/edrlab/thorium-reader/issues/1987

danielweck:
@initramfs could you please share one or two publications that exhibit Thorium's bug (EPUB fxl, and EPUB reflow), so I can validate the fix in scroll mode as well as paginated mode. The edge case I am concerned about is inter-document progression with vertical writing mode (top to bottom + RTL) in each individual document (because of missing page-progression-direction in some EPUBs, as well as missing dir attribute or CSS text-direction on HTML or BODY, and of course vertical-writing-mode inconsistencies too). It would really help if you could provide real-world publications that are typical examples of published EPUBs (good or bad).

JayPanoz:
Thanks for the review of docs, that is massively helpful. 

@danielweck I guess we should probably rewrite the CSS docs so that it is crystal clear on first glance. Perhaps providing a clear path of decision for each language/script. Could even be an interactive form actually. 

Right now we have things in different places and I can understand how confusing it can become.

danielweck:
Thank you so much!

danielweck:
A simple fix, but thank you so much for taking the time to explain our error! 👍

PR #3051, opened 2025-06-30
Dev/home slider reset

### LINKED ISSUE #3037: horizontal strip of publications can only be panned with buttons with arrow icons in the upper right corner, should scroll so mouse wheel etc. can be used
In this screenshot I just added overflow scroll X auto, to see how it feels...I like it (but of course the horizontal scroll bar occupies some GUI space, and the buttons left/right logic actually interferes now)

Image

arthur-lemeur:
Fixes #3037


danielweck (changes requested):

src/renderer/assets/styles/components/slider.scss
```diff
@@ -8,8 +8,12 @@
 
   &_wrapper {
     position: relative;
-    overflow: hidden;
+    overflow: scroll;
```
danielweck, line 11: There is a 1px scroll offset vertically, please use `overflow-x: hidden` and `overflow-y: scroll`

danielweck:
merge conflicts :)

arthur-lemeur:
I'll try to implement this solution this afternoon

PR #3061, opened 2025-07-07
fix bad handling of disabled buttons

arthur-lemeur:
"this.wrapperRef.current" was "undefined" at first render and so the condition handling the disabled "slide right" button was broken.


danielweck (changes requested):
I think we need to comment out the forceUpdate() logic and the updateScrollPosition() "scroll" event callback as the fine-grain disableLeft/Right state now does a better job.

PR #3076, opened 2025-07-23
fix: properly handle protocol URLs and file associations on Windows

bjnoel:
- Fix protocol registration to specify electron path and app path in development mode
- Add URL detection in second-instance handler to capture Windows command-line arguments
- Extract protocol URLs (opds://, thorium://, http://, https://) from argv array
- Improve OPDS feed title extraction using hostname or path segments
- Remove unnecessary initialization delay in OPDS URL processing

  This fixes two related issues on Windows:
  1. OPDS protocol URLs (opds://) not being imported when clicked in browsers
  2. LCPL files not opening when double-clicked while Thorium is running

  Both issues were caused by Windows command-line arguments not being properly
  passed to the second instance handler. The fix ensures argv is correctly
  processed for both protocol URLs and file paths.

  Fixes the issue where clicking opds:// links would open Thorium but fail to
  import the catalog. Also fixes the issue where double-clicking .lcpl files
  would focus Thorium but not open the file.

### DIFF
--- a/src/main/cli/lock.ts
+++ b/src/main/cli/lock.ts
@@ -80,6 +80,19 @@ export function lockInstance() {
             debug("Someone tried to run a second instance, we should focus our window", argv);
             debug("#####");
 
+            // On Windows, protocol URLs are passed as the last argument
+            // Look for opds:// or thorium:// URLs in the argv
+            let protocolUrl: string | undefined;
+            for (let i = argv.length - 1; i >= 0; i--) {
+                const arg = argv[i];
+                if (arg && (arg.startsWith("opds://") || arg.startsWith("thorium://") || 
+                           arg.startsWith("http://") || arg.startsWith("https://"))) {
+                    protocolUrl = arg;
+                    debug("Found protocol URL in argv:", protocolUrl);
+                    break;
+                }
+            }
+
             // Someone tried to run a second instance, we should focus our window.
             debug("comandLine", argv, _workingDir);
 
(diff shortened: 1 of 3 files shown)


danielweck:
very interesting, thank you for taking the time to troubleshoot and propose a PR.
we will look into this asap, we need to run tests on Mac and Linux to ensure no regressions.

danielweck:
Could you please merge `develop` into your branch, you will have conflicts but please simply adopt:

https://github.com/edrlab/thorium-reader/commit/39d736f56721ad1b391865b94b96c817e5d7d0d0

As you can see I copy/pasted your code, thank you very much! :)

bjnoel:
Hey @danielweck - I've force pushed a change to my branch to bring it up to date to develop.

Windows passes protocol URLs and file paths as command-line arguments to the second instance, but these were not being properly extracted from `argv`.

  **Fixed by:**
  - Detecting protocol URLs (`opds://`, `thorium://`, `http://`, `https://`) in the second-instance handler's argv
  - Routing detected URLs directly to `setOpenUrl()` instead of through the generic command-line entry point
  - Using protocol constants from `streamerProtocol.ts` instead of hardcoded strings

  **Additional improvements:**
  - Enhanced OPDS feed titles to show hostname instead of full URL
  - Improved debug logging for troubleshooting

  ## Testing
  Tested on Windows by:
  - Double-clicking `.lcpl` files → now opens in Thorium
  - Clicking `opds://` links in browser → now opens in Thorium

  **Files changed:**
  - `src/main/cli/lock.ts` - argv protocol detection
  - `src/main/redux/sagas/event.ts` - improved OPDS feed handling

PR #3144, opened 2025-09-11
fix: note persistence switched to sqlite storage instead of json reduxState

panaC:
- the notes array from the reader publication registry are migrated to the sqlite db under the table `note_experimental` 
- both actions add/update&remove are catched in reduxSagas and redirected to sqlite INSERT/UPDATE/DELETE functions
- reader win reduxState hydrated by IPC as before but now from the sqlite db
- data synchronization still relies on redux IPC
- backward compatibity with thorium 3.2, saving to json registry state when closing the reader window


gautierchomel (approved):

src/main/db/sqlite/note.ts
```diff
@@ -0,0 +1,119 @@
...
+            pub_id TEXT NOT NULL,
+            note_id TEXT NOT NULL UNIQUE,
+            note_json TEXT NOT NULL
+            ) STRICT
+        `);
+    } catch (e) {
+        debug("SQLITE INIT TABLE NOTE ERROR !!!");
+        debug(e);
+
+        throw new Error("Cannot instanciate sqlite database with the table 'note_experimental'");
```
danielweck, line 28: `note_experimental` here too

danielweck (changes requested):

danielweck:
Can we rely on some sort of "last modified" timestamp on JSON-persisted notes? New-Thorium can check dates, and if SQL DB is stale (i.e. date behind JSON) then the app can infer that JSON was persisted by old-Thorium and in this case notes DB must be reconstructed entirely by migrating the JSON forward.

PR #3192, opened 2025-10-01
fix opds publications pagination when publications and groups

### LINKED ISSUE #3189: [opds][css] opds pagination transparency/margin when not enabled
Image

tested with https://test.opds.io/2.0/home.json

arthur-lemeur:
Fixes #3189

### DIFF
--- a/src/renderer/library/components/opds/PageNavigation.tsx
+++ b/src/renderer/library/components/opds/PageNavigation.tsx
@@ -73,7 +73,9 @@ class PageNavigation extends React.Component<IProps, undefined> {
         const element = this.fixedElementRef.current;
         const container = document.getElementById("opds_browserResults");
         const cardWrapper = document.getElementById("card_wrapper");
-        if (cardWrapper.scrollHeight >container.clientHeight) {
+        const sections = document.getElementsByTagName("section");
+        const sectionsHeightSum = Array.from(sections).reduce((sum, section) => sum + section.offsetHeight, 0);
+        if ((cardWrapper.scrollHeight + sectionsHeightSum) >container.clientHeight) {
           element.style.position = "unset";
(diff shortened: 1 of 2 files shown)


src/renderer/library/components/opds/PageNavigation.tsx
```diff
@@ -73,7 +73,9 @@ class PageNavigation extends React.Component<IProps, undefined> {
         const element = this.fixedElementRef.current;
         const container = document.getElementById("opds_browserResults");
         const cardWrapper = document.getElementById("card_wrapper");
-        if (cardWrapper.scrollHeight >container.clientHeight) {
+        const sections = document.getElementsByTagName("section");
```
danielweck, line 76: This seems like a very fragile approach, what happens if at some point in the future another `section` element is added in the DOM for a totally unrelated purpose?

Also, can the `opds_browserResults` and `card_wrapper` element IDs not be hard-coded as strings? I am concerned that this code is currently not compiler-checked so any future potential refactoring of CSS selectors and/or DOM structure will silently break this code (probably in other places as well, I haven't checked).

danielweck (changes requested):

src/renderer/library/components/opds/PageNavigation.tsx
danielweck: The problem is how fragile code like this is when future JSX evolutions refactor the DOM. Anyway, let's merge and track the issue: https://github.com/edrlab/thorium-reader/issues/3237

arthur-lemeur: Another way to handle this would be I think to pass a boolean (are there groups or not in this page) in the EntryPublicationList component, do you think it would be better?

PR #3199, opened 2025-10-06
change catalog design for profile publications slider and displayed c…

arthur-lemeur:
change catalog design for profile publications slider and displayed catalogs in libraryheader


src/renderer/library/components/catalog/Catalog.tsx
```diff
@@ -69,6 +69,7 @@ class Catalog extends React.Component<IProps, undefined> {
         const secondaryHeader = <span style={{ display: "flex", justifyContent: "end", alignItems: "end", height: "53px", borderBottom: "1px solid var(--color-verylight-grey-alt)", paddingBottom: "30px" }}><PublicationAddButton /></span>;
 
         const customizationProfileProvision = this.props.customizationProvision.find(({id}) => this.props.customizationProfileId === id);
+        const profileTitle = customizationProfileProvision?.title && typeof customizationProfileProvision.title === "object" ? customizationProfileProvision.title[this.props.locale] || customizationProfileProvision.title["en"] || "" : "";
```
danielweck, line 72: Should the fallback string not be empty but `this.props.__("catalog.customization.publications")` instead? (I haven't tested the code, just reading / reviewing it...so I could be totally wrong, sorry :)

danielweck (changes requested):

PR #3222, opened 2025-10-10
fix: HTTP requests fencing with isURL()

danielweck:


panaC:
Why is `localhost` restricted, but not `127.0.0.1`?

I’m running into this while testing a local LCP server for PR #3642.

PR #3244, opened 2025-10-17
fix(OPDS): allow same-origin navigation in authentication window

bjnoel:
The OPDS authentication window's will-navigate handler was blocking legitimate navigation within the same origin, causing authentication flows to open in external browsers instead of completing within Thorium.

Changes:
- Extract the origin from the initial authentication URL
- Allow navigation to any URL with the same origin (scheme + host + port)
- Preserve existing security checks for opds://authorize/ callbacks
- Maintain protection against cross-origin navigation

This fixes authentication flows that involve multiple redirects or form posts within the same domain (e.g., login page → authorize page → callback).

Previously allowed:
- opds://authorize/ callbacks
- Navigation to exact same URL

Now also allows:
- Navigation within same origin (e.g., auth.example.com/login → auth.example.com/authorize)

Fixes authentication issues where OAuth providers use multi-step flows within their authentication domain.

### DIFF
--- a/src/main/redux/sagas/auth.ts
+++ b/src/main/redux/sagas/auth.ts
@@ -819,6 +819,16 @@ function createOpdsAuthenticationModalWin(url: string): BrowserWindow | undefine
 
     win.loadURL(url);
 
+    // Extract the origin from the initial auth URL to allow navigation within same domain
+    let initialOrigin: string | undefined;
+    try {
+        const initialUrl = new URL(url);
+        initialOrigin = initialUrl.origin;
+        debug("Auth window initial origin:", initialOrigin);
+    } catch (e) {
+        debug("Failed to parse initial auth URL origin:", e);
+    }
+
     const willNavigate = (navUrl: string | undefined | null) => {
 
         if (!navUrl) {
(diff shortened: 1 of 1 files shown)


bjnoel:
Yes, this works - thank you. Do you happen to know when the next official release will be?

danielweck:
PS: I should point out that this PR fix (as well as my adapted one) causes a "regression" in the sense that we really want the Electron BrowserWindow displayed by Thorium (as a child of the main library window) to only be used for OPDS OAuth ... unfortunately, now users can click on links and navigate the website to arbitrary pages, such as "create account", "recover password", "contact support" etc. This is problematic because Thorium is not a secure web browser, the HTTP(S) URL is not displayed to the user, there is no proper cookie handling, etc. etc.

danielweck:
> It would be great if you could create a test account for us in your system, is this something you could enable?

You mentioned this URL in a different thread, is that an endpoint we could us for testing?

https://gateway.ebooks.land/library/v1/library

danielweck:
@bjnoel @mpdunlop you both mentioned OAuth Authorization Code with PKCE, but unless I am mistaken "vanilla" Thorium only supports OPDS authentication with Implicit Grant. Do you maintain a fork of Thorium with added support for this OAuth extension? Sorry if I am getting this wrong :)

mpdunlop:
@bjnoel was mistaken - for Thorium (and other OPDS readers) we follow OPDS Authentication Document 1.0 specifications and use OAuth 2.0 Implicit Grant.

Our own internal applications use OAuth Authorization Code with PKCE.

bjnoel:
Yes, my mistake, apologies

PR #3256, opened 2025-10-21
replace the catalog header title previously set to the feed user title with the opds metadata.title

arthur-lemeur:


panaC:
no need to create a new action, you can add a `title` key in `IOpdsHeaderState` and dispatch it from `updateHeaderLinkWatcher` 

https://github.com/edrlab/thorium-reader/blob/85f4f2272189cf670b724c881c7cda6b8718c56a/src/common/redux/states/renderer/opds.ts#L15-L20

https://github.com/edrlab/thorium-reader/blob/85f4f2272189cf670b724c881c7cda6b8718c56a/src/renderer/library/redux/reducers/opds.ts#L59-L75

https://github.com/edrlab/thorium-reader/blob/85f4f2272189cf670b724c881c7cda6b8718c56a/src/renderer/library/redux/sagas/opds.ts#L94-L125

panaC:
Previously : 

image


Now: 


image

image

fallback on '-' when no metadata.title found or the opds feed is loading, no memorization of the previously last opds feed title

PR #3257, opened 2025-10-21
filter buy/borrow/subscribe link with supportedFileTypeLinkArray

### LINKED ISSUE #3032: Ignore book links with unknown media types
Currently, each link with `rel="http://opds-spec.org/acquisition/buy"` will be shown as a "Buy" button and processed via opening in a browser. That's confusing if the catalogue uses a custom media type with specific rules. As suggested in https://github.com/edrlab/thorium-reader/discussions/3031#discussioncomment-13595643, it is better to ignore such links.

The discussion is about buy links, and this is the most important case for our catalogue. However, maybe other types of links should be processed in the same way.

panaC:
Fixes #3032


panaC:
https://github.com/edrlab/thorium-reader/blob/e4014bad0edddbe6535ea81a0475dff47e448d9a/src/renderer/library/opds/handleLink.ts#L57-L59

There is already a fallback on shell.openExternal

PR #3292, opened 2025-11-22
fix(GUI): library covers opened book on dock icon tap

### LINKED ISSUE #3291: Dock click on macOS brings Library window to front even when a Reader window is open
On macOS, with both the Library and a Reader window open, clicking the app’s Dock icon raises the Library window above the Reader.

**Expected**: if a Reader window exists, Dock activation should restore/show that Reader (or the last active reader) rather than surfacing the Library and pushing the book behind.
**Current behavior**: Dock activation always prefers the Library, so the book window ends up hidden behind it.
Impact: disrupts reading flow; users must manually refocus the Reader after returning from another app. For app switching hotkey this works fine. Library stays behind.

**Steps to Reproduce**:

- Open Library window.
- Open a book (Reader window).
- Focus the Reader window.
- Switch to another app.
- Click the app icon in the macOS Dock.
- Library jumps to front, Reader stays behind.

**Proposed Fix**:

In the app.on('activate') path, prefer restoring/showing an existing Reader window; only fall back to Library when no reader is available.

**Environment**:
- macOS 15.x
- macOS 14.x

byte-me-studios:
Fixes #3291

### DIFF
--- a/src/main/redux/sagas/win/library.ts
+++ b/src/main/redux/sagas/win/library.ts
@@ -41,6 +41,22 @@ export function* appActivate() {
         error(filename_ + "appActivate", new Error("closing process not completed"));
     } else {
 
+        const readers = yield* selectTyped((state: RootState) => state.win.session.reader);
+        const readerWindows = ObjectValues(readers)
+            .map((reader) => getReaderWindowFromDi(reader.identifier))
+            .filter((w) => w && !w.isDestroyed() && !w.webContents.isDestroyed());
+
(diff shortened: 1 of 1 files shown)


panaC:
> I am wondering if this is an opportunity to review the library/reader window activation/restoration logic:

Yes we can
Let's open a dedicated issue for this

Thanks you very much @byte-me-studios, I appreciate your contribution, tested and seems more logic indeed.

I'm merging

PR #3330, opened 2025-12-18
Fix/debug chokidar

panaC:


src/main/customization/watcher.ts
```diff
@@ -31,9 +33,8 @@ export function customizationStartFileWatcherFromWellKnownFolder(wellKnownFolder
     const watcher = chokidar.watch(wellKnownFolder, {
```
danielweck: https://github.com/paulmillr/chokidar/blob/c6d772a92f30004fdc437d00ecd54fbed041f30e/src/index.ts#L460-L468

==>

https://github.com/paulmillr/chokidar/blob/c6d772a92f30004fdc437d00ecd54fbed041f30e/src/index.ts#L155

===>

https://github.com/paulmillr/chokidar/blob/c6d772a92f30004fdc437d00ecd54fbed041f30e/src/index.ts#L178-L180

danielweck: apparently we should be using `options.cwd` (second argument) and the first argument of the `watch()` function can be a glob?

panaC: Yes, thanks

Instead, why i need to replace backslash with slash if a utility function care of this : `normalizePathToUnix`

https://github.com/paulmillr/chokidar/blob/c6d772a92f30004fdc437d00ecd54fbed041f30e/src/index.ts#L180

https://github.com/paulmillr/chokidar/blob/c6d772a92f30004fdc437d00ecd54fbed041f30e/src/index.ts#L160

 https://github.com/paulmillr/chokidar/blob/c6d772a92f30004fdc437d00ecd54fbed041f30e/src/index.ts#L460

It should works, without cwd and wellknown folder conversion

PR #3351, opened 2026-01-14
feat(l10n): enable i18next plurals

### LINKED ISSUE #2660: i18n: Add support for plural forms

danielweck:
Fixes https://github.com/edrlab/thorium-reader/issues/2660

Follow-up to PR #2743
We need to adapt the i18n tooling (check, scan, sort, typed)

Also see:
https://github.com/felixmosh/i18next-locales-sync/issues/40
https://github.com/jwbay/i18next-json-sync/issues/22


danielweck:
`i18next` `compatibilityJSON` `v4`

I fixed Weblate (was `v3`)

https://github.com/edrlab/thorium-reader/blob/2f6b212db08a4ab928d887e258218e21eadf525d/src/common/services/translator.ts#L52

https://github.com/edrlab/thorium-reader/blob/2f6b212db08a4ab928d887e258218e21eadf525d/package.json#L78-L79

https://www.i18next.com/misc/json-format#i18next-json-v4

danielweck:
https://docs.weblate.org/en/latest/user/translating.html#plurals

https://www.i18next.com/translation-function/plurals

EN "one" "other"
https://github.com/edrlab/thorium-reader/blob/31f5ace3c2f551c03cb5f7809ae62130efe68df2/src/resources/locales/en.json#L142-L143

FR "one" "many" "other"
https://github.com/edrlab/thorium-reader/blob/31f5ace3c2f551c03cb5f7809ae62130efe68df2/src/resources/locales/fr.json#L142-L144

Weblate: https://hosted.weblate.org/translate/thorium-reader/thorium-reader-translation/en/?checksum=57a2578b136eb3c1#translations

Confirmed by i18next tooling:

https://jsfiddle.net/6bpxsgd4

i18next and Weblate are consistent with the specification:

https://cldr.unicode.org/index/downloads

PR #3355, opened 2026-01-17
Translations update for "About Thorium"

neitsmarr:
Original translation did not make any sence for native Russian speaker, so I removed it and made new one (en ->ru).


danielweck:
Apologies for the confusion and for wasting your time, these HTML files are not actually used in Thorium Desktop Reader anymore.

PR #3357, opened 2026-01-19
[PARKED] add text description when clicking on image in reader

arthur-lemeur:
Fixes ##3139


gautierchomel:
Great! 
To be discussed:
* "altText": "Alternative text" not being used, it can be commented. 
* Instead, the `Summary` of the `detail` element is filled with the content of the `alt` argument. That's great as `alt` would be very hard to explain in an human understandable way. But we should remember that no limit exist on the lenght of `alt`
* Also, the `Summary` should also describe the action that it triggers.
* If no `alt` or empty alt, then a generic "about": "About this image" is used. Correctly authored epub should never have content here, so in case of empty `alt` we would probably  end up with a "About this image" that will contain nothing.

Because of that I suggest to use a generic descriptive label (Expand or collapse textual descriptions). for the `Summary` and place (if it exists) the `alt` as the first `p` under.

danielweck:
could you please undo your changes to `package-lock.json`?

gautierchomel:
Apologises for mis manipulation ending with commit and revert commit being pushed here.

danielweck:
@gautierchomel I am afraid your commits broke the PR, there are changes in the Flox manifest, the GitHub Actions YAML, etc.

panaC:
@gautierchomel @arthur-lemeur What is the current status of this issue? The issue has been converted into a discussion.

danielweck:
This PR/branch contains hard-to-resolve merge conflicts, can this PR now be closed as obsolete or is it still relevant? (the PR code diff will remain in the GitHub records so we will be able to check back on the proposed changes, if necessary ... I just want to make sure that stale PRs with significant technical debt and no activity do not "pollute" the list of pending pull requests)

PR #3364, opened 2026-01-22
Add filters and sorting feature to all publication grid

arthur-lemeur:


danielweck:
When using the table (column/row) presentation of the data, its own filtering state is not in sync with this new state, causing lots of confusion.

danielweck:
I tested edge cases, there are still some inconsistencies but this very useful feature doesn't cause major UX issues and we need to get the translation rolling (Weblate is locked). Let's merge and iterate from here! Thank you

src/renderer/library/components/searchResult/AllPublicationPage.tsx
```diff
@@ -108,6 +110,9 @@ import * as OnGoingBookIcon from "readium-desktop/renderer/assets/icons/ongoingB
 import debounce from "debounce";
 import { useSelector } from "readium-desktop/renderer/common/hooks/useSelector";
 import { ICommonRootState } from "readium-desktop/common/redux/states/commonRootState";
+import { useTranslator } from "readium-desktop/renderer/common/hooks/useTranslator";
+import { HoverEvent } from "node_modules/@react-types/shared/src";
```
danielweck, line 114: oops, must fix this in develop!

danielweck:
`"Select Table Headers"` is not translated.

https://github.com/edrlab/thorium-reader/issues/3487

PR #3383, opened 2026-02-08
Refactoring: migrating r2-xxx-js package dependencies into Thorium Desktop

danielweck:


naglis:
IIUC this vendors several large dependencies into Thorium instead of pulling them via npm? Vendoring makes them harder to update (need to do it manually), track changes with upstream if there are modifications, verify integrity, etc. Could you please share why this is done?

danielweck:
Hello @naglis the `r2-xxx-js` Readium "2" packages are in fact specific to Thorium Desktop Reader (with a caveat explained below). The original intent was to provide a "Readium Desktop SDK" but in reality since the inception of this software project (i.e. before it was renamed to Thorium) the team has had to maintain these packages separately which has been an unnecessary development overhead that has continually caused headaches when aligning non-Readium transitive package dependencies with the Thorium stack (and build tools). Over the years we have gradually migrated significant amounts of code from some of these R2 packages into the Thorium source tree (we also deprecated non-trivial features such as the generic HTTP server which is now implemented using native Electron functionality). This PR finally completes the integration which will allow the team to iterate faster and improve the overall quality of the stack.

Now, there is a caveat: a command line feature in the `r2-shared-js` package was developed separately to provide a tool unrelated to Thorium. Similarly, the `r2-streamer-js` will continue to exist separately as it is used in a different context (Thorium Desktop Reader doesn't actually need it anymore, the entire code is commented out in this PR).

PR #3388, opened 2026-02-11
add specific locals to dock options labels

### LINKED ISSUE #3365: Make menu position labels expliciter
Received from external testing:
> The reading menu position buttons are labeled simply as "left", "right", and "default". It may be difficult for screen readers to determine the element associated with these controls without additional context.
> Table of Contents reading menu buttons in the top-right corner -  "Left", "Right", and "Default". We recommend updating the labels to describe both the function and the element being modified.  For example, "Move Navigation Panel - left", "Move Navigation Panel - right", "Move Navigation Panel - center" can be used to notify screen readers of the element associated with the controls.

Image

arthur-lemeur:
Fixes #3365


danielweck:
I fixed the locales issues. Merging now...

PR #3402, opened 2026-02-20
Reenable footnotes on TTS and MO stop

### LINKED ISSUE #3399: Popup annotations/footnotes stop working after TTS
I set up hidden `aside` footnotes and they are shown as expected when I click the corresponding `noteref`, but if I play synchronized audio (or even the built-in TTS if there is no SMIL) and then stop it, those clicks will no longer have any effect and no popup is shown. Weirdly, this also persists for one EPUB across Thorium restarts.

kerams:
Fixes https://github.com/edrlab/thorium-reader/issues/3399

I haven't built and tried this locally, but the issue seems to be straightforward - footnotes are disabled in the TTS and MO start handlers, but then never reenabled. `restoreAnnotationStateAfterTTSorMOStop` is called from both stop handlers, so resetting `noFootnotes` there should do the trick.

### DIFF
--- a/src/renderer/reader/components/Reader.tsx
+++ b/src/renderer/reader/components/Reader.tsx
@@ -3020,7 +3020,9 @@ class Reader extends React.Component<IProps, IState> {
             this.setState({ previousReaderConfigAnnotationDefaultDrawView: undefined });
         }
 
-        this.props.setConfig({ noFootnotes: false });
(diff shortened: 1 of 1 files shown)


danielweck:
Thank you very much for proposing a fix. I addressed the issue more completely via https://github.com/edrlab/thorium-reader/commit/d35632b7c4665104751b560c6c4c1a0cbed83da1

PR #3407, opened 2026-02-24
[DRAFT] revert redux state registry and session

panaC:
This commit introduce the revert of the redux win.session mechanism (currently broken) to save the current session (one library, multiple reader) at closing.

This remove also the reader registry persistence from the redux store, to rely on fileSystem publication directory index (appData/config-data/reader/<uuid>/...). Publication data preference will not be written in redux state in main process anymore, but instead written directly the disk. This allow to save massive data IO redux state persistence to disk by successive patches and final entire state dump.

next steps:

- migrate registry to appData config-data/reader/<uuid> directory
- handle library bound in appData/config-data by getter/setter functions in main/di
- read/write publication state bound/config/locator.json to appData config-data/reader/<uuid>


panaC:
replace by #3455

PR #3409, opened 2026-02-25
introduce publication-data & registry migration/hydration

panaC:
### Migration: 

win.registry[pubId]  ∪ publication.db[pubID]
- persisted to publication-data -> {config,locator,RTLFlip,bound}
- persisted to publication-storage -> {config,locator,RTLFlip}

### Runtime: 

Actions: 
- readerActions.setConfig
- readerActions.setLocator
- readerActions.disableRTLFlip
- readerActions.disableRTLFlip

are persisted to
- publication-data ->  {config,locator,RTLFlip,bound} -> WITHOUT DELAY
- publication-storage -> {config,locator,RTLFlip} -> WITH A DEBOUNCE OF 10s

publication-data is colocated with `config-data-json{-dev}` under `publication/<uuid>` folder
publication-storage is colocated with the stored publication under the default vault `%appData%/publications{-dev}/<uuid>/{config,locator,RTLFlip}` or an another directory if setting up by the user (not released yet)

- `bound.json` is not stored by publication-storage as it belongs to the appData folder

- `publication-data` handle information attached to the register publication and localized on appData
- `publication-storage` handle publication files `{book.epub,cover.png,manifest.json,locator.json,config.json,disableRTLFlip.json}` localized under a default vault which is in appData/publications{-dev}. A user vault chained with the default vault can be setting up and serve as an external publication vault in the chosen directory.

### DIFF
--- a/src/main/storage/publication-data.ts
+++ b/src/main/storage/publication-data.ts
@@ -96,7 +96,7 @@ export class PublicationData {
             debug("Publication-data destroy() done");
             debug(res);
         } catch (e) {
-            debug(e);
+            debug(`${e}`);
         }
     }
 
@@ -128,7 +128,7 @@ export class PublicationData {
                     try {
                         await fileHandle.close();
                     } catch (e) {
-                        debug(e);
+                        debug(`${e}`);
                     }
                     return; // already open
                 }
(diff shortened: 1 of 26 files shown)


src/main/storage/publication-data.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    public async write(pubId: string, type: TFileType, data: AnyJson) {
+        if (this.lock) return ;
+        assertUUIDv4(pubId);
+
+        this.assertAndGetFileName(type);
+
+        let file = this.filterFilesByType(type).find((a) => pubId === a.pubId);
+        if (!file) {
+            await this.open(pubId, type);
```
panaC, line 187: write call 2 times before pubId is open, means both fileHandle will be opened and added to files array, I think

PR #3412, opened 2026-02-27
fix: remove win persistence

panaC:
- disable win redux state diff patch
- remove win.session persistence and keep a subset of win.registry.reader. list of readers, to maintain the backward compatibility with 3.x if the next release is called 4.0


src/main/redux/store/memory.ts
```diff
@@ -481,6 +481,8 @@ export async function initStore()
                 // TODO: parallel !? libuv = 4 threads
                 const publicationData = diMainGet("publication-data");
                 const publicationStorage = diMainGet("publication-storage");
+
+                // TODO: fix: "config" | "locator" | "divina" | "disableRTLFlip" | "allowCustomConfig" | "noteTotalCount" | "pdfConfig"
```
panaC: TODO: missing persistence keys !

PR #3413, opened 2026-03-01
introduce watchdog

panaC:
- One second counter watchdog to config-data-json{-dev}/watchdog.json
- raise a dialog.showMessageBoxwhen watchdog file not removed when Thorium closes


panaC:
too old

PR #3421, opened 2026-03-04
update wizard content for version 3.4

### LINKED ISSUE #3404: disable or update the launch "wizard" (information is out of date)
For initial onboarding of new users or existing users (information about breaking and/or significant changes), I think what would work better is a simple scrolling Markdown with minimal amount of text and a hyperlink to the online documentation. The popup modal dialog would still have the "do not show again" checkbox, but there would not be complex "wizard"-like step-by-step tutorial, just a plain text document with basic formating for readability (headings, bullet points).

arthur-lemeur:
Fixes #3404


danielweck:
I think what is/was missing in this PR is a reset of the "do not show again" tickbox boolean check (if true, then reset to false in order to guarantee that users will see the 3.4.0 message). The easiest (but dirtiest) way to do this is to rename the flag (it's a JSON property in the Redux state / store). The more complex and error prone (but cleanest) way to do this is to compare Thorium Desktop versions, establish whether it's the first launch, and if so make sure to force "do not show again" to false. Any other ideas?

PR #3423, opened 2026-03-04
feat(persistence): split persisted redux state for 3.3.0 and 3.4.0 with forward/backward compatibility

panaC:
This pull request introduces a change in how application state is handled between versions 3.3.0 and 3.4.0 of Thorium Reader. The objective is to simplify the Redux state structure in 3.4.0 while maintaining full backward compatibility with 3.3.0.

In version 3.3.0, the persisted Redux state contains a deeply nested structure under win.registry.reader. This part of the state significantly increases the overall size of the serialized data and impacts performance during application startup and shutdown. In version 3.4.0, this nested reader registry data is no longer included in the main persisted state. Instead, reader-related data is loaded separately during state hydration through publication-data. This results in a flatter and more straightforward application state, improving performance and reducing complexity.

Because users may downgrade from 3.4.0 to 3.3.0, the full legacy state must still be preserved. To support this, two global state files are written when the application closes. The first is a complete state compatible with 3.3.0, including win.registry.reader, stored as `state.json`. The second is a derived version intended for 3.4.0, which is a copy of the full state with win.registry.reader removed, stored as `state_v340.json`. This ensures that 3.4.0 benefits from the optimized state structure while retaining the ability to restore the full legacy state if needed.

backup `state.${timestamp}.json` state is disabled when `state.runtime.js` + `state.patch.json` doesn't equal the json state.
Can be used in the next release, to restore lost publications.

After the reader registry migration process, both the full legacy state and the optimized 3.4.0 state are persisted. This guarantees that upgrading to 3.4.0 and downgrading back to 3.3.0 can occur without data loss or inconsistent state behavior.


panaC:
reader data hydration took between 10ms to 50ms to load, so for ~100 reader publications registered, it will be between 1s to 5s in the worst case
- In the next release, it will be optimized to down to 10-20ms max

panaC:
Tests matrix is: 

- v340 -> v340
- v340 -> crash -> v340
- v340 -> v330
- v340 -> crash -> v330
- v330 -> v330
- v330 -> crash -> v330
- v330 -> v340
- v330 -> crash -> v340

panaC:
Some observations: 

- v340 → v340
    
    - 52 publications imported
        
    - 14 in progress

        
- v340 → crash → v340
    
    - Librivox audiobook loaded
        
    - Locator timestamp preserved at 137s

        
- v340 → crash → v330
    
    - Librivox audiobook loaded
        
    - Locator timestamp preserved at 137s
        
    - Publication not displayed
        
    - Publication not present in state.json due to previous crash
        
    - Inconsistent state
        
- v330 → v340
    
    - v330 closed normally
        
    - state.json updated without reader registry
        
    - Migration from publication-data to memory not triggered
        
    - State replaced instead of merged
        
    - Publication not present in publications DB but stored on disk
        
    - Publication appears corrupted
        
    - Export/import restores publication but locator and config lost
        
    - Locator manually reset to 19:17
        
    - Data loss due to replace instead of merge
        
- v340 → v340
    
    - Audiobook locator not preserved
        
    - Act 4 – 21:40 not retained across restart
        
    - Seems to be a regression in 3.3.0
        
- v340 → v330
    
    - Audiobook loaded

        
- v330 → v330

    - Architecture PDF 150MB (initial state: page 332, zoom 200%)
    
- v330 → crash → v330
    
    - Publication corrupted in publications DB
        
    - File still present on disk
        
    - Export/import restores publication
        
    - Locator and config lost
        
    - Manually reset to page 228, zoom 300%
        
    - Crash corrupts DB entry
        
- v340 → crash → v330
    
    - Publication not loaded
        
    - Publication found in allPublications
        
    - Partial inconsistency
        
- v330 → crash → v340
    
    - Publication not loaded anymore
        
    - Touching state.json triggers migration in 3.4
        
    - Data not migrated from 3.4 to 3.3 lost
        
    - Migration after crash causes irreversible data loss

PR #3436, opened 2026-03-12
fix: apply hydration on reader config

panaC:


danielweck:
I wonder if `await diMainGet("publication-data")` could be cached as a "global" (javascript module scope) lazy-instantiated variable, to avoid coloring the "transformer" function as async / Promise-returning.

panaC:
> I wonder if `await diMainGet("publication-data")` could be cached as a "global" (javascript module scope) lazy-instantiated variable, to avoid coloring the "transformer" function as async / Promise-returning.

```
PublicationData.readJsonObj(pubId: string, type: TFileTypePubData): Promise<object | undefined>
```

PR #3450, opened 2026-03-18
fix unmount cycle of toast component

### LINKED ISSUE #3448: [ui] Toast notification wrapper remains non-empty after the toast is removed and blocks clicks on Settings
https://github.com/edrlab/thorium-reader/blob/af6386584fb2d67b24e09b3df67cec59df76210b/src/renderer/common/components/toast/Toast.tsx#L106

componentWillRemove -> componentWillUnmount

https://github.com/edrlab/thorium-reader/blob/af6386584fb2d67b24e09b3df67cec59df76210b/src/renderer/common/components/toast/Toast.tsx#L95

`handleTransitionEnd` can never properly unmount and remove the toast notifcation when the dom event not fired.

arthur-lemeur:
Fixes #3448

### DIFF
--- a/src/renderer/common/components/toast/Toast.tsx
+++ b/src/renderer/common/components/toast/Toast.tsx
@@ -51,6 +51,8 @@ export class Toast extends React.Component<IProps, IState> {
     private ref: React.RefObject<HTMLDivElement>;
     private timer: number | undefined;
     private ignoreTimer: boolean | undefined;
+    private fallbackTimer: number | undefined;
+    private isRemoving = false;
 
     constructor(props: IProps) {
         super(props);
@@ -71,6 +73,10 @@ export class Toast extends React.Component<IProps, IState> {
         if (this.timer) {
             window.clearTimeout(this.timer);
             this.timer = undefined;
+        }
+        if (this.fallbackTimer) {
+            window.clearTimeout(this.fallbackTimer);
+            this.fallbackTimer = undefined;
         }
         if (ignoreTimer) {
             this.ignoreTimer = true;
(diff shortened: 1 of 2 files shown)


panaC (changes requested):
Good, but doesn't fix the issue

danielweck:
> Good, but doesn't fix the issue

@panaC , @arthur-lemeur pushed another commit, does it work now?

src/renderer/common/components/toast/Toast.tsx
```diff
@@ -132,6 +126,7 @@ export class Toast extends React.Component<IProps, IState> {
 
         return (
             <div
+                onTransitionEnd={this.handleTransitionEnd}
```
panaC, line 135: If the event is not fired, the this.props.close function from the ToastManager component will never be called, meaning the toast will not be unmounted and its reference will not be removed.

danielweck:
@panaC will you have time early this week to confirm that this PR works? It would be great to fix the bug in 3.4.0

panaC:
@danielweck There is still an issue with the `transitionend` [event ](https://developer.mozilla.org/en-US/docs/Web/API/Element/transitionend_event) may not fire in some cases.
It would be more robust to handle all related transition events, including `transitionrun`, `transitionstart`, and `transitioncancel`.
Currently, the toast removal logic relies solely on the `transitionend` event. If this event is not triggered, a div overlay can block interaction with the settings button (except near its border).
To make this more reliable, I think it would be safer to include a fallback mechanism like a timeout, to ensure the toast state stays in sync with the UI.

panaC (changes requested):
See my last message.

panaC (approved):

PR #3451, opened 2026-03-18
fix the wrapping behavior of catalogs header

### LINKED ISSUE #3442: long OPDS feed label / title / name does not break (needs horizontal length constraint, ellipsis)
Image

arthur-lemeur:
Fixes #3442


danielweck:
Screenshot 2026-03-19 at 09 29 06
Screenshot 2026-03-19 at 09 29 23

danielweck:
In the screenshots above, I see 2 issues:
1) clipped vertical scrollbar
2) even when there is enough horizontal space to fit the name / title of the OPDS feed, there is a "line break" that pushes the search input field down

arthur-lemeur:
True, I forgot to remove a width value before pushing, thought I did, my bad.
These changes come with an inconsistency in the top margin of catalogs content between header on 1 line (40px of top margin) and header on 2 lines (0px of top margin). I was not able to find another solution for the moment which would still allow a truly flexible and responsive handling of the header.

Issue #3439, opened 2026-03-13
[win] Launch Thorium, double-click a publication in Explorer, double-click the same publication again, then close the Library window, the second reader remains open, detached in a dead state.


panaC:
1/ launch thorium
2/ double-click on a publication in explorer
3/ double-click again on a same publication
4/ close the library window
5/ see the library and the first reader is closed but not the second one (detached in a dead state)

PR #3464, opened 2026-03-23
review the reader browser window lifecycle in main process

panaC:
Features list: 
* Renamed `pubId` and `winId` for clarity and consistency
* Removed/Comment `win.registry` redux state/reducer/actions, no longer needed.
* Added `reader_pub` from synced actions (sender key)
* Refactored reader open/close logic to use saga generator actions instead of Redux actions
* Fixed and improved `winBound` behavior (new reader windows are offset and no longer overlap the library window)
* Added `readingFinished` handling
* Reviewed browser window startup lifecycle (reader initialization and closing flow)
* Replaced global reader Redux state persistence with atomic persistence via synced actions (`config` / `locator`)
* Removed the reader session feature (no longer saves and restores multiple reader states across restarts)


panaC:
supersede and close #3455

panaC:
Fixes #3439

PR #3467, opened 2026-03-25
fix: window bounds rectangle issues 

### LINKED ISSUE #3466: Rembember window state when closing application
This is not a bug report. The application window state isn't stored when the application is closed. The next time you open the application it defaults to its original position and size. While the width is okay, the height is too low, so I always have to resize it. It would be practical if it remembered its last state like most Windows 11 applications.

I'm using Windows 11 version 10.0.26100 and Thorium v3.3.0

panaC:
Fixes #3466 

**Change list:**

* set default window size to 1024×768, centered on the primary display (previously 800×600 positioned at one-third of the primary screen)
* ensure immutability by creating a copy of the window bounds object before modifying and saving it (fixes session reducer issue)
* Electron window bounds handling:

  * `getBounds()`: used to persist window size and position
* refactor `readerNewWindowBound` (`getWinBound`)

  * use Electron BrowserWindow instance bounds rectangle instead of the global Redux session state
  * on first publication open, initialize reader window position based on the library window position to target the active screen, with a position shift of 30px if possible
  * loop on the window bounds rectangle position until a free position is found with an offset of 30px
* use VS Code logic to normalize window bounds rectangle and to get the default display
* use VS Code logic to get the default display on single or multiple screens to compute the default rectangle

  * on macOS, new windows will be centered on the display with the current cursor

---

**Browser window bounds rectangle logic:**

**First start of Thorium:**

* window centred on the default display (active cursor position on macOS)
* window default size is 1024 x 768 if the display size allows it (on macOS, the display size excludes the top bar and dock, only the active display screen)
* window minimal resolution is 700 x 600 to target the minimal supported resolution of 800x600

**First opening of a publication:**

* will have the same size and position as the library window, before persistence of the bounds rectangle

**Multiple reader windows opened:**

* new reader window with the same position as current windows, will be shifted x/y of 20px until a free position is found
* macOS & Linux: these OSes seem to be pretty good at ensuring that a window is never outside of its bounds
* on MS Windows: it is possible to have a window with a size that makes it fall outside of the screen. The strategy is to try as much as possible to keep the window within the monitor bounds. We are not as strict as macOS and Linux and allow the window to exceed the monitor bounds as long as some pixels (128) of the window remain visible on the screen for the user to drag it back

**Single display:**

* strict x/y positioning, window bounds rectangle saved will be applied on the next start window, even if the window exceeds the monitor bounds (see above). On Linux and macOS, windows will not be truncated; only Windows allows it. No normalization is done when saving, only when the bounds are applied to a new BrowserWindow

**Multiple display:**

* window bounds rectangle normalization will try to restore the window position on the original display screen. If the restoration fails and the display is not available, the window position will either be on the intended display with the desired width/height if possible, or, as a fallback, on the default screen with the default size
* macOS: maximum width/height of the second display resolution is capped to the default display size. Fullscreen windows on the external display will be saved but restored to the default maximum size of the first display and positioned in the top-left corner
* Linux Wayland: x/y positioning is not possible
* MS Windows: window composition works; fullscreen or maximized/minimized state is not saved or applied on next start, only the window bounds rectangle is saved
* macOS: same applies to the fullscreen desktop state/mode

### DIFF
--- a/src/common/constant.ts
+++ b/src/common/constant.ts
@@ -15,6 +15,9 @@ export const ABOUT_BOOK_TITLE_PREFIX = "EDRLAB.ThoriumReader-";
 export const WINDOW_MIN_WIDTH = 700;
 export const WINDOW_MIN_HEIGHT = 600;
 
+export const WINDOW_DEFAULT_WIDTH = 1024;
+export const WINDOW_DEFAULT_HEIGHT = 768;
+
 export const FORCE_PROD_DB_IN_DEV = false;
 
 export const USER_DATA_FOLDER = app.getPath("userData");
--- a/src/common/rectangle/window.ts
+++ b/src/common/rectangle/window.ts
@@ -5,8 +5,8 @@
 // that can be found in the LICENSE file exposed on Github (readium) in the project repository.
 // ==LICENSE-END==
 
-import { /*BrowserWindow,*/ Rectangle, screen } from "electron";
-import { WINDOW_DEFAULT_HEIGHT, WINDOW_DEFAULT_WIDTH } from "../constant";
+import { /*BrowserWindow,*/ Rectangle, screen, Display } from "electron";
+import { WINDOW_DEFAULT_HEIGHT, WINDOW_DEFAULT_WIDTH, WINDOW_MIN_HEIGHT, WINDOW_MIN_WIDTH } from "../constant";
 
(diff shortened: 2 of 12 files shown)


panaC:
https://github.com/user-attachments/assets/e7da915a-816c-44d6-aceb-6faeae762c71

PR #3478, opened 2026-04-02
fix the render of table titles in allPublications when locale change

arthur-lemeur:


danielweck:
Regression: https://github.com/edrlab/thorium-reader/issues/3486

danielweck:
Fixed regression: https://github.com/edrlab/thorium-reader/commit/59b0263644ebdc8f6ef3a0837abb267d70d37071

PR #3480, opened 2026-04-02
{DO NOT MERGE] main process didn't close properly on Windows, triggering breakpoint error

### LINKED ISSUE #3479: [Windows11 | linux] Shutdown/Reboot does not wait for main process to exit and save application state
During system shutdown/reboot, the application does not complete saving its final JSON state. As a result, on the next launch, a runtime + patch backup is restored instead of the latest state.

~~Additionally, performing a shutdown/reboot with version 3.4 and then restarting with version 3.3 potentially results a loss of the reader win registry, including locator, configs, and window bounds.~~ This is partially fixed with https://github.com/edrlab/thorium-reader/commit/4b53a27cf5857bb9f8f0f3ef8af070a5297c2c45

panaC:
Fixes #3479 

During shutdown on Windows, there is not enough time to persist the final state.
When the application waits for persistence to finish before closing the main process, a Windows breakpoint error is raised. This error blocks the shutdown entirely instead of allowing the process to terminate gracefully in the background.

This PR is currently broken. I was not able to resolve the issue in time for the 3.4 release without refactoring how final persistence is handled during the Windows session-end events.

This is not a major issue, as the runtime and patches are recovered on the next startup. Thanks to this commit https://github.com/edrlab/thorium-reader/commit/4b53a27cf5857bb9f8f0f3ef8af070a5297c2c45

The session-end event should immediately close or destroy all BrowserWindow instances, without attempting to perform additional work in the main process.
This issue may not be fully fixable on Windows, as the OS does not reliably allow enough time to persist the final state during shutdown or reboot.


src/main/redux/sagas/win/session/library.ts
```diff
@@ -27,14 +29,44 @@ function* libraryClosureManagement(action: winActions.session.registerLibrary.TA
...
+                    }).catch(() => {
+                        // nothing
+                    });
+
+            };
+            library.once("query-session-end", queryHandler);
+
+            const endHandler = (event: Electron.Event) => {
+                debug("WINDOWS11 session-end event", event);
+                if (!library.isDestroyed()) {
```
panaC, line 62: Without force-closing the library windows, a system breakpoint error dialog is raised, blocking the shutdown process until the user confirms it by "OK"

PR #3527, opened 2026-04-23
add aria-label to reader and library settings for better differentiation

### LINKED ISSUE #3145: Diferenciate better "Settings" and "Preferences"
There are two sets of parameters: 
* "Settings" is a menu for application-wide user-configurable settings exposed in the library window, 
* "Preferences" is a publication-specific options panel in the reader window(s).

This may not be very clear and requires context, which is not always evident, especially for non-visual readers. 

Proposal: 

Add context to those button titles, like "Application settings" and "Reading preferences"

arthur-lemeur:
Fixes #3145


danielweck:
Just a friendly reminder NOT to merge this Pull Request without first checking Weblate's "insights / history", merging the accumulated commits (Weblate PR), locking WebLate so that Thorium's upstream locales changes can be pushed without causing merge rebase conflicts.

Issue #3518, opened 2026-04-21
Accessibility information: links in comments


gautierchomel:
All comments at https://github.com/edrlab/thorium-reader/blob/198b8fb33fd606e11d741dd50ec3a6a386517df1/src/renderer/common/components/dialog/publicationInfos/PublicationInfoA11y2.tsx#L94
should be updated to https://www.w3.org/community/reports/publishingcg/CG-FINAL-epub-techniques-20251222/#visual-adjustments

panaC:
I do  not understand your issue, can you be more precise please

gautierchomel:
The link in comments is no longer valid, and we updated the code to stick to the last available version of the recommandation. Then the link in comments should be updated to [https://www.w3.org/community/reports/publishingcg/CG-FINAL-epub-techniques-20251222](https://www.w3.org/community/reports/publishingcg/CG-FINAL-epub-techniques-20251222/). This link should remain available even in the case of updated recommandation and allows us to remember to which version we complky. 

Sorry, the initial comment was too quick and I failed to reference the correct line :)

https://github.com/edrlab/thorium-reader/blob/198b8fb33fd606e11d741dd50ec3a6a386517df1/src/renderer/common/components/dialog/publicationInfos/PublicationInfoA11y2.tsx#L89

panaC:
So in that case this [commit](https://github.com/edrlab/thorium-reader/commit/b5b1d1dd8c6b59e3f3861857b25cf0c5328d981d) fixes the issue.

PR #3534, opened 2026-04-25
Localization of A11y issues, Storage/Recovery for new 3.5 Features, and label Fixes across the Project

panaC:
fixes #3533 

Now Fixes https://github.com/edrlab/thorium-reader/issues/3517 and Fixes #3518

PR #3538, opened 2026-04-26
A11y

### LINKED ISSUE #3517: Accessibility information: visual adjustments
visual adjustments: when `is_fixed_layout` is true, the technique says to display “Appearance cannot be modified”, but the component currently uses the same localized key as the modifiable case, so fixed-layout publications is labeled incorrectly.

https://github.com/edrlab/thorium-reader/blob/198b8fb33fd606e11d741dd50ec3a6a386517df1/src/renderer/common/components/dialog/publicationInfos/PublicationInfoA11y2.tsx#L94

https://www.w3.org/community/reports/publishingcg/CG-FINAL-epub-techniques-20251222/#visual-adjustments

panaC:
fixes #3517


panaC:
@gautierchomel please do a carefully review of the change according to the spec

gautierchomel:
Also closes #3518

gautierchomel (approved):
Works as expected on my side.

panaC:
I'm closing merged to #3534

Issue #3444, opened 2026-03-16
LCP publication download fails with ETIMEDOUT on high-latency connections (Happy Eyeballs 250ms timeout too short)


phardouin:
## Installation method: .deb package from official website
 
---
 
## Summary
 
When importing a `.lcpl` file, Thorium fails to download the LCP-protected publication from the content server. The download systematically fails with a `FetchError` / `ETIMEDOUT` error. This is caused by Node.js's Happy Eyeballs algorithm using a **250ms timeout** for the first IPv4 connection attempt before falling back to IPv6. For users with high-latency connections (e.g., overseas/Pacific region), the round-trip time exceeds 250ms, causing the IPv4 attempt to be aborted. The IPv6 fallback then also fails because the content server does not support IPv6.
 
This bug affects **all users whose latency to the content server exceeds 250ms**, which includes users in French overseas territories (Polynésie française, Nouvelle-Calédonie, Antilles, Réunion, etc.), and potentially users in Africa, Asia, or South America connecting to European LCP content servers.
 
---
 
## Error messages
 
**Error 1 (download attempt):**
```
Downloading [9782340113657] failed: [[AssertionError [ERR_ASSERTION]: http GET error: FetchError: request to https://content-lcp.editions-ellipses.fr/28fd820f-c7b8-4a08-bcb2-d1d5f31ff8bf failed, reason: (undefined) [https://content-lcp.editions-ellipses.fr/28fd820f-c7b8-4a08-bcb2-d1d5f31ff8bf]]]
```
 
**Error 2 (import attempt):**
```
Importing [/home/patrice/Téléchargements/9782340113657.lcpl] failed: [Error: download path undefined]
```
 
---
 
## Steps to reproduce
 
1. Install Thorium Reader 3.3.0 on a Linux machine located in a region with >250ms latency to the LCP content server (e.g., French Polynesia to France)
2. Download a `.lcpl` file from a publisher (e.g., Éditions Ellipses)
3. Import the `.lcpl` file into Thorium (drag-and-drop or "+" button)
4. Thorium attempts to download the publication and fails
 
---
 
### Network is functional
 
`curl` successfully downloads the file from the same machine:
 
```bash
$ curl -4 -I https://content-lcp.editions-ellipses.fr/28fd820f-c7b8-4a08-bcb2-d1d5f31ff8bf
HTTP/1.1 200 OK
Content-Length: 13877071
```
 
### IPv6 is NOT supported by the content server
 
```bash
[...]

PR #3540, opened 2026-04-26
IPV4 250ms timeout when IPV6 addresses provided from DNS but not supported from the server

panaC:
I asked ChatGPT about issue #3444 (and it seems the author of the issue does the same thing, lol).

It seems justified, and the fixes look valid.

Here’s ChatGPT’s explanation:

`autoSelectFamilyAttemptTimeout` is a Node TCP connection option used when Node’s address-family auto-selection is enabled.

In plain English: when a hostname resolves to both IPv4 and IPv6 addresses, Node tries them in sequence so it can quickly pick the one that works. That behavior is often called Happy Eyeballs. `autoSelectFamilyAttemptTimeout` is the per-address wait time, in milliseconds, before Node gives up on the current connection attempt and tries the next resolved address.

For this issue, the failure path was roughly:

```text
content-lcp.editions-ellipses.fr
  -> IPv4 address exists and works, but connection latency is about 270ms
  -> IPv6 address exists in DNS, but the server does not actually accept IPv6

Node tries IPv4
  -> default attempt timeout is 250ms in the affected runtime
  -> IPv4 is aborted just before it would connect

Node tries IPv6
  -> IPv6 fails

Result
  -> AggregateError / ETIMEDOUT
  -> Thorium reports the LCPL publication download failure
```

Node docs for the option: [socket.connect `autoSelectFamilyAttemptTimeout`](https://nodejs.org/api/net.html).

The question is why this is not a de facto default configuration for IPv4/IPv6, especially since IPv6 is not widely used. But I have the feeling the DNS may be misconfigured—why does it return an IPv6 address when the server does not support it?

### DIFF
--- a/src/main/network/http.ts
+++ b/src/main/network/http.ts
@@ -50,6 +50,7 @@ const filename_ = "readium-desktop:main/http";
 const debug = debug_(filename_);
 
 const DEFAULT_HTTP_TIMEOUT = 30000;
+const DEFAULT_AUTO_SELECT_FAMILY_ATTEMPT_TIMEOUT = 5000;
 
 let authenticationToken: Record<string, IOpdsAuthenticationToken> = {};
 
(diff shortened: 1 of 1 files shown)


panaC:
> The question is why this is not a de facto default configuration for IPv4/IPv6, especially since IPv6 is not widely used. But I have the feeling the DNS may be misconfigured—why does it return an IPv6 address when the server does not support it?

DNS can legitimately return an AAAA (IPv6) record if one exists, even if the actual server/network path is broken or IPv6 is misconfigured. That may be:

- bad DNS records
- broken dual-stack configuration
- server advertises IPv6 but doesn’t serve it properly

PR #3553, opened 2026-04-30
fix initial focus in library and reader modals

### LINKED ISSUE #3120: Place Focus on the menu list when opening Application Settings Panel
I am using the keyboard to navigate. 
When I reach the Application Settings Panel, the focus is placed on the close button. One tabulation and I go to "Language". 

Consequently, I have to tab through the entire General Panel to reach the `role="tablist"` element, which acts as a menu. I might never discover that there are more tabs.

The focus should be placed on the `role="tablist"` menu, with the first tab allowing access to the close button.

arthur-lemeur:
Fixes #3120


panaC:
It will change the accessibility tree, I don't know if it will have an impact.

arthur-lemeur:
@gautierchomel any insight about moving the modal control panel (buttons for dock position and closing modal) in last position of the modal tree ?

gautierchomel:
The current arrangement causes the `<h2>` title to be misplaced at the end and empty. A better approach would be to separate the title from the buttons, keeping the title at the top and the buttons at the bottom.
Furthermore, we've identified the following issues:

- The `<h2>` title is not nested under any `<h1>`, making it inaccessible to screen readers.
- When the panel is docked, this title disappears, and the navigation tree begins with an `<h3>`.

We need to determine how to maintain a consistent structure, whether the panel is modal or docked, and ensure this structure is semantically correct (e.g., always beginning with an `<h1>`).

We can quickly fix the issue here by removing the `<h2>` from the accessibility tree (`aria-hidden`) and opening an issue to address the global structure, maybe by using `aria-level` on tab and select elements.

gautierchomel:
I now see in both modes:
```
<h1 id="radix-:ra:">Reader settings</h1>
<h2>Display</h2>
<h4>Theme</h4>
```
`h4` should be `h3`. 

Screenshot 2026-05-13 at 16 56 02

gautierchomel (changes requested):
I now see in both modes:
```
<h1 id="radix-:ra:">Reader settings</h1>
<h2>Display</h2>
<h4>Theme</h4>
```
`h4` should be `h3`.

src/renderer/assets/styles/partials/variables.scss
```diff
@@ -60,15 +60,15 @@ $theme-buttons-border-dark: var(--theme-buttonsBorder_dark, $thorium-theme-butto
 
 $color-palette: (
   light: (
-    // ===== MAIN ===== //
+    /* ===== MAIN ===== //*/
```
panaC, line 63: remove all of this please

panaC (changes requested):

gautierchomel:
In the dialog view, the Tabs have `<h3>` and that should not be. 

IMO this PR can be merged and this fixed in a next one.

gautierchomel (approved):

panaC (approved):

panaC (commented):
Fixed gautier request

PR #3558, opened 2026-05-04
[PARKED FOR NOW] System tray

fschwiet:
Instead of having the library open as a second window:

image

It might be more convenient for the library to minimize to the system tray:

image

This let's the library window minimize by hiding itself, allowing it to be re-opened from the system tray.   I had an LLM suggest how to implement that in a minimal way for experimentation and discussion.

I also put in a change to hide the library window when the reader opens. This seems like nicer behavior, it might address https://github.com/edrlab/thorium-reader/issues/3505  and https://github.com/edrlab/thorium-reader/issues/2826 because I think the library window would be hidden in their cases.

I thought I remember seeing complaints about the app taking two windows from another user but couldn't find it. I agreed with that issue even if I also hallucinated it.

### DIFF
--- a/src/main/redux/sagas/reader.ts
+++ b/src/main/redux/sagas/reader.ts
@@ -7,7 +7,6 @@
 
 import debug_ from "debug";
 import { clipboard } from "electron";
-import { ReaderMode } from "readium-desktop/common/models/reader";
 import { Action } from "readium-desktop/common/models/redux";
 import { ActionWithDestination, ActionWithSender, SenderType } from "readium-desktop/common/models/sync";
 import { ToastType } from "readium-desktop/common/models/toast";
@@ -185,8 +184,8 @@ function* readerOpenRequest(action: readerActions.openRequest.TAction) {
         // );
         // const readersArray = ObjectValues(readers);
 
-        const mode = yield* selectTyped((state: RootState) => state.mode);
-        if (mode === ReaderMode.Attached) {
+        //const mode = yield* selectTyped((state: RootState) => state.mode);
+        //if (mode === ReaderMode.Attached) {
             try {
                 const libWin = getLibraryWindowFromDi();
                 if (libWin && !libWin.isDestroyed() && !libWin.webContents.isDestroyed()) {
@@ -195,7 +194,7 @@ function* readerOpenRequest(action: readerActions.openRequest.TAction) {
             } catch (_err) {
                 debug("library can't be loaded from di");
             }
-        }
+        //}
 
         const windowIdentifier = uuidv4();
         yield* callTyped(createReaderWindow,
--- a/src/main/redux/sagas/win/browserWindow/createLibraryWindow.ts
+++ b/src/main/redux/sagas/win/browserWindow/createLibraryWindow.ts
@@ -7,7 +7,7 @@
 
 import debug_ from "debug";
 import { encodeURIComponent_RFC3986 } from "@r2-utils-js/_utils/http/UrlUtils";
-import { BrowserWindow, Event as ElectronEvent, HandlerDetails, shell, WebContentsWillNavigateEventParams } from "electron";
+import { BrowserWindow, Event as ElectronEvent, HandlerDetails, Menu, shell, Tray, WebContentsWillNavigateEventParams } from "electron";
 import * as path from "path";
 import { normalizeWinBoundRectangle } from "readium-desktop/common/rectangle/window";
 import { diMainGet } from "readium-desktop/main/di";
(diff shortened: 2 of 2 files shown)


src/main/redux/sagas/win/browserWindow/createLibraryWindow.ts
```diff
@@ -59,10 +61,28 @@ export function* createLibraryWindow(_action: winActions.library.openRequest.TAc
...
-        icon: path.join(__dirname, "assets/icons/icon.png"),
+        icon: icon_path,
     });
     debug("LibraryWindow new BrowserWindow instancied");
 
+      // Initialize Tray with an icon
+    const tray = new Tray(icon_path);
+    const contextMenu = Menu.buildFromTemplate([
+        { label: 'Show Library', click: () => libWindow.show() },
+        { label: 'Quit', click: () => /*??*/null }
```
danielweck, line 72: TODO: translate labels

src/main/redux/sagas/win/browserWindow/createLibraryWindow.ts
```diff
@@ -59,10 +61,28 @@ export function* createLibraryWindow(_action: winActions.library.openRequest.TAc
             webSecurity: true,
             webviewTag: false,
         },
-        icon: path.join(__dirname, "assets/icons/icon.png"),
+        icon: icon_path,
     });
     debug("LibraryWindow new BrowserWindow instancied");
 
+      // Initialize Tray with an icon
+    const tray = new Tray(icon_path);
```
danielweck, line 69: is this specific to Microsoft Windows OS, or also applicable to Linux? (not MacOS I guess)

src/main/redux/sagas/reader.ts
```diff
@@ -185,8 +184,8 @@ function* readerOpenRequest(action: readerActions.openRequest.TAction) {
         // );
         // const readersArray = ObjectValues(readers);
 
-        const mode = yield* selectTyped((state: RootState) => state.mode);
-        if (mode === ReaderMode.Attached) {
+        //const mode = yield* selectTyped((state: RootState) => state.mode);
+        //if (mode === ReaderMode.Attached) {
```
danielweck, line 188: I think hiding the library window would remain necessary on MacOS (or even Linux) where a system tray is not available ? (assuming this is a Microsoft Windows -only feature)

panaC:
Very good idea, thanks...

src/main/redux/sagas/win/browserWindow/createLibraryWindow.ts
fschwiet: TODO: implement QUIT

The menu seems kind of bare, other options might make sense (perhaps open documents should be listed).

fschwiet:
Responding to @danielweck questions about this being windows only.. Unfortunately I don't know and do not have a MacOS machine to test this one. Let me try my Linux machine though and report how it behaves.

fschwiet:
Exploring the behavior on Ubuntu I don't see any difference running the changes in this branch. I don't see a system tray. The dock with active applications has a single Electron icon with two lights indicating its there for two windows. Alt-tabbing through windows I always see both.

image

According to the documentation (http://electronproject.org/tray-2.html) on MacOS and Linux support but I did not see it.

> On MacOS and Ubuntu, the Tray will be located on the top right corner of your screen, adjacent to your battery and wifi icons. On Windows, the Tray will usually be located in the bottom right corner.

It may be a mute point though for Linux at least, looking at the behavior of BrowserWindow.hide() it behaves the same as minimize (the window is no longer visible but is still active in the application’s process and in the desktop environment’s window manager). This means the user is still going to encounter the library when alt-tabbing through windows and no real advantage is gained over existing behavior.

It may be the same case on MacOS. I don't have any intuitions about what the MacOS behavior would or should be.

fschwiet:
My sense then is that this should be a Windows-only change. It'd be interesting to see how the current changes behave on MacOS to see if it is an improvement.

panaC:
I do not think this PR can be included in the upcoming 3.5 release, which is already in progress. However, I’m really interested in your work and your idea, so I’d like to assign it to the 3.6 project, with the agreement from the other maintainers.

fschwiet:
For sure. I had offered the PR more as an idea and didn't think it was fully baked.

Checking Electron documentation, it looks like the behavior could be made Windows only by checking process.platform === 'win32'.

Before I do that, it'd be interesting to see if the functionality would be helpful on MacOS. I don't know MacOS or have such a device. Does Thorium also end up with a second window on MacOS that we might want to suppress this way?

fschwiet:
updated branch to address concerns raised, new PR is #3601

PR #3561, opened 2026-05-05
design of the storage settings tab

### LINKED ISSUE #3555: Storage: UX and localization details
Related to #3529.

Publications are stored in a "[appData](https://www.electronjs.org/docs/latest/api/app#appgetpathname)" directory: 
on Windows: C:\Users\ user name \AppData\Roaming\EDRLab.ThoriumReader\publications\
on MacOS: /Users/ user name /Library/Application Support/EDRLab.ThoriumReader/publications/
on Debian: /home/ user name /.config/EDRLab.ThoriumReader/publications/ 

We had complaints about system disks being full, and users were willing to move the storage to a different location. Thorium 3.5 offers this new feature **not** by automatically moving all publications to a new directory (which could fail if the target directory is too small), but by adding a storage directory to the configuration. Every publication imported into the application after such "additional" storage has been configured will be stored there. Advanced users will also be able to manually move publications from system storage to additional storage without disrupting the application.   

Note: I prefer the term "additional" storage because "external" storage seems to mean only an external disk. Most users - especially on Windows - will use an internal disk in their PC. If an external disk is used, unplugging it triggers user warnings in the application. 

A set of modifications to the labels associated with the feature is listed below, reflecting the change from "internal"/"external" to "system"/"additional". 

Let's first look at UX refinements:

The Storage config panel: 
1. I see a large white area and "Storage" subtitle: please remove this subtitle and align the blurb text with the content of the other panels. 
2. First block: Replace the subtitle "Locations" with "System storage"; replace "Default internal storage" with the label "Publications managed by the application are by default stored in"; make the path string smaller, and not active; add a button labelled "Open". 
3. Second block: The process is simplified: there are too many different states now. Replace the subtitle "Configuration" with "Additional storage"; The label displayed when there is no additional storage yet is updated in the next comment; in this case, the button reads "Select an additional storage". There is no "Cancel" button. If an additional storage has been selected, the label is updated (see below), the path to the additional storage appears in this block (not in the first block), and there are two buttons: "Update additional storage" and "Remove additional storage". 
4. Third block: labels are updated  (see below). The "Check again" button has the same style as the other buttons. After recovery, the button "

arthur-lemeur:
Fixes #3555


gautierchomel (commented):
I suggest the following text lines 760 to 769: 

> This feature is currently in beta. Learn more. 
> Thorium does not automatically transfer your books. If you change the storage location, you must manually move your publications.
> You must keep this folder safe and accessible. Avoid deleting it, restricting access, or using slow or unstable storage, like network drives.
> Do not manually alter the files in this folder, as this may cause issues with your books or the reader. Think of this folder as a secure library managed by Thorium.
> This feature works only with recent versions of Thorium; external folder books will not appear in Thorium 3.4 or earlier.
>

panaC:
@gautierchomel @arthur-lemeur You forget existing of this issue https://github.com/edrlab/thorium-reader/issues/3555#issuecomment-4359251961 from Laurent and this PR from me https://github.com/edrlab/thorium-reader/pull/3534.

We must now reconcile them both.

panaC:
@gautierchomel We need to know the exact label/translation to use. What is your final choice?

panaC (dismissed):
Internationalization conflicts must be resolved before the merge

danielweck:
also, please do not merge before checking the WebLate state (insights >> history), lock the environment, merge the WebLate PR in upstream before merging this PR :)

panaC:
>also, please do not merge before checking the WebLate state (insights >> history), lock the environment, merge the WebLate PR in upstream before merging this PR :)

Please see my comment in the EDRLab Slack.

panaC:
I revert https://github.com/edrlab/thorium-reader/pull/3561/commits/52a8961d7ff11182e08fd1a3126eb259512db9d8. Now localization will only come from this. [PR](https://github.com/edrlab/thorium-reader/pull/3534)

llemeurfr:
Re. labels, the only difference is Gautier's proposal for an explanation of the feature (l758-762 is my proposal). 
Gautier's proposal is good BUT its focus on "change the storage location" is ambiguous. The feature is not about "changing" the storage location, it is about "adding a storage location". 
Notes : What is the "learn more" text linking to? "folder" is not used by Windows users; "directory" is more generic.

Final wording: see https://github.com/edrlab/thorium-reader/pull/3534#issuecomment-4420963329

gautierchomel:
"learn more" is because the title is a button, so the action needs to be informed.

Issue #3563, opened 2026-05-06
Omit display of "No information about prerecorded audio is available"


gautierchomel:
https://w3c.github.io/cg-reports/publishingcg/CG-FINAL-epub-techniques-20250422/#instructions-1 let implementer choose:

> ELSE either omit this display field if metadata is missing or display "No information about prerecorded audio is available"[ID: ways-of-reading-prerecorded-audio-no-metadata].

We should omit this information as it is strange to have it display on so many ebooks that are just textuals.

PR #3570, opened 2026-05-11
Omit display of "No information about prerecorded audio is available"

gautierchomel:
Replace #3564 to add fix for #3563. Authored with help of GitHub Copilot and tested with several publications.


danielweck:
Just a quick FYI (no action needed): the `ways-of-reading-prerecorded-audio-no-metadata` JSON locale key/value will be removed as it is now unused. This will be done automatically via `npm run i18n-scan` to update the English master locale, and `npm run i18n-check` to verify-in/apply-to all other locales)

danielweck:
Follow-up commit: https://github.com/edrlab/thorium-reader/commit/5a71756cbdc912b388fecec8285ece36da0ab041
(I locked Weblate in order to reset their fork and avoid merge/rebase conflicts)

PR #3576, opened 2026-05-12
fix styles in storage tab settings

### LINKED ISSUE #3571: [CSS] button text overflow
Image

### LINKED ISSUE #3572: storage: make the path string smaller
from https://github.com/edrlab/thorium-reader/issues/3555#issue-4364103314 :  

This is partially fixed

First block: Replace the subtitle "Locations" with "System storage"; replace "Default internal storage" with the label "Publications managed by the application are by default stored in"; make the path string smaller, and not active; add a button labelled "Open".

Image

### LINKED ISSUE #3573: storage: no "Cancel" button
from https://github.com/edrlab/thorium-reader/issues/3555#issue-4364103314 :

This is partially fixed

Second block: The process is simplified: there are too many different states now. Replace the subtitle "Configuration" with "Additional storage"; The label displayed when there is no additional storage yet is updated in the next comment; in this case, the button reads "Select an additional storage". **There is no "Cancel" button**. If an additional storage has been selected, the label is updated (see below), the path to the additional storage appears in this block (not in the first block), and there are two buttons: "Update additional storage" and "Remove additional storage".

Image

arthur-lemeur:
Fixes #3571 
Fixes #3572 
Fixes #3573

### DIFF
--- a/src/renderer/assets/styles/components/buttons.scss
+++ b/src/renderer/assets/styles/components/buttons.scss
@@ -56,11 +56,6 @@ button {
   transition: 200ms linear;
   align-items: center;
   justify-content: center;
-  overflow: hidden;
-  display: -webkit-box;
-  -webkit-line-clamp: 1;
-  -webkit-box-orient: vertical;
-  line-height: 1.4;
 
   &:focus-within, &:focus {
     outline: var(--color-brand-primary) solid 1px !important;
@@ -170,11 +165,6 @@ button {
   color: var(--color-brand-primary);
   align-items: center;
   justify-content: center;
-  overflow: hidden;
-  display: -webkit-box;
-  -webkit-line-clamp: 1;
-  -webkit-box-orient: vertical;
-  line-height: 1.4;
 
   &:hover {
     color: var(--color-brand-primary);
--- a/src/renderer/assets/styles/components/buttons.scss
+++ b/src/renderer/assets/styles/components/buttons.scss
@@ -54,9 +54,13 @@ button {
   background-color: var(--color-brand-primary);
   color: var(--color-neutral-base);
   transition: 200ms linear;
-  display: inline-flex;
   align-items: center;
   justify-content: center;
+  overflow: hidden;
+  display: -webkit-box;
+  -webkit-line-clamp: 1;
+  -webkit-box-orient: vertical;
+  line-height: 1.4;
 
   &:focus-within, &:focus {
     outline: var(--color-brand-primary) solid 1px !important;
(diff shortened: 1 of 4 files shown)


panaC (commented):
I need more precision because this does not seem to address all the listed issues.

Could you please explain how you are fixing each issue?

arthur-lemeur:
This PR add some contraints to the buttons in order to prevent overflow in case of too long labels (Issue #3571). It also execute the openFolderPicker function directly after confirming the alert dialog when adding an external storage directory which removes the step where there were a "cancel" button ( Issue #3573). Regarding the issue #3572 we agreed with Laurent to leave the path strings as the buttons itself and not add an "open" button next to it yet. We can discuss further if necessary, for now juste moved the styles to the css files.

panaC:
@arthur-lemeur In response to your last message: according to https://github.com/edrlab/thorium-reader/issues/3572, the issue should be closed in that case, with a comment explaining why the modification is not requested for now.

For https://github.com/edrlab/thorium-reader/issues/3573, the “Cancel” button step still exists as far as I know. 
Need to remove it instead of comment it

src/renderer/assets/styles/components/buttons.scss
```diff
@@ -54,9 +54,13 @@ button {
   background-color: var(--color-brand-primary);
   color: var(--color-neutral-base);
   transition: 200ms linear;
-  display: inline-flex;
   align-items: center;
   justify-content: center;
+  overflow: hidden;
```
panaC: With overflow hidden the directory path will be truncated.

The change seems to impact every button across the codebase. Is it tested ?

src/renderer/library/components/settings/Settings.tsx
```diff
@@ -771,58 +772,33 @@ const StorageSettings: React.FC<{}> = () => {
...
-                                            height: "auto",
-                                            minHeight: "unset",
-                                            textAlign: "left",
-                                            whiteSpace: "normal",
-                                            overflowWrap: "anywhere",
-                                            wordBreak: "break-word",
-                                            lineHeight: 1.4,
-                                            fontSize: "14px",
-                                        }}
+                                    <button className={stylesButtons.button_nav_tertiary}
```
panaC: userDirectory will be truncated at the end, but I don’t think there’s a tooltip to display the full filePath.

Am I wrong?

panaC (changes requested):

src/renderer/library/components/settings/Settings.tsx
panaC: At least add a title attribute on the button ?

panaC (changes requested):
https://github.com/user-attachments/assets/67166a9c-c24d-46ab-b6fe-7c8c3d636344


I cannot change the directory

PR #3583, opened 2026-05-18
fix tts playback rate combobox and container width in highlight preview

### LINKED ISSUE #3579: Clearing the TTS Playback Rate box results in blank window
When the setting is set to an empty string (Fig.1)(e.g. clicking backspace multiple times until field is cleared), the whole window becomes blank (Fig. 2).

# Fig. 1
Image

# Fig. 2
Image

arthur-lemeur:
Fixes #3579

### DIFF
--- a/src/renderer/assets/styles/components/readerHeader.scss
+++ b/src/renderer/assets/styles/components/readerHeader.scss
@@ -239,7 +239,7 @@
     color: var(--color-text-primary);
 
     details {
-        width: 0px !important;
+        width: 0 !important;
         flex-basis: 100%;
         margin-top: 10px;
         border: 1px solid var(--color-gray-250);
--- a/src/renderer/assets/styles/components/readerHeader.scss
+++ b/src/renderer/assets/styles/components/readerHeader.scss
@@ -238,6 +238,14 @@
     /* gap: 20px; */
     color: var(--color-text-primary);
 
+    details {
+        width: 0px !important;
+        flex-basis: 100%;
+        margin-top: 10px;
+        border: 1px solid var(--color-gray-250);
+        padding: 6px;
+    }
+
     @media only screen and (width <= 980px) {
         max-height: 520px;
         overflow-y: auto;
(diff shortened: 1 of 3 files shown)


src/renderer/assets/styles/components/readerHeader.scss
```diff
@@ -238,6 +238,14 @@
     /* gap: 20px; */
     color: var(--color-text-primary);
 
+    details {
```
panaC, line 241: Are you 100% sure that nothing has been changed with this new selector?

src/renderer/reader/components/ReaderHeader.tsx
```diff
@@ -792,15 +792,17 @@ export class ReaderHeader extends React.Component<IProps, IState> {
                                                                                     2
                                                                             }
                                                                             onSelectionChange={(ev) => {
-                                                                                const v = playbackRate.find((option) => option.id === ev).value;
```
panaC, line 795: This is more a limitation of the combobox element than a validation of whether the value belongs to the playback rate value set. The combobox element should be more restrictive and should not allow entering values that are not part of the allowed value set. What do you think  ?

panaC (changes requested):

src/renderer/reader/components/ReaderHeader.tsx
panaC: Please check the fully controlled ComboBox example: https://react-spectrum.adobe.com/ComboBox#fully-controlled

panaC: But it might be too complex to implement within the scope of this PR.

panaC: Sorry you're right, the time to reload the context in my mind. I'm approve the change.

src/renderer/assets/styles/components/readerHeader.scss
arthur-lemeur: There was a bug with the width of the details element in the tts popover. As far as I know, there is only one details element as child of the container with the classname : "Tts_popover_container" which this style is a child of in the stylesheet so no unexpected cascade should be occuring here.

panaC:
@arthur-lemeur The PR does not pass the CI.

danielweck:
src/renderer/assets/styles/components/readerHeader.scss
  242:17  ✖  Disallowed unit  length-zero-no-unit

PR #3601, opened 2026-05-23
app minimizes to the system tray on Windows

fschwiet:
This PR is an update from #3558, handling the concerns discussed in that PR. The code is now internationalized (with no new strings needed) and only affect Windows.

Something I'm less sure about is adding a new file to export enableSystemTrayIntegration. I like checking enableSystemTrayIntegration  instead of checking if its Windows directly as that communicates better what is going on, particularly in reader.ts. But I wasn't quite sure where to put the export, there didn't seem to be a place to put things like that. There are checks on process.platform in other parts of the code, one might consider replacing those with new exports that communicate intent.

## Overview

Instead of having the library open as a second window:

image

It is more convenient for the library to minimize to the system tray:

image

This let's the library window minimize by hiding itself, allowing it to be re-opened from the system tray.   I had an LLM suggest how to implement that in a minimal way for experimentation and discussion.

I also put in a change to hide the library window when the reader opens. This seems like nicer behavior, it might address https://github.com/edrlab/thorium-reader/issues/3505  and https://github.com/edrlab/thorium-reader/issues/2826 because I think the library window would be hidden in their cases.

I thought I remember seeing complaints about the app taking two windows from another user but couldn't find it. I agreed with that issue even if I also hallucinated it.

### DIFF
--- a/src/main/redux/sagas/reader.ts
+++ b/src/main/redux/sagas/reader.ts
@@ -17,6 +17,7 @@ import { takeSpawnEvery } from "readium-desktop/common/redux/sagas/takeSpawnEver
 import { takeSpawnLeading } from "readium-desktop/common/redux/sagas/takeSpawnLeading";
 import { diMainGet, getAllReaderWindowFromDi, getLibraryWindowFromDi, getReaderWindowFromDi } from "readium-desktop/main/di";
 import { error } from "readium-desktop/main/tools/error";
+import { enableSystemTrayIntegration } from "readium-desktop/main/tools/platform";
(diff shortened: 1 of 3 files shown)


fschwiet:
Are we parked?

danielweck:
Hello, thank you for the updates :)
So, what happens when only the library window is displayed, no reader windows are opened? As a user, if I minimise the library window it goes in the deeply-hidden system tray, it does not show in the "task bar" (dock?) anymore? Sorry I haven't tested this change in my Windows VM yet (lack of time at the moment)

PR #3604, opened 2026-05-26
[PARKED] create generic primary and secondary button and refactor buttons stylesheet

arthur-lemeur:
**Structure**

Wrapping the button text in a `span` tag allows us to handle text overflow behavior (text exceeding the button's bounds) by forcing it to stay on a single line and adding ellipses (…). To achieve this behavior combined with alignment alongside a preceding SVG icon, this method is **mandatory**.

**Variants**

The button currently has two variants:

- primary
- secondary

This affects the class assigned to the button (`button_primary` or `button_secondary`), which avoids code duplication and makes this button easily scalable.

**Props**

```typescript
type IButtonProps = Omit<React.ComponentPropsWithoutRef<"button">, "className" | "title" | "aria-label"> & {
    label: string;
    variant?: ButtonVariant;
    svg?: ISVGProps;
    svgClassname?: string;
    extendedLabel?: string;
}
```

- `label` is the only required prop, all others are optional.
- `label` automatically populates the button's text content as well as the `aria-label` and `title` attributes (if no `extendedLabel` is provided), avoiding code duplication.
- The `variant` prop (`"primary" | "secondary"` enum) defaults to `"primary"`.

**Component usage**

The button can be called as either:

- `ThButton`
- `ThButtonPrimary`
- `ThButtonSecondary`

The plain `ThButton` call allows conditionally applying a `"primary"` or `"secondary"` style. Example:

```tsx
<ThButton
    variant={isPrimary ? "primary" : "secondary"}
    ...
/>
```

This avoids having to write:

```tsx
{isPrimary ?
    <ThButtonPrimary
        ...
    />
:
    <ThButtonSecondary
        ...
    />
}
```

If its style is immutable, using the variant-specific component makes the code easier to read and saves time.

**Component**

```
export const ThButton = React.forwardRef<HTMLButtonElement, IButtonProps>(
    ({ variant = "primary", extendedLabel, label, svg, svgClassname, ...props }, ref) => (
        <button
            {...props}
            ref={ref}
            title={extendedLabel || label}
            aria-label={extendedLabel || label}
            className={variantStyles[variant]}
        >
            {svg && <SVG ariaHidden svg={svg} className={svgClassname} />}
            <span>{label}</span>
        </button>
    ),
);
```


panaC (changes requested):
Please fix #3626 independently of this PR.

PR #3617, opened 2026-05-27
fix lost of focus in settings recovery button

### LINKED ISSUE #3591: [UI] space or enter on settings/storage/Recovery check button lost focus on it
Image

Image

keyboard interaction with the button move the focus on the entire modal and do not stay on the button.

arthur-lemeur:
Fixes #3591


panaC (changes requested):
The button onclick is still active when disabled

PR #3620, opened 2026-05-27
fix: webpack config jsdom

### LINKED ISSUE #3619: develop production build package:win crash on runtime
Image

panaC:
Fixes #3619 

No idea why !

### DIFF
--- a/webpack.config.main.js
+++ b/webpack.config.main.js
@@ -331,6 +331,10 @@ if (nodeEnv !== "production") {
     config.optimization = {
         ...(config.optimization || {}),
         nodeEnv: false,
+        // jsdom generated WebIDL wrappers access CommonJS exports dynamically
+        // (Impl.implementation). Production usedExports can incorrectly remove
+        // those implementation exports from the bundled main process.
+        usedExports: false,
         minimize: true,
         minimizer: [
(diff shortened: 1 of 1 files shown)


panaC:
https://github.com/jsdom/jsdom/blob/d312832b344c8055c0d61e55bacd1a675ad5826e/lib/jsdom/living/nodes/XMLDocument-impl.js#L2

EDIT: more specifically for version `26.1.0`:
https://github.com/jsdom/jsdom/blob/v26.1.0/lib/jsdom/living/nodes/XMLDocument-impl.js#L4

danielweck:
thank you Pierre
https://webpack.js.org/configuration/optimization/#optimizationusedexports

why is this breaking now??!
I'd like to regression-test this before we merge this solution/workaround.

danielweck:
I suspect that the transition from terser-webpack-plugin to minimizer-webpack-plugin has something to do with this. I will check.

danielweck:
I am taking over this fix, thank you very much for figuring this out Pierre!

danielweck:
Fixed via https://github.com/edrlab/thorium-reader/commit/5ef1172f3eb6fcf9fc813cafb0d00e5b977f4895

PR #3633, opened 2026-06-03
add a setting checkbox for shared computer to remove/replace expired LCP publication on interval and import/open.

### LINKED ISSUE #3491: Feature: Add a setting to automatically remove/delete expired LCP publication licenses on startup or scheduled Intervals
Use case: Thorium Reader is used on a shared computer. Students loan a book for some hours, and leave the computer. The book can stay accessible until its license expires. Once the license has expired, the book is automatically deleted from the storage. Another student can then import the same book, there will be no message like "the title already exists".

panaC:
Fixes #3491 

This PR adds explicit handling for expired or no-longer-usable LCP publications. In personal/default mode, Thorium keeps expired LCP publications and preserves user data during replacement imports. In shared-computer mode, Thorium removes LCP publications that are confirmed unusable after LSD synchronization. A new command-line flag, `--shared-computer` forces shared-computer mode for the current process and disables the settings checkbox.

Features: 

- Runs cleanup when the setting is enabled, after startup, and periodically every 15 minutes.
- Shared-computer setting is disabled by default and can be enabled from Settings.
- Adds `--shared-computer` as a runtime CLI override.
- `--shared-computer` forces shared-computer mode at startup. The checkbox is shown as checked and disabled, so the user cannot turn it off while the app was launched with this argument.
- In shared-computer mode, Thorium deletes LCP publications only when LSD confirms a terminal status: expired, revoked, returned, or cancelled.
- In personal mode, replacement import preserves user data from the old LCP publication and restores it into the newly imported publication.
- Preserved personal-mode data includes notes, bookmarks/annotations counters, reading locator, reader config, PDF config, divina config, RTL flip setting, custom config allowance, and window bounds.
- Updated the settings UI to disable the checkbox when shared-computer mode is forced by CLI.
- Keeps CLI-forced shared-computer mode process-local and non-persisted.
- Updated personal-mode replacement logic so user data is not captured/restored when shared-computer mode is active.

Code changes:

- Added LCP usability helpers in `src/common/lcp.ts`.
- Added new settings state, actions, reducer handling, and sync support for:
  - `lcpAutoDeleteExpiredPublications`
  - `lcpAutoDeleteExpiredPublicationsForced`
- Updated publication open flow to invoke cleanup after LCP/LSD synchronization.
- Hardened publication deletion with LCP-aware file locking.
- Updated the publication API delete signature to support lock-owned cleanup paths.
- Updated import flows to delete unusable previous LCP copies and retry import once.
- Added preservation/restoration of notes, note counts, locator/config, PDF config, bounds, and reader state during personal-mode replacement imports.
- Added `sqliteTableNoteInsertOrReplace()` for restoring notes onto replacement publications.
- Ensured CLI-forced shared-computer mode is cleared before persistence and ignored during store hydration.
- Guarded deferred post-import LCP license updates against publications deleted meanwhile.

Impact: 

- Personal users keep their notes and reading state when replacing an expired LCP publication with a new license/publication.
- Shared workstation deployments can enforce privacy-oriented cleanup without relying on each user to enable the setting manually.
- The forced CLI mode is scoped to the current application process. Relaunching Thorium without `--shared-computer` does not permanently force the setting.

LCP flow: 

- On startup, Thorium loads persisted settings, resets any persisted forced override, then applies --shared-computer if present.
- When shared-computer mode is active, the periodic LCP cleanup scans publications and runs LSD synchronization before deletion.
- Deletion happens only after the latest publication state confirms the LCP license is no longer usable.
- When importing a new instance of an already-imported LCP publication, Thorium can remove the old unusable LCP publication first, then import the new one.
- In personal mode, user data is captured before deletion and restored after the replacement import.
- Automatic cleanup is conservative and waits for confirmed LSD status. Manual replacement import can use the local expiry date as a fallback, because the user is actively replacing the publication, and Thorium tries to preserve their personal data.

There are two ways Thorium can decide an LCP publication is no longer usable:
- Confirmed server/LSD status:
   The license status says: `Expired`, `Revoked`, `Returned`, or `Cancelled`.
- Local `rights.end` date:
   The license file contains an end date, and that date is now in the past.

For background shared-computer cleanup, Thorium only deletes a publication when the server/LSD status confirms it is unusable. It does not delete just because the local `rights.end` date is expired, because that could be stale, offline, affected by clock issues, or not fully synchronized.

For user-initiated replacement imports, the app is more permissive. If the user imports a new copy and Thorium detects an already-imported old LCP copy whose local `rights.end` is expired, it can delete the old copy and retry the import. In personal mode, it first captures local notes and reading state, then restores them onto the replacement publication when possible.

Risk:

- Main risk is around LCP state classification and deletion timing. This is mitigated by requiring confirmed LSD terminal status for background/shared-computer cleanup.
- Replacement import allows local rights.end fallback because it is user-initiated and needed to replace an expired local license.
- Concurrency risk is mitigated with publication file locks, re-fetching publication state before deletion, and final checks immediately before calling the delete flow.
- Persistence risk is mitigated by explicitly excluding the CLI forced flag from persisted state and resetting it during store initialization.

Test:

> Personal mode keeps expired LCP data

Steps:
1. Start Thorium normally, without `--shared-computer`.
2. Verify the shared-computer checkbox is unchecked and editable.
3. Import an LCP publication.
4. Add notes, bookmarks, and change reading settings.
5. Make the LCP license expire or use an already expired test license.
6. Trigger LSD sync by opening the publication or refreshing the license state.
7. Confirm the publication is not deleted from the library.

Expected result:
The expired LCP publication remains in the library, and notes/user data are preserved.

> Personal mode replaces expired LCP and restores user data

Steps:
1. Start Thorium normally.
2. Import an LCP publication.
3. Add notes, bookmarks, reading position, PDF/reader settings if applicable.
4. Import a new valid license/publication for the same book.
5. Do not manually delete the old publication.
6. Open the newly imported publication.

Expected result:
The old unusable LCP publication is replaced, and notes plus reading settings are restored in the new publication.

> Shared-computer setting deletes unusable LCP publications

Steps:
1. Start Thorium normally.
2. Enable the shared-computer checkbox in Settings.
3. Import an LCP publication with a terminal LSD status, such as expired, revoked, returned, or cancelled.
4. Trigger LSD sync.
5. Wait for cleanup or reopen the app to allow the cleanup scan.

Expected result:
The unusable LCP publication is deleted from the library, including notes and reading state.

> CLI forced shared-computer mode

Steps:
1. Launch Thorium with `--shared-computer`.
2. Open Settings.
3. Check the shared-computer checkbox state.
4. Try to toggle it with mouse and keyboard.

Expected result:
The checkbox is checked, disabled, non-editable, and shared-computer cleanup is active.

> CLI forced mode is not persisted

Steps:
1. Launch Thorium with `--shared-computer`.
2. Confirm the checkbox is checked and disabled.
3. Close Thorium.
4. Relaunch Thorium without `--shared-computer`.
5. Open Settings.

Expected result:
The checkbox is no longer forced. It returns to the persisted user setting and is editable again.

> Shared-computer mode does not preserve personal user data on replacement

Steps:
1. Enable shared-computer mode or launch with `--shared-computer`.
2. Import an LCP publication.
3. Add notes and reading settings.
4. Replace it with a new license/publication for the same book.
5. Open the new publication.

Expected result:
The old publication and its user data are removed through the deletion flow. Notes and reading state are not restored.

> Non-LCP publications are not affected

Steps:
1. Import a normal EPUB or PDF without LCP.
2. Enable shared-computer mode.
3. Restart Thorium and trigger normal library refresh.
4. Open and close the non-LCP publication.

Expected result:
The non-LCP publication is not deleted and behaves normally.

> Concurrent open/import safety

Steps:
1. Import an expired or terminal-status LCP publication.
2. Start importing a replacement license/publication.
3. While import is running, try opening the old publication.
4. Repeat once with shared-computer mode enabled.

Expected result:
Thorium does not crash, does not create duplicate corrupted entries, and ends with either the valid replacement publication or the expected deletion state.

> Concurrent open and 400ms then delete the publication

Steps:
1. Import an LCP publication.
2. remove them programmatically 400ms after.
3. waiting during delete because LCP lock is active
4. Repeat once with shared-computer mode enabled.


panaC:
Mode `--shared-computer` forced: 

image

llemeurfr:
A comment: the impact of requesting the status document for every publication in Thorium every 15 minutes is limited by the fact that it is only active in shared computer mode. Shared computers are used in libraries. Publications are (mostly?) loans in such cases, which expire after a few hours. Thorium's local catalog will be empty most of the time.

llemeurfr:
A question: did you test the behavior of the shared-computer mode if Thorium is offline (-> the status document request fails)? 
Oh I see: "Automatic cleanup is conservative and waits for confirmed LSD status. Manual replacement import can use the local expiry date as a fallback, because the user is actively replacing the publication, and Thorium tries to preserve their personal data." This is good.

llemeurfr:
Please replace the text in the setting with the following (we must not display technical jargon to end users):

**Shared computer**

(!) Adapted to library and university environments; makes sure that expired loans don't remain visible, with personal reading location and annotations attached. 

**[ ] Automatically suppress expired publications and associated user data**

PR #3642, opened 2026-06-12
Fix LCP license refresh when the publication is updated

### LINKED ISSUE #3638: When a License Status Document advertises an updated LCP License (usually for loan renewal, or revokation), there is a possibility that the publication itself is updated
I experienced this with a LCP-protected PDF from a well-known vendor: the LCP license profile was updated from 1.0 to one of the newer 2.x profiles (that's fine), the passphrase changed (this causes user friction if the passphrase needs to be entered manually, but this is otherwise a transparent change if the OPDS publication feed transports the hashed passphrase) ... but the (unchanged) `rel=publication` link URL in the JSON license file resolved to a different encrypted PDF (with a drastically different byte length actually, but that's a separate issue altogether ... looks like a LCP deployment bug or bad practice, I am investigating this separately).

So, the problem here is that Thorium Desktop reader does not check that a publication has changed during a license update (we could compare the length and integrity hash `link` object fields) ... is that even a "normal" real-world scenario or is this an anomaly?

ping @llemeurfr

### LINKED ISSUE #3646: LCP hash integrity broken when encoded as base64 instead of base16
Image

panaC:
Fixes #3638
Fixes #3646 

LCP/LSD license refresh handling when the updated LCPL points to a different protected publication archive through its `rel="publication"` link.

Now detects when an LCP license publication archive is updated from the publication link `hash` or `length`. When that happens, Thorium downloads the updated archive, validates its integrity, injects the refreshed LCPL, replaces the stored publication files transactionally, and updates the publication document metadata/cache.

- Added SHA-256 helper support for `hex`, `base64`, `base64url`, and padded `base64url`.
- Fixed hash comparison so one valid encoding match is enough.
- Added LCP publication-link change detection in `src/main/services/lcp.ts`.
- Added transactional publication file replacement in `PublicationStorage`.
- Shared cover/custom-cover document patch logic between normal import and LCP archive replacement.
- Replaced `Object.assign` usage with spread syntax in touched code.
- Preferred `undefined` over `null` in new/updated optional-value paths.
- Added the `undefined` over `null` preference to `AGENTS.md`.
- Added unit tests for hash handling, document patching, and storage rollback/finalize behavior.

`src/main/services/lcp.ts` now compares the previous and refreshed LCPL `rel="publication"` link. If the normalized hash or finite length changed, it downloads the new archive, checks length/hash integrity, injects the new LCPL, parses the publication metadata, and stages the replacement.

`src/main/storage/publication-storage.ts` now returns an `IPublicationFilesReplacement` handle from `replacePublicationFiles()`. The replacement can be finalized after the database save succeeds, or rolled back if saving/cache update fails.

`src/main/tools/fileIntegrity.ts` centralizes SHA-256 computation and comparison. This avoids duplicating stream hashing and handles the different digest encodings LCP servers may emit.

`src/main/tools/publicationDocument.ts` centralizes the logic that separates archive files from cover files and decides whether to keep or generate a custom cover.

## Tests
1. Start Thorium from this branch.

2. Import an LCP-protected publication whose LCPL contains an LSD status link.

3. Unlock the publication with the valid passphrase and confirm it opens normally.

4. On the LSD/LCP test server, update the license response so the refreshed LCPL contains a `links` item with `rel="publication"` pointing to a different protected archive.

5. Make sure the new `publication` link has either:
   - a different SHA-256 `hash`, or
   - a different `length`.

6. Trigger license refresh in Thorium:
   - open the publication again, or
   - use renew/return/status refresh flow depending on the test fixture.

7. Expected result:
   - Thorium detects the archive changed.
   - It downloads the new archive.
   - It validates length/hash.
   - It injects the refreshed LCPL.
   - It replaces the stored publication files.
   - The publication opens with the updated content.

8. Verify metadata:
   - title updates if the replacement archive has a different title,
   - cover updates if the replacement archive has a real cover,
   - custom cover remains/generated only when no real cover exists.

9. Negative test: change the server-provided `hash` to an invalid value and refresh again.

10. Expected negative result:
   - replacement is rejected,
   - old publication remains readable,
   - no partial `book.*`, `cover.*`, `manifest.json`, or `license.lcpl` replacement remains in storage.

----

Add tests for LCP archive replacement failure handling

- Verify archive download failure does not stage publication file replacement
- Verify rollback guard still runs when no replacement was staged
- Update LCP publication-link test to match current href-ignore behavior


panaC:
Branch created on `edrlab/lcp-server` to test publication and license update: https://github.com/panaC/lcp-server/blob/5b560399506d513c784f465c3be4a0ca9ff7a076/lcp-update.md


see https://github.com/edrlab/lcp-server/pull/9


https://github.com/user-attachments/assets/f0fcce7e-6f2c-4958-82e7-df8e2d78d14d


To run Thorium under windows command console: 
`set THORIUM_ISURL_REQUIRE_TLD_FALSE=true && npm run start:dev:quick`

panaC:
After the publication is replaced via LCP, the personal settings and annotations remain present. However, the selectors do not surround exactly the same note anymore, as shown in the example below:

image

danielweck:
> After the publication is replaced via LCP, the personal settings and annotations remain present. However, the selectors do not surround exactly the same note anymore

is that the expected outcome based on your test fixtures (EPUB replacement), or is that unexpected (is this a bug?)

panaC:
I created an issue on the `readium/lcp-specs` repo : https://github.com/readium/lcp-specs/issues/59

panaC:
> is that the expected outcome based on your test fixtures (EPUB replacement), or is that unexpected (is this a bug?)

Thorium's internal locator allows this kind of selection, if I'm not mistaken, so this is a feature rather than a bug. When a newer publication changes slightly, the annotation note still covers some of the previously annotated content.
We need to work on this and clearly define the expected behavior and specifications.

danielweck:
> We need to work on this and clearly define the expected behavior and specifications.

well, CFI includes a text assertion feature that supposedly helps recover character-range text sélections after minor editorial changes, and the Colibrio implementation is being upgraded for this use case (we currently vendor a specific revision of their lib).
Also, the W3C URI Text Fragments model is designed for loose fuzzy matching, or at least not so dependent on strict DOM structural constraints, so this could help too (although we know from experience that the Google Chrome polyfill is not super reliable, and web browsers lack a native API for this)

PR #3673, opened 2026-06-23
update settings tabtitle when locale changes

### LINKED ISSUE #3672: The title of the Settings modal does not immediately change if the user modifies the language
Steps:
- open the Settings modal; you are in the General tab.
- check the title of the modal
- modify the language
- check the title of the modal: it has not changed.
- move to another tab in this modal (e.g. Appearence)
- move back to the General tab
- check the title of the modal: it is now properly localized

arthur-lemeur:
Fixes #3672

### DIFF
--- a/src/renderer/library/components/settings/Settings.tsx
+++ b/src/renderer/library/components/settings/Settings.tsx
@@ -65,6 +65,7 @@ import { convertMultiLangStringToString } from "readium-desktop/common/language-
 import * as VisuallyHidden from "@radix-ui/react-visually-hidden";
 import moment from "moment";
 import os from "node:os";
+import { TTranslatorKeyParameter } from "readium-desktop/typings/en.translation-keys";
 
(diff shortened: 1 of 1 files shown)


src/renderer/library/components/settings/Settings.tsx
```diff
@@ -1002,7 +1003,9 @@ export const Settings: React.FC<ISettingsProps> = () => {
     const locale = useSelector((state: ICommonRootState) => state.i18n.locale);
     const isRTL = langStringIsRTL(locale);
 
-    const [tabTitle, setTabTitle] = React.useState(__("settings.tabs.general"));
+    const [activeTabKey, setActiveTabKey] = React.useState<TTranslatorKeyParameter>("settings.tabs.general");
```
panaC: This is breaking the `npm run i18n:typed` script, a local need to be called by the translator function

panaC (changes requested):

PR #3683, opened 2026-06-25
create whats new link and delete wizard

### LINKED ISSUE #3681: When a new version has been installed, push users to check what's new
The what's new wizard developed for v3.0 and still present in v3.4 is complex to maintain, and can be annoying for users (some to not click "do not show"). Textual content is localized raw text, which does not allow detailing new features; it is uneasy to add illustrations. There are currently two tabs, the second lists resources, with a hard coded Discord link that expired too soon. 

All we need, in practice, is a way to push users on a newly installed version to check the online what's new section of the Thorium website. 

We will then remove the wizard codebase (it will still be in the GitHub history is later we change our mind) and use  as a notification the same artefact we display when a new version is available, i.e., a sentence in the bottom of the home screen, with a link to the release notes. 

For reference, the message displayed when a new version is available:
Image

The sentence will be, in English: "A new version was installed, click to learn more." followed by the current version "(3.5.0)".  It will be localized (the key could be named "after", in the "update" section). 

Following the link updates a status that will avoid any new display of the message. A persistent state is therefore put in place, and reset when a new version is installed (the state is open_350). 

Release notes have a templated syntax: `https://www.thoriumreader.com/release-notes/desktop/` and this syntax is stabilized "forever". The user is redirected to the localized page corresponding to his brower settings, with a fallback on English. 

It is safer to use the parent page of specific release notes as a hyperlink, because we may not create a page for each patch version we release. Release notes are presented in ante-chrono order, therefore it is ok for users. 

As part of the release process, user-optimized release notes will need to be in place on thoriumreader.com BEFORE we release a new version.

arthur-lemeur:
Fixes #3681


src/renderer/library/components/catalog/AboutThoriumButton.tsx
```diff
@@ -114,6 +117,33 @@ class AboutThoriumButton extends React.Component<IProps, IState> {
...
+                                    const os = encodeURIComponent_RFC3986(osName);
+                                    const href = `https://thorium.edrlab.org/?lang=${locale}&v=${app_version}&source=${source}&os=${os}`;
+                                    if (href && /^https?:\/\//.test(href)) { /* ignores file: mailto: data: thoriumhttps: httpsr2: thorium: opds: etc. */
+                                        shell.openExternal(href).then(() => { /* noop */ }).catch((err: unknown) => { console.log(err); }); // .finally(() => { /* noop */ });
+                                    }
+                                }).catch((err: unknown) => { console.log(err); });
+                            }}>
+                                {/* TODO: add a translation for this string */}
+                                {/* {`${this.props.__("app.whatsnew.message", { version: _APP_VERSION })}`} */}
+                                See what is new in Thorium {`v${_APP_VERSION}`}
```
danielweck: this needs translating.

src/renderer/library/components/catalog/AboutThoriumButton.tsx
```diff
@@ -258,6 +289,9 @@ const mapDispatchToProps = (__dispatch: TDispatch, _props: IBaseProps) => {
         // openReader: (publicationView: PublicationView) => {
         //     dispatch(readerActions.openRequest.build(publicationView.identifier));
         // },
+        whatsNewOpened: () => {
+            __dispatch(whatsNewActions.setWhatsNew.build(true));
```
danielweck: `__dispatch` needs to be renamed to `dispatch`

src/common/redux/states/renderer/libraryRootState.ts
```diff
@@ -38,6 +38,6 @@ export interface ILibraryRootState extends IRendererCommonRootState {
             userDirectory?: string,
         }
     };
-    wizard: IWizardState;
+    whatsNew: IWhatsNewState;
```
danielweck, line 41: this is essentially a breaking change in the "database schema" of the application, we'll need to make sure there are no regressions when upgrading to 3.5.0 and reverting back to 3.4.0

danielweck (changes requested):

PR #3695, opened 2026-07-03
add language tagging to pub title and description

### LINKED ISSUE #3691: Language tagging in the user interface
Originally identified by @jonaslil at https://github.com/daisy/transitiontoepub/issues/92#issuecomment-4842743751

> While reading an EPUB with a screen reader, if the publication language differs from the system language (for example, an English book on a Finnish Windows installation, using Thorium + NVDA), some content is announced using the system language voice, even when the language is correctly marked up in the publication. Examples include:
> 
> Publication content:
> 
> Image alt text
> Table accessible name (derived from the caption) when navigating directly to the table
> Table headers when navigating between data cells
> ARIA labels
> Navigation items and metadata:
> 
> Navigation table of contents
> Publication title when opening the book
> Publication metadata

I can confirm that reading an English title with Thorium setted up in French, I observe no language tagging in the Book Information Panel.

Image

arthur-lemeur:
Fixes #3691


src/renderer/common/components/dialog/publicationInfos/publicationInfoContent.tsx
```diff
@@ -441,7 +441,7 @@ export const PublicationInfoContent: React.FC<React.PropsWithChildren<IProps>> =
                 <div className={stylePublication.publicationInfo_rightSide}>
                     <section>
                         <h2 className={classNames(stylesBookDetailsDialog.allowUserSelect, stylesGlobal.my_10, stylePublication.book_title)}
-                            dir={pubTitleIsRTL ? "rtl" : undefined}>
+                            dir={pubTitleIsRTL ? "rtl" : undefined} lang={publicationViewMaybeOpds.languages ? publicationViewMaybeOpds.languages?.[0] : undefined}>
```
danielweck: I don't think this is correct, please see the existing `pubTitleLang` variable in this code (this is already used to compute `pubTitleIsRTL`)

src/renderer/common/components/dialog/publicationInfos/PublicationInfoDescription.tsx
```diff
@@ -28,93 +28,66 @@ interface IProps {
...
+    }, []);
+
+    React.useEffect(() => {
+        const timeout = setTimeout(needSeeMoreButton, 500);
+        return () => clearTimeout(timeout);
+    }, [publicationViewMaybeOpds, needSeeMoreButton]);
+
+    React.useEffect(() => {
+        const lang = publicationViewMaybeOpds.languages?.[0];
+        if (!lang || !descriptionRef.current) return;
```
danielweck: I think this needs to align with the "title" approach (both `lang` and `dir` can be set on the HTML text container):

```
    const locale = useSelector((state: ICommonRootState) => state.i18n.locale);
    const pubTitleLangStr = convertMultiLangStringToLangString((publicationViewMaybeOpds as PublicationView).publicationTitle || publicationViewMaybeOpds.documentTitle, locale);
    const pubTitleLang = pubTitleLangStr && pubTitleLangStr[0] ? pubTitleLangStr[0].toLowerCase() : "";
    const pubTitleIsRTL = langStringIsRTL(pubTitleLang);
    const pubTitleStr = pubTitleLangStr && pubTitleLangStr[1] ? pubTitleLangStr[1] : "";
```

danielweck (changes requested):

PR #3703, opened 2026-07-13
feat(a11y): screen reader activation based on auto-detection plus user approval, GUI affordances to announce option (before skip link, and in status bar) (Fixes #3349 )

### LINKED ISSUE #3349: Smarter way to activate screen reader support
As reported by screen reader users, other software electron based like VSCode act as follow:

If a screen reader is detected, a context window ask the user "do you want to activate screen reader support"? 

That makes the option visible for readers who need it.

danielweck:
Fixes https://github.com/edrlab/thorium-reader/issues/3349


danielweck:
MacOS screenshots that show the default keyboard shortcut `SHIFT` + `CTRL` + `F12` changed to `F9`, with the screen reader (Voice Over) announcing the tabbable toggle button at the same level as the "skip link" for early discovery, and also with the status bar ARIA live assertive announcement (there is a toast notification to confirm yes/no activation of the optimized screen reader support)

Screenshot 2026-07-13 at 11 29 14 am
Screenshot 2026-07-13 at 11 29 17 am

PR #3711, opened 2026-07-20
fix behavior issue with radix popover

arthur-lemeur:


danielweck:
See https://github.com/edrlab/thorium-reader/issues/3707#issuecomment-5061795662

danielweck:
See https://github.com/edrlab/thorium-reader/issues/3721

PR #3714, opened 2026-07-21
change h3 to span in custom checkbox labels

### LINKED ISSUE #3713: Options Menu for Annotations and Bookmarks
Each option currently appears in bold because it is an h3. It is not good as an h3 because no explanation is given below the label. This is different from the general Settings (where each option title is followed by an explanation. 

Solution: replace the h3 on each label with  a span, keep the aria-hidden attribute, both in Annotations and Bookmarks. 

Image

arthur-lemeur:
Fixes #3713

### DIFF
--- a/src/renderer/assets/styles/global.scss
+++ b/src/renderer/assets/styles/global.scss
@@ -468,6 +468,7 @@ nav + nav + .breadcrumb + .main {
     align-items: center;
     gap: 10px;
     cursor: pointer;
+    margin-bottom: 10px;
 
     h4 {
       margin: 10px 0;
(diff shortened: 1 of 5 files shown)


danielweck:
feel free to merge if you are confident that the h3 CSS selector wasn't injecting any important styling directive that the span replacement will void (for example, text ellipsis etc. just to name a concrete example of potential breaking cascade change)

danielweck:
will the additional 10px bottom margin disrupt other components?
https://github.com/search?q=repo%3Aedrlab%2Fthorium-reader+checkbox_custom&type=code

PR #3717, opened 2026-07-22
revert to previous radix version for the popover to fix bug

### LINKED ISSUE #3707: regression in GUI context menu popout
menus don't close when opening a new one

Image

arthur-lemeur:
Fixes #3707

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@github/paste-markdown": "^1.5.3",
         "@radix-ui/react-alert-dialog": "^1.1.19",
         "@radix-ui/react-dialog": "^1.1.19",
-        "@radix-ui/react-popover": "^1.1.19",
+        "@radix-ui/react-popover": "^1.1.15",
         "@radix-ui/react-radio-group": "^1.4.3",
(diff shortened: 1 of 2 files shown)


danielweck:
We need to version-lock the top-level direct dependency `@radix-ui/react-popover` in `package.json` (not just `package-lock.json`) at `1.1.15` instead of `^1.1.15`. Then, we also need to check that there are no other transitive `@radix-ui/react-popover` dependencies in the tree, using the NPM command line `npm list | grep xxx` or just inspecting the resulting lockfile. I will do this.

danielweck:
Side note: this triggered my mild dyslexia `@radix-ui/react-popover` vs. `@radix-ui/react-popper` when inspecting`package-lock.json` :D

danielweck:
> We need to version-lock the top-level direct dependency `@radix-ui/react-popover` in `package.json` (not just `package-lock.json`) at `1.1.15` instead of `^1.1.15`. Then, we also need to check that there are no other transitive `@radix-ui/react-popover` dependencies in the tree, using the NPM command line `npm list | grep xxx` or just inspecting the resulting lockfile. I will do this.

working on this...

PR #3747, opened 2026-08-04
Highlight active OPDS navigation entries

### LINKED ISSUE #3697: OPDS facets rel="self" not handled as the currently active facet
https://github.com/edrlab/thorium-reader/blob/6614a10efff935ad94b9fa717d035079d6466578/src/main/converter/opds.ts#L674

https://specs.opds.io/opds-2.0.html#24-facets:~:text=may%20use%20self%20in%20the%20rel%20of%20a%20Link%20Object%20to%20indicate%20which%20facet%20is%20currently%20active

> may use self in the rel of a Link Object to indicate which facet is currently active

panaC:
Fixes https://github.com/edrlab/thorium-reader/issues/3697

image

### DIFF
--- a/src/common/views/opds.ts
+++ b/src/common/views/opds.ts
@@ -85,6 +85,7 @@ export interface IOpdsNavigationLinkView {
     subtitle?: string;
     url: string;
     numberOfItems?: number;
+    active?: boolean;
 }
 
 export interface IOpdsFeedMetadataView {
--- a/src/main/converter/opds.ts
+++ b/src/main/converter/opds.ts
@@ -698,8 +698,10 @@ export class OpdsFeedViewConverter {
             : "";
 
         const links = r2OpdsFacet.Links?.map(
-            (item) =>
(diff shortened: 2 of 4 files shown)


danielweck (approved):

panaC:
I used `aria-current` because the OPDS spec says `rel="self"` identifies the facet that is currently active, and `aria-current` is the ARIA attribute meant to expose “this item is the current one in a set of links” to assistive tech.

So in this UI, the facet options are rendered as links. Marking the active one as:

```tsx
aria-current="page"
```

lets screen readers announce it as the current destination/filter, while CSS can also style it visibly.

That said, `page` may be a little too page-oriented for a facet/filter. A more neutral choice could be:

```tsx
aria-current="true"
```

That would still communicate “current item” without implying page navigation. I’d probably switch to `true` here for semantic precision.

panaC:
@danielweck What do you think ?

arthur-lemeur (approved):

danielweck:
agreed :) thank you

PR #3749, opened 2026-08-04
fix size of viewer dialog and scale on images to center in the modal

arthur-lemeur:


danielweck (changes requested):
just need to fix linting error.
otherwise LGTM

PR #3748, opened 2026-08-04
Add EPUBCFISelector target for annotation selectors

panaC:
Fixes https://github.com/edrlab/thorium-reader/issues/3709

PR #3750, opened 2026-08-04
handling of CFI selectors during annotation import/export

panaC:
This PR updates the handling of CFI selectors during annotation import/export.

    On import, both the legacy CfiSelector and the standard EPUBCFISelector are supported.
    On export, only EPUBCFISelector, which is compliant with the specification, is generated.
    Removed the legacy Readium annotation export of FragmentSelector with the CFI conformsTo value.

Ref: from this commit https://github.com/readium/annotations/commit/f52c729e78e77aadf675fe0daf5ba99eae20a858


panaC:
Closes #3748

danielweck (approved):

Issue #3523, opened 2026-04-22
Implement TTS verbose mode


gautierchomel:
_This is a summary of why and how to implement TTS verbose mode._

The existing Thorium TTS options already offer a simplified reading experience: “skippability” that hides auxiliary content such as footnotes, endnotes, page breaks, and asides. A verbose mode is the complementary idea: instead of skipping content, it would allow users to deliberately hear more structure meaningful for comprehension, when needed in scholarly, academic and advanced reading.

## What it should do and solve
The goal is to surface only cues that help distinguish, understand, or navigate the publication, which would include:

1. Read extended descriptions attached to images, then continue just after their reference, so the reading jumps through that section while staying in the logical order of the document.
2. When it reaches a details block, consider it expanded and read its full content.
3. Expose a selection of semantics labels as short, predictive cues, like page numbers and page breaks as “Page break” or “Page {n}”, to help you track the physical or logical pagination of the book.
4. Speak long-form of abreviations.

<details><summary>The proposal was initially made in discussion #2975. It covers or was mentioned in issues: </summary>

- #1717: aria-label on generic elements with overridden ARIA role should take precedence over raw text content in TTS.
- #1933, #1974, #1983, #1311 (and related TTS/reading‑aloud tickets): allow configurable verbosity for auxiliary content such as footnotes, endnotes, page breaks, notes, examples, and semantic labels.
- #1923, #2099, #2655, and #2679 are asking for granular skipability and refer to the reading of footnotes and endnotes.
</details>

## What it should not do
Indiscriminately read all available semantics. It should not expose raw ARIA or internal markup. Its purpose is to surface meaningful structural and auxiliary cues when helpful, not to turn TTS into a noisy technical trace of the document.

Proposing a cherry-picked list of what to read or not would create a disproportionate burden and inflate the setting panel, which is already large.

Try to differentiate added verbosity from the main content. This is not possible, as it would result in excessive detail or a shift in voice that would negatively impact the experience.

Remediate confusing, poorly used ARIA or HTML; this is an authoring responsibility. Non-visual reading experiences must not include redundant or decorative labels. If a label simply restates visible text without adding new context, it is an authoring error, as it duplicates information. We can, however, add a disclaimer using a tooltip at the setting level.

## How to implement it
Proposed implementation is to add a second option in Settings / Readaloud next to skippability. Each option should be accompanied by information describing its behaviours:

- Skipability: suppress auxiliary content (aside, footnotes, endnotes).
[...]

Issue #1717, opened 2022-07-04
TTS readaloud: aria-label on generic elements with overridden ARIA role, still speaks element textual content


danielweck:
(aria-label should take precedence, just as AT/screen readers do)

pkra:
> (aria-label should take precedence, just as AT/screen readers do)

But please be aware that aria-label is not allowed for all roles, cf. the exclude list at https://w3c.github.io/aria/#aria-label. It's a common authoring error is to add aria-label attributes when they aren't allowed.

In that case, browsers should ignore aria-label when building the accessibility tree and thus AT would not have access to it. (And the reality is much more complicated still.)

danielweck:
Ha ha, sounds like a sweet rabbit hole. Thanks for letting us know :)

danielweck:
Related issue: https://github.com/readium/r2-navigator-js/issues/76

PR #3781, opened 2026-08-17
feat(tts): add a "Read subscripts and superscripts" preference

whosawhatsis:
Refs #3778. Depends on #3779 — stacked on that branch, so until it
merges this PR's diff will also show its commits. Independent of the script-boundary PR.

## Why

#3779 makes `sub`/`sup` content spoken by default, which is the correct default:
in `H₂O`, `x²` and `R₀` the content is meaning, and dropping it silently is a bug.

But there is a real class of books where the current behaviour is what a listener wants —
those that mark footnotes with a bare `<sup>1</sup>` carrying no `epub:type="noteref"`
and no `role="doc-noteref"`. Nothing in the markup distinguishes those from an exponent,
so no heuristic can skip them, and the listener gets a bare digit interrupting every few
sentences.

The skippability mechanism already handles *properly marked* note references, and
#3779 adds `"noteref"` to `_skippables` so they are covered. This preference is
the escape hatch for the improperly marked ones.

## What it does

Adds `ttsEnableSubSup` to the reader config, **default `true`**, surfaced in the
read-aloud settings panel as a checkbox:

> **Read subscripts and superscripts**
> Speak the contents of subscripts and superscripts, such as exponents and indices.
> Turn off for books that use raw superscripts as footnote markers.

Turned off, `generateTtsQueue()` output is **byte-identical to `develop`** — the tags go
back to being skipped wholesale. So this is a true revert switch, not a third behaviour
to reason about.

## Shape of the change

Deliberately modelled on the existing `ttsEnableSentenceDetection` /
`mediaOverlaysEnableSkippability` preferences, so it lands in the same places and needs
no new patterns — 13 files, +86/−7 for this PR's own commit. (Because it is stacked,
GitHub will show 14 files and +196/−15 until #3779 merges — the difference is #3779's
two files.)

| area | file | what |
|---|---|---|
| config model | `common/models/reader.ts`, `common/redux/states/reader.ts` | field + default `true` |
| IPC | `r2-navigator-js/electron/common/events.ts` | `R2_EVENT_TTS_SUBSUP_ENABLE` + payload |
| webview | `webview/state.ts`, `webview/preload.ts` | state field, init, reset, IPC handler |
| renderer | `renderer/readaloud.ts`, `renderer/dom.ts` | `ttsSubSupEnable()`, init on load |
| consumer | `renderer/common/dom-text-utils.ts` | gates the deep-dive |
| saga | `reader/redux/sagas/readerConfig.ts` | wires the setter; added to the TTS-restart condition |
| UI | `reader/components/ReadingAudio.tsx` | the checkbox |
| i18n | `resources/locales/en.json`, regenerated `typings/en.translation*.d.ts` | label + description |

Only `en.json` is populated; the other locales fall back, as usual for a new key.

### The one subtle part

The preference gates whether the element is **descended into**, not whether it is
recognised as a `sub`/`sup`:

```ts
// isSubSup is the tag test only - it must stay independent of the preference
const isSubSup = childTagNameLow === "sup" || childTagNameLow === "sub";
let subSupNeedsDeepDive = isSubSup && win.READIUM2.ttsSubSupEnabled && !hidden;
```

They have to stay separate because #3779 removes `sup`/`sub` from
`_doNotProcessDeepChildTagNames`. If the preference were folded into `isSubSup`, then
with the preference **off** the element would no longer be recognised as a sub/sup at
all, would fall through to the generic path, and would be read anyway — the exact
opposite of what the switch is for. Worth a glance in review; it is the one place where
the obvious simplification is wrong.

Note also that the saga must restart TTS when this changes (it is in the restart
condition alongside `ttsEnableOverlayMode` and `ttsEnableSentenceDetection`), because the
utterance queue is built once at play time — without that, toggling mid-playback appears
to do nothing until the next document.

## Verification

- `tsc --noEmit` exit 0, `eslint` clean
- Preference **on**: the test fixture from #3778 reads every case as its own prose announces
- Preference **off**: `generateTtsQueue()` output compared item-by-item against
  `develop` over the same fixture — identical
- Toggling mid-playback restarts TTS and takes effect immediately

## Note on scope

If the maintainers would rather ship #3779 without a new preference, that is a
reasonable call and this PR can be dropped — #3779 stands on its own. This exists
because "we changed what your books sound like and you cannot change it back" is a
harder sell than "here is the switch".


danielweck:
Related "verbosity" topic: https://github.com/edrlab/thorium-reader/issues/3523

whosawhatsis:
Thanks for the pointer — I read #3523 properly, and I think it means **this PR is packaged wrong**, so I would rather say so than have you talk me out of it in review.

Two direct collisions:

1. #3523 says a cherry-picked list of what to read "would create a disproportionate burden and inflate the setting panel, which is already large", and proposes **one** new axis (Verbosity) beside Skippability. This PR adds a third checkbox specific to `sub`/`sup` — precisely the kind of per-feature toggle that argues against.
2. @gautierchomel proposes verbosity be three-level (`some` / `most` / `all`). A boolean sub/sup switch does not compose with that; it would have to become one input to a level, not its own control.

So I would suggest one of:

- **(a)** Close this PR. #3779 stands on its own — with it, sub/sup content is spoken, which is the correct default and matches every other inline element. The escape hatch then arrives with verbosity rather than ahead of it.
- **(b)** Keep the mechanism, drop the UI. The plumbing here (`ttsEnableSubSup` through config → saga → IPC → webview, with the TTS-restart on change) is what a verbosity level would need anyway; it could land without the checkbox and be wired to a level later.
- **(c)** Leave it open as a marker for the sub/sup case in #3523 and close it when that lands.

My preference is **(a)** or **(b)** — I do not think a standalone toggle should ship ahead of the design in #3523. I am happy to do the work either way; just say which and I will push it.

One thing worth separating out: what the preference is really an escape hatch *for* is unmarked footnote markers — a bare `<sup>1</sup>` with no `epub:type="noteref"` and no `role="doc-noteref"`. That is a **skippability** problem, not a verbosity one, and it is unsolvable from markup alone, which is why I reached for a preference. If verbosity ends up levelled, this case may still need somewhere to live on the skippability side. Test 15 in the test EPUB from #3778 (the build script is inlined at the bottom of that issue) is exactly that specimen, if it is useful for #3523.

Unrelated, and I did not want to bury it: #3523 item 3 and #1717 both describe `aria-label` as being read **before** an element's content. #3779 currently has `aria-label`/`title` on a `sup`/`sub` **replace** the content (`<sup aria-label="squared">2</sup>` → "squared", not "squared 2"). For an exponent, replacing reads better to me, but it diverges from the model in #3523 and I would rather align with whatever you settle on there than have two conventions in the same file.

PR #3783, opened 2026-08-17
fix(tts): insert a word boundary between separable scripts

### LINKED ISSUE #3782: Read-aloud drops words that mix Greek or Cyrillic with Latin letters (macOS)
## Summary

Speech engines **discard** a token that mixes two alphabetic scripts. This is not
mispronunciation, it is silent omission:

> "the change Δx is small"  →  *"the change is small"*

Any book that writes Greek variables inline is affected, which in mathematical and
scientific prose is most of them. As with `sub`/`sup`, the listener gets no cue that
anything was dropped — the sentence is simply, quietly wrong.

## Evidence

Synthesised speech timed on macOS with Samantha, Alex and Daniel (identical results
across all three):

| utterance | duration |
|---|---|
| `we call it here` (nothing at all) | 1.070s |
| `we call it Δx here` | **1.163s** — the token is not spoken |
| `we call it Δ x here` | 2.173s |
| `we call it дx here` (Cyrillic) | **1.163s** — identical behaviour |
| `we call it д x here` | 2.173s |

Cyrillic behaving exactly like Greek is what makes this a *script-boundary* effect
rather than a Greek quirk, and inserting a word boundary at script transitions is the
established remedy for it.

## Platform scope

**The defect is macOS-only.** I measured all three desktop platforms afterwards and it
does not reproduce on either of the other two. Timings are of the Web Speech API, the
same path Thorium uses, so each column is that platform's real engine — macOS Samantha,
espeak-ng via speech-dispatcher on Linux, SAPI `Speech_OneCore` under Chrome 150 on
Windows:

| utterance | macOS | Linux / espeak-ng | Windows / SAPI |
|---|---|---|---|
| baseline, no symbol | 1.070s | 1.297s | 1.746s |
| `we call it θx here` | **1.163s — dropped** | 1.810s — spoken | 2.296s — spoken |
| `we call it θ x here` | 2.173s | 1.927s | 2.358s |
| `we call it дx here` | **1.163s — dropped** | 1.778s — spoken | 2.048s — spoken |
| `we call it д x here` | 2.173s | 1.788s | 2.057s |

On Linux and Windows the fused and separated forms are within noise of each other and
both well clear of baseline — the token is spoken either way, so this change buys
nothing there. What it costs there is also close to nothing: a slightly different
cadence at the boundary, no content changed, nothing dropped or mispronounced.

Two consequences for any fix. A separator inserted at the boundary is **effective on
macOS and near-free elsewhere** — on Linux and Windows the only measurable effect is a
slightly different cadence, with no content changed and nothing mispronounced. So a fix
could reasonably apply it unconditionally, keeping behaviour uniform across platforms, or
gate it on `process.platform === "darwin"` to touch only the platform with the defect.
The trade is uniform-and-simple against precise-but-platform-conditional.

The third option is to decline to work around it at all: an author can insert a
zero-width separator in the EPUB itself, which is what I did for my own book before
looking at Thorium's source. That leaves every other book affected.

## Scope of a fix — what should and should not be separated

Only **Latin, Greek and Cyrillic**, and the rule is symmetric between them — a
Greek/Cyrillic transition is separated just as a Greek/latin one is. It is a script-pair
rule, not a latin-centric one.

### Why this is not, and should not be, script-agnostic

The obvious generalisation is "insert a boundary at every script transition". That
destroys Japanese, which mixes Han, Hiragana and Katakana **inside single words**:

| text | script-agnostic rule would produce |
|---|---|
| `食べる` (one verb) | `食 べる` |
| `東京タワー` | `東京 タワー` |
| `私はiPhoneを使います` | `私 は iPhone を 使 います` |

So the separable set has to be an explicit list, and CJK cannot be in it.

### Why CJK/latin is excluded even as a special case

Grouping Han/Hiragana/Katakana together and separating only that group from latin is
more tempting, but measurement says no. With a Japanese voice (Eddy, ja_JP):

| utterance | duration |
|---|---|
| `私はiPhoneを使います` | 2.990s |
| `私は iPhone を使います` | **3.358s** |
| `私はそれを使います` (no latin at all) | 1.966s |

Two conclusions. The latin word is **not dropped** when fused — 2.990s against 1.966s
for the same sentence without it — so there is no defect to fix here. And the spaced
variant is 12% longer, which is an inserted prosodic pause: the separator is *not* free.

That is the opposite trade from Greek and Cyrillic. Mixed Greek/latin runs are rare and
are always notation, so a boundary costs nothing and rescues content that is otherwise
lost outright. Mixed CJK/latin runs are ordinary prose that CJK readers encounter
constantly, and a boundary there would degrade everyday reading to hedge against a
defect not observed on this platform.

If another engine does drop CJK/latin tokens, the fix is to add the pair to
`_separableScripts` on that evidence — the list is a one-line extension point. Scripts
with no measurements are left alone rather than guessed at.

Same-script text is untouched by construction, so these are all unaffected — verified,
not assumed:

| specimen | why unchanged |
|---|---|
| `caféx` | accented latin is still latin |
| `Δθ` | one script; already spoken correctly |
| `Δ2` | digits are not a separable script |
| `(ρ)` | punctuation, not a script boundary |
| `𝐅x` | Mathematical Alphanumeric Symbols are script Common |
| `漢x` | excluded, see above |

### Zero-width formatting characters between the scripts

Skipped when tracking script runs, rather than being allowed to end one. If a character
sits between the two scripts it would otherwise hide the boundary behind it — and that
matters, because not all of them separate the token for the engine:

| between `Δ` and `x` | engine speaks it? | boundary found? |
|---|---|---|
| nothing | ✗ dropped | ✓ added |
| U+2060 WORD JOINER | ✗ **still dropped** | ✓ added — this is the case that needs it |
| U+FEFF, U+200B | ✓ already separated | ✓ added, redundant but harmless |
| a real space | ✓ | ✗ correctly none — a space is not default-ignorable |

WORD JOINER is the awkward one: an author might use it precisely to stop `Δx` breaking
across a line, and it leaves the token fused, so without this the content stays dropped
and the fix never reaches it.

Where a separator is already present the doubled boundary is acoustically identical —
`Δ x`, `Δ<FEFF>x` and `Δ<FEFF> x` all measure 2.197s.

## To reproduce

The script that builds a test EPUB is at the bottom of this issue, in a collapsed block —
106 lines of `bash`, no dependencies beyond `zip`. It writes
`tts-script-boundary-testcase.epub`, which validates clean under epubcheck. Eleven cases:
four that a fix should change, seven that it must not. Each states what it should sound
like in the same sentence as the specimen, so it checks itself by ear.

**On macOS this is easier to observe than the `sub`/`sup` issue in #3778**, because the
Greek characters do reach the read-aloud display — you can watch the highlight travel
across `Δx` and hear nothing at all.

On Linux and Windows every case already sounds correct, so there is nothing to observe
there.

## Pull request

- #3783 — inserts a word boundary at Latin/Greek/Cyrillic transitions.
  Stacked on #3779, so its diff shows that PR's commits until it merges.

## Relationship to #3778

Different bug, filed separately on purpose. #3778 is content Thorium itself discards and
reproduces on every platform. This one is a speech-engine defect that Thorium could work
around, and only on macOS. Neither depends on the other.

---

<details>
<summary><b>build-script-boundary.sh</b> — generates <code>tts-script-boundary-testcase.epub</code> (click to expand)</summary>

```bash
#!/bin/bash
# Builds a minimal EPUB 3 demonstrating that speech engines drop a token which mixes
# two alphabetic scripts, e.g. a Greek variable fused to a latin letter.
set -euo pipefail
cd "$(dirname "$0")"
OUT="${1:-$PWD/tts-script-boundary-testcase.epub}"
rm -rf src2 "$OUT"
mkdir -p src2/META-INF src2/EPUB

printf 'application/epub+zip' > src2/mimetype

cat > src2/META-INF/container.xml <<'XML'
<?xml version="1.0" encoding="UTF-8"?>
<container version="1.0" xmlns="urn:oasis:names:tc:opendocument:xmlns:container">
  <rootfiles>
    <rootfile full-path="EPUB/package.opf" media-type="application/oebps-package+xml"/>
  </rootfiles>
</container>
XML

cat > src2/EPUB/package.opf <<'XML'
<?xml version="1.0" encoding="UTF-8"?>
<package xmlns="http://www.idpf.org/2007/opf" version="3.0" unique-identifier="uid" xml:lang="en">
  <metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
    <dc:identifier id="uid">urn:uuid:5f6c1b8e-0a2d-4f6b-9c3a-1b2c3d4e5f61</dc:identifier>
    <dc:title>TTS script boundary test case</dc:title>
    <dc:language>en</dc:language>
    <dc:creator>Read-aloud regression fixture</dc:creator>
    <meta property="dcterms:modified">2026-01-01T00:00:00Z</meta>
    <meta property="schema:accessMode">textual</meta>
    <meta property="schema:accessModeSufficient">textual</meta>
    <meta property="schema:accessibilityFeature">structuralNavigation</meta>
    <meta property="schema:accessibilityHazard">none</meta>
    <meta property="schema:accessibilitySummary">Plain text test fixture for text-to-speech behaviour.</meta>
  </metadata>
  <manifest>
    <item id="nav" href="nav.xhtml" media-type="application/xhtml+xml" properties="nav"/>
    <item id="t1" href="tests.xhtml" media-type="application/xhtml+xml"/>
  </manifest>
  <spine><itemref idref="t1"/></spine>
</package>
XML

cat > src2/EPUB/nav.xhtml <<'XML'
<?xml version="1.0" encoding="UTF-8"?>
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en" lang="en">
<head><meta charset="utf-8"/><title>Contents</title></head>
<body><nav epub:type="toc" id="toc"><h1>Contents</h1>
<ol><li><a href="tests.xhtml">Test cases</a></li></ol></nav></body>
</html>
XML

cat > src2/EPUB/tests.xhtml <<'XML'
<?xml version="1.0" encoding="UTF-8"?>
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
<head><meta charset="utf-8"/><title>TTS script boundary test cases</title></head>
<body>
<h1>Read-aloud test cases for mixed-script tokens</h1>

<p>Start read-aloud at the top. Each test says what you should hear, then the
specimen follows immediately.</p>

<h2>A. Currently dropped</h2>

<p>Test one. Should hear: delta, ex. Specimen: &#x394;x.</p>

<p>Test two. Should hear: dee, theta. Specimen: d&#x3B8;.</p>

<p>Test three. Cyrillic rather than Greek, same effect. Should hear: the Russian
letter de, then ex. Specimen: &#x434;x.</p>

<p>Test four. In a full sentence. Should hear the whole sentence including the
symbol: the change delta ex is small. Specimen: the change &#x394;x is small.</p>

<h2>B. Should not change</h2>

<p>Test five. A Greek letter on its own, already spoken correctly. Should hear:
the density, rho, of the air. Specimen: the density (&#x3C1;) of the air.</p>

<p>Test six. Two Greek letters together, one script, already correct. Should hear:
delta theta. Specimen: &#x394;&#x3B8;.</p>

<p>Test seven. Greek followed by a digit, already correct. Should hear: delta two.
Specimen: &#x394;2.</p>

<p>Test eight. Accented latin is still latin, so no boundary belongs here. Should
hear one word, not two. Specimen: caf&#xE9;x.</p>

<p>Test nine. A Han character next to a latin one. Measures identically with or
without a boundary, so it is deliberately left alone. Specimen: &#x6F22;x.</p>

<p>Test ten. Mathematical bold capital F, which is script Common rather than latin.
Should be unchanged. Specimen: &#x1D405;x.</p>

<p>Test eleven. Ordinary prose with no mixed-script token anywhere in it. Should
hear this sentence exactly as written.</p>
</body>
</html>
XML

cd src2
zip -qX0 "$OUT" mimetype
zip -qXr9D "$OUT" . -x mimetype
cd ..
rm -rf src2
echo "wrote $OUT"
```

</details>

whosawhatsis:
Fixes #3782. Depends on #3779 — stacked on that branch, since it generalises the
separator mechanism introduced there. Independent of #3781.

## The problem

Speech engines **discard** a token that mixes two alphabetic scripts. This is not
mispronunciation, it is silent omission:

> "the change Δx is small"  →  *"the change is small"*

Any book that writes Greek variables inline is affected, which in mathematical and
scientific prose is most of them. As with `sub`/`sup`, the listener gets no cue that
anything was dropped — the sentence is simply, quietly wrong.

## Evidence

Synthesised speech timed on macOS with Samantha, Alex and Daniel (identical results
across all three):

| utterance | duration |
|---|---|
| `we call it here` (nothing at all) | 1.070s |
| `we call it Δx here` | **1.163s** — the token is not spoken |
| `we call it Δ x here` | 2.173s |
| `we call it дx here` (Cyrillic) | **1.163s** — identical behaviour |
| `we call it д x here` | 2.173s |

Cyrillic behaving exactly like Greek is what makes this a *script-boundary* effect
rather than a Greek quirk, and inserting a word boundary at script transitions is the
established remedy for it.

## What is and isn't separated

Only **Latin, Greek and Cyrillic**, and the rule is symmetric between them — a
Greek/Cyrillic transition is separated just as a Greek/latin one is. It is a script-pair
rule, not a latin-centric one.

### Why this is not, and should not be, script-agnostic

The obvious generalisation is "insert a boundary at every script transition". That
destroys Japanese, which mixes Han, Hiragana and Katakana **inside single words**:

| text | script-agnostic rule would produce |
|---|---|
| `食べる` (one verb) | `食 べる` |
| `東京タワー` | `東京 タワー` |
| `私はiPhoneを使います` | `私 は iPhone を 使 います` |

So the separable set has to be an explicit list, and CJK cannot be in it.

### Why CJK/latin is excluded even as a special case

Grouping Han/Hiragana/Katakana together and separating only that group from latin is
more tempting, but measurement says no. With a Japanese voice (Eddy, ja_JP):

| utterance | duration |
|---|---|
| `私はiPhoneを使います` | 2.990s |
| `私は iPhone を使います` | **3.358s** |
| `私はそれを使います` (no latin at all) | 1.966s |

Two conclusions. The latin word is **not dropped** when fused — 2.990s against 1.966s
for the same sentence without it — so there is no defect to fix here. And the spaced
variant is 12% longer, which is an inserted prosodic pause: the separator is *not* free.

That is the opposite trade from Greek and Cyrillic. Mixed Greek/latin runs are rare and
are always notation, so a boundary costs nothing and rescues content that is otherwise
lost outright. Mixed CJK/latin runs are ordinary prose that CJK readers encounter
constantly, and a boundary there would degrade everyday reading to hedge against a
defect not observed on this platform.

If another engine does drop CJK/latin tokens, the fix is to add the pair to
`_separableScripts` on that evidence — the list is a one-line extension point. Scripts
with no measurements are left alone rather than guessed at.

Same-script text is untouched by construction, so these are all unaffected — verified,
not assumed:

| specimen | why unchanged |
|---|---|
| `caféx` | accented latin is still latin |
| `Δθ` | one script; already spoken correctly |
| `Δ2` | digits are not a separable script |
| `(ρ)` | punctuation, not a script boundary |
| `𝐅x` | Mathematical Alphanumeric Symbols are script Common |
| `漢x` | excluded, see above |

### Zero-width formatting characters between the scripts

Skipped when tracking script runs, rather than being allowed to end one. If a character
sits between the two scripts it would otherwise hide the boundary behind it — and that
matters, because not all of them separate the token for the engine:

| between `Δ` and `x` | engine speaks it? | boundary found? |
|---|---|---|
| nothing | ✗ dropped | ✓ added |
| U+2060 WORD JOINER | ✗ **still dropped** | ✓ added — this is the case that needs it |
| U+FEFF, U+200B | ✓ already separated | ✓ added, redundant but harmless |
| a real space | ✓ | ✗ correctly none — a space is not default-ignorable |

WORD JOINER is the awkward one: an author might use it precisely to stop `Δx` breaking
across a line, and it leaves the token fused, so without this the content stays dropped
and the fix never reaches it.

Where a separator is already present the doubled boundary is acoustically identical —
`Δ x`, `Δ<FEFF>x` and `Δ<FEFF> x` all measure 2.197s.

## Note on #3779's review

@danielweck has left review comments on #3779 proposing that the offset walker keep its
existing cascading-ternary shape with an `isSUBSUP ? (length + 2) : length` branch, and
that the `pendingSubSupSeparator` module state be replaced by a direct parent-tag test.
Both are good simplifications for #3779 taken alone.

They do partly conflict with this PR, which is why I am flagging it rather than letting it
be discovered: a fixed `+2` assumes exactly one separator at each end, and script
boundaries need a variable number at arbitrary offsets, hence the `__TTS_SEPS` array here.
The parent-tag suggestion applies cleanly either way and I will adopt it.

**I will rebase this onto whatever shape #3779 lands in**, and I am happy for that to mean
reworking the mechanism below, or dropping this PR entirely if you would rather not carry
a per-platform speech workaround. #3779 does not depend on this one.

## Implementation

Generalises the mechanism from #3779. Instead of a boolean meaning "fence both
ends", a text node carries `__TTS_SEPS` — the offsets at which `TTS_SEPARATOR` is
inserted when the utterance string is assembled. `sup`/`sub` contributes `[0, length]`,
script changes contribute their own offsets, and the two compose:
`<em>Δ</em><sub>x</sub>y` → `"Δ x y"`.

The separators exist only in the utterance string, **never in the DOM**, so rendering,
CFIs and locators are untouched.

Both offset walkers in `readaloud.ts` now share `ttsOffsetToDomOffset()`, which
subtracts the separators preceding an offset and clamps into the node. The `i`-th
separator sits at utterance offset `seps[i] + i`, which is what makes the subtraction a
simple count.

## Verification

- `tsc --noEmit` exit 0, `eslint` clean
- Every utterance offset mapped back to its DOM character across script boundaries,
  `sup`/`sub`, both combined, and all the excluded cases — **0 mismatches**, so word-
  and sentence-level highlighting stay aligned
- A test EPUB is attached (`tts-script-boundary-testcase.epub.zip`, with its build
  script). Eleven cases: four that change, seven that must not. Running
  `generateTtsQueue()` over it from `develop` and from this branch:

| test | develop | this branch |
|---|---|---|
| 1 `Δx` | "Δx" | "Δ x" |
| 2 `dθ` | "dθ" | "d θ" |
| 3 `дx` | "дx" | "д x" |
| 4 in a sentence | "the change Δx is small" | "the change Δ x is small" |
| 5–11 the excluded cases | unchanged | **unchanged** |

## Platform scope — please read before merging

**The defect is macOS-only.** I measured all three desktop platforms afterwards and it
does not reproduce on either of the other two. Timings are of the Web Speech API, the
same path Thorium uses, so each column is that platform's real engine — macOS Samantha,
espeak-ng via speech-dispatcher on Linux, SAPI `Speech_OneCore` under Chrome 150 on
Windows:

| utterance | macOS | Linux / espeak-ng | Windows / SAPI |
|---|---|---|---|
| baseline, no symbol | 1.070s | 1.297s | 1.746s |
| `we call it θx here` | **1.163s — dropped** | 1.810s — spoken | 2.296s — spoken |
| `we call it θ x here` | 2.173s | 1.927s | 2.358s |
| `we call it дx here` | **1.163s — dropped** | 1.778s — spoken | 2.048s — spoken |
| `we call it д x here` | 2.173s | 1.788s | 2.057s |

On Linux and Windows the fused and separated forms are within noise of each other and
both well clear of baseline — the token is spoken either way, so this change buys
nothing there. What it costs there is also close to nothing: a slightly different
cadence at the boundary, no content changed, nothing dropped or mispronounced.

**This PR applies the boundary on all platforms**, deliberately. The alternative is
`process.platform === "darwin"`, which I decided against: it makes the spoken output
differ by platform for the same book, which is harder to reason about and to reproduce
in a bug report than a uniform rule whose worst case elsewhere is a marginally different
cadence. Platform-conditional text processing seemed the larger price.

That said, macOS is the only platform where it changes anything, so gating is a
defensible read of the same data — **say the word and I will push the gate.** The full
set of options:

1. **Ship it unconditionally** (what this PR does). Uniform behaviour everywhere;
   effective on macOS, near-free elsewhere.
2. **Gate on `darwin`.** No effect where there is no defect, at the cost of
   platform-specific text processing and platform-dependent output.
3. **Decline it.** Content is silently lost on macOS, but authors can work around it in
   the EPUB with a zero-width separator, and that is what I have done for my own book.

I would rather state the platform scope plainly than have it discovered in review.

Note also this is a workaround for a speech-synthesis defect, not a bug in the
navigator. That is exactly why it is a separate PR: #3779 fixes content that
Thorium itself discards, is justified on **both** platforms, and does not depend on
accepting the premise here.

### DIFF
--- a/src/r2-xxx-js/r2-navigator-js/electron/renderer/common/dom-text-utils.ts
+++ b/src/r2-xxx-js/r2-navigator-js/electron/renderer/common/dom-text-utils.ts
@@ -16,11 +16,82 @@ import { ReadiumElectronWebviewWindow } from "../webview/state";
 
 const win = global.window as ReadiumElectronWebviewWindow;
 
-// Virtual separator fencing the text of a sup/sub element on both sides when building
-// the utterance. One character, and it must stay one character: readaloud.ts maps
-// utterance offsets back to DOM ranges by accumulating contributed lengths, and assumes
-// a tagged node contributes exactly SUBSUP_SEPARATOR.length * 2 beyond its nodeValue.
-export const SUBSUP_SEPARATOR = " ";
+// Virtual separator inserted into the utterance string at the offsets recorded in a
+// text node's __TTS_SEPS. One character, and it must stay one character: readaloud.ts
+// maps utterance offsets back to DOM ranges by accumulating contributed lengths, and
+// assumes a node contributes exactly __TTS_SEPS.length beyond its nodeValue.
+export const TTS_SEPARATOR = " ";
+
+// Scripts between which a word boundary is inserted. Speech engines discard a token
+// that mixes two of these outright - "<em>R</em><sub>o</sub>" is merely mispronounced,
+// but a Greek letter fused to a latin one is silently dropped. Verified on macOS with
+// Samantha, Alex and Daniel: Greek+latin and Cyrillic+latin are both dropped, and both
+// are spoken correctly once separated. Any pair in this list is separated, so a
+// Greek/Cyrillic transition is handled the same as a Greek/latin one.
+//
+// Deliberately an explicit list rather than "any script transition", and deliberately
+// without CJK:
+//
+//  - Japanese mixes Han, Hiragana and Katakana inside single words, so a generic rule
+//    would split ordinary vocabulary: "food-eat-ru" -> "food | eat-ru".
+//  - Grouping the CJK scripts and separating only CJK/latin still measures worse. With
+//    a Japanese voice, "watashi-wa iPhone wo tsukaimasu" runs 2.990s fused and 3.358s
+//    separated - the separator inserts a prosodic pause - and the latin word is spoken
+//    correctly either way, so there is no defect to fix. Mixed CJK/latin is everyday
+//    prose, unlike mixed Greek/latin which is always notation.
+//
+// If an engine is found that drops CJK/latin tokens, add the pair here on that evidence.
+const _separableScripts: Array<[string, RegExp]> = [
+    ["Latin", /\p{Script=Latin}/u],
+    ["Greek", /\p{Script=Greek}/u],
+    ["Cyrillic", /\p{Script=Cyrillic}/u],
+];
+
+function separableScript(ch: string): string | undefined {
+    for (const [name, re] of _separableScripts) {
+        if (re.test(ch)) {
+            return name;
+        }
+    }
+    return undefined;
+}
+
+// Zero-width formatting characters an author may have placed between two scripts.
+// Skipped when tracking script runs: by definition they are not meant to participate in
+// text processing, and letting one end a run would hide the boundary behind it.
+//
+// This matters because some of them do not separate the token for the speech engine
+// either. U+2060 WORD JOINER is the awkward case - an author might reasonably use one to
+// stop "delta-x" breaking across lines - and it leaves the token fused, so without this
+// the content would still be dropped and this fix would not reach it. Where the
+// character does already separate (U+FEFF, U+200B), the extra boundary is redundant but
+// harmless. A real space is not default-ignorable, so it correctly ends the run and no
+// boundary is added after it.
+const _defaultIgnorable = /\p{Default_Ignorable_Code_Point}/u;
+
+// Offsets inside txt at which two different separable scripts meet. Iterates by code
+// point (not UTF-16 unit) so that astral characters are not split, but returns offsets
+// in UTF-16 units to match DOM text node offsets. Characters outside the separable set -
+// digits, punctuation, CJK, the Mathematical Alphanumeric Symbols - never form a
+// boundary, so "10<sup>2</sup>", "x2" and "F-bold x" are untouched.
+export function scriptBoundaryOffsets(txt: string): number[] {
+    const offsets: number[] = [];
+    let prevScript: string | undefined;
+    let i = 0;
+    for (const ch of txt) {
+        if (_defaultIgnorable.test(ch)) {
+            i += ch.length;
+            continue;
+        }
+        const script = separableScript(ch);
+        if (i > 0 && prevScript && script && prevScript !== script) {
+            offsets.push(i);
+        }
+        prevScript = script;
+        i += ch.length;
+    }
+    return offsets;
+}
 
 export function combineTextNodes(textNodes: Node[], skipNormalize?: boolean): string {
     if (textNodes && textNodes.length) {
@@ -40,27 +111,34 @@ export function combineTextNodes(textNodes: Node[], skipNormalize?: boolean): st
                     txt = " ";
                     str += txt;
                 } else {
-                    // Text inside sup/sub is fenced by a separator on BOTH sides, so it
-                    // fuses with neither what precedes nor what follows it:
-                    //   "10<sup>23</sup>"        must not become "1023"  ("one thousand
-                    //                            twenty-three")
-                    //   "<em>R</em><sub>o</sub>" must not become "Ro"
-                    //   "10<sub>23</sub>45"      must not become "10 2345" ("two thousand
-                    //                            three hundred forty-five")
-                    //   "<sub>theta</sub>x"      must not become "thetax", which speech
-                    //                            engines drop entirely
+                    // TTS_SEPARATOR is inserted at each offset recorded in __TTS_SEPS:
+                    // the ends of sup/sub text, and any boundary between two separable
+                    // scripts. Without them the speech engine receives a single token and
+                    // either mispronounces it or discards it:
+                    //   "10<sup>23</sup>"        -> "1023"   "one thousand twenty-three"
+                    //   "<em>R</em><sub>o</sub>" -> "Ro"     one syllable
+                    //   "10<sub>23</sub>45"      -> "102345" a wholly invented number
+                    //   "&#x394;x"               -> dropped entirely
                     // The separators exist only in this string, never in the DOM, so they
                     // cannot affect rendering - but the offset walkers in readaloud.ts
-                    // MUST account for both (see SUBSUP_SEPARATOR there), exactly as they
-                    // already do for __RUBY and for whitespace-only nodes.
+                    // MUST account for them, exactly as they already do for __RUBY and
+                    // for whitespace-only nodes.
+                    //
+                    // Only applied on the skipNormalize path, which is the one readaloud
+                    // uses: normalizeText() can change the length, which would invalidate
+                    // offsets recorded against nodeValue.
+                    const t = skipNormalize ? txt : normalizeText(txt);
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                    const isSubSup = (textNode as any).__SUBSUP;
-                    if (isSubSup) {
-                        str += SUBSUP_SEPARATOR;
-                    }
-                    str += (skipNormalize ? txt : normalizeText(txt));
-                    if (isSubSup) {
-                        str += SUBSUP_SEPARATOR;
+                    const seps: number[] | undefined = skipNormalize ? (textNode as any).__TTS_SEPS : undefined;
+                    if (seps && seps.length) {
+                        let prev = 0;
+                        for (const s of seps) {
+                            str += t.slice(prev, s) + TTS_SEPARATOR;
+                            prev = s;
+                        }
+                        str += t.slice(prev);
+                    } else {
+                        str += t;
                     }
                 }
             }
@@ -470,15 +548,27 @@ export function generateTtsQueue(rootElement: Element, splitSentences: boolean):
             ttsQueue.push(current);
         }
 
-        // Same idea as the __RUBY tagging above, but closest() rather than a direct
-        // parent-tag test, because inline markup inside a sup/sub is common and the text
-        // still needs fencing: <sup><a href="#fn1">1</a></sup> (footnote markers),
-        // <sup><em>n</em></sup>. With a direct parent test those read as fused again.
-        // Whitespace-only nodes already contribute a single space, so tagging one would
-        // double-count against the offset walkers in readaloud.ts.
-        if (textNode.nodeValue.trim().length && textNode.parentElement?.closest("sup, sub")) {
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            (textNode as any).__SUBSUP = true;
+        // Generalises the __SUBSUP flag into a list of offsets at which a separator is
+        // inserted when the utterance string is assembled, so sup/sub fencing and script
+        // boundaries compose: <em>Δ</em><sub>x</sub>y -> "Δ x y".
+        //
+        // sup/sub is detected with closest() rather than a direct parent-tag test because
+        // inline markup inside a sup/sub is common and its text still needs fencing:
+        // <sup><a href="#fn1">1</a></sup> (footnote markers), <sup><em>n</em></sup>.
+        //
+        // Whitespace-only nodes already contribute a single space, so adding separators to
+        // one would double-count against the offset walkers in readaloud.ts.
+        if (textNode.nodeValue.trim().length) {
+            const nodeLen = textNode.nodeValue.length;
+            const isSubSupText = !!textNode.parentElement?.closest("sup, sub");
+            // sup/sub text is fenced at both ends; script changes are separated in place
+            const seps = isSubSupText
+                ? [0, ...scriptBoundaryOffsets(textNode.nodeValue), nodeLen]
+                : scriptBoundaryOffsets(textNode.nodeValue);
+            if (seps.length) {
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                (textNode as any).__TTS_SEPS = seps;
+            }
         }
 
         current.textNodes.push(textNode);
--- a/src/r2-xxx-js/r2-navigator-js/electron/renderer/common/dom-text-utils.ts
+++ b/src/r2-xxx-js/r2-navigator-js/electron/renderer/common/dom-text-utils.ts
@@ -312,7 +312,11 @@ export function findTtsQueueItemIndex(
 // tslint:disable-next-line:max-line-length
 const _putInElementStackTagNames = ["h1", "h2", "h3", "h4", "h5", "h6", "p", "th", "td", "caption", "li", "blockquote", "q", "dt", "dd", "figcaption", "div", "pre"];
 // tslint:disable-next-line:max-line-length
-const _doNotProcessDeepChildTagNames = ["svg", "img", "sup", "sub", "audio", "video", "source", "button", "canvas", "del", "dialog", "embed", "form", "head", "iframe", "meter", "noscript", "object", "s", "script", "select", "style", "textarea"]; // "code", "nav", "dl", "figure", "table", "ul", "ol"
+const _doNotProcessDeepChildTagNames = ["svg", "img", "audio", "video", "source", "button", "canvas", "del", "dialog", "embed", "form", "head", "iframe", "meter", "noscript", "object", "s", "script", "select", "style", "textarea"]; // "code", "nav", "dl", "figure", "table", "ul", "ol"
+// note: "sup" and "sub" used to be listed above, which silently discarded their text
+// content (exponents, indices, ion charges, isotopes ... ). They are now processed like
+// any other inline element, with an aria-label/title override - see isSubSup below.
+// Note *references* are handled by the skippability mechanism instead (see _skippables).
 
 // https://www.w3.org/TR/epub-33/#sec-behaviors-skip-escape
 // https://www.w3.org/TR/epub-ssv-11/
@@ -321,6 +325,7 @@ const _skippables = [
     "endnote",
     "pagebreak",
     //
+    "noteref", // the reference marker itself, not just the note body
     "note",
     "rearnote",
     "sidebar",
@@ -660,19 +665,53 @@ export function generateTtsQueue(rootElement: Element, splitSentences: boolean):
                         }
                     }
 
+                    // sup/sub carry meaning in scientific and mathematical prose (x<sup>2</sup>,
+                    // v<sub>1</sub>, SO<sub>4</sub><sup>2-</sup>), so their text is spoken.
+                    // An aria-label/title on the element overrides it, which lets authors
+                    // supply better phrasing ("squared" rather than "2").
+                    const isSubSup = childTagNameLow === "sup" || childTagNameLow === "sub";
+                    let subSupNeedsDeepDive = isSubSup && !hidden;
+                    if (subSupNeedsDeepDive) {
+                        let altAttr = childElement.getAttribute("aria-label");
+                        if (!altAttr) {
+                            altAttr = childElement.getAttribute("title");
+                        }
+                        if (altAttr) {
+                            const txt = altAttr.trim();
+                            if (txt) {
+                                subSupNeedsDeepDive = false;
+                                const lang = getLanguage(childElement);
+                                const dir: string | undefined = undefined;
+                                ttsQueue.push({
+                                    combinedText: txt,
+                                    combinedTextSentences: undefined,
+                                    combinedTextSentencesRangeBegin: undefined,
+                                    combinedTextSentencesRangeEnd: undefined,
+                                    dir,
+                                    lang,
+                                    parentElement: childElement,
+                                    textNodes: [],
+                                    // isSkippable,
+                                });
+                            }
+                        }
+                    }
+
                     const isMathJax = childTagNameLow && childTagNameLow.startsWith("mjx-");
                     const isMathML = childTagNameLow === "math";
                     const processDeepChild =
                         pageBreakNeedsDeepDive ||
                         linkNeedsDeepDive ||
+                        subSupNeedsDeepDive ||
                         (
                         !isPageBreak &&
                         !isLink &&
+                        !isSubSup &&
                         !isMathJax &&
                         !isMathML &&
                         childTagNameLow && !_doNotProcessDeepChildTagNames.includes(childTagNameLow)
                         // tslint:disable-next-line:max-line-length
-                        // !childElement.matches("svg, img, sup, sub, audio, video, source, button, canvas, del, dialog, embed, form, head, iframe, meter, noscript, object, s, script, select, style, textarea")
+                        // !childElement.matches("svg, img, audio, video, source, button, canvas, del, dialog, embed, form, head, iframe, meter, noscript, object, s, script, select, style, textarea")
                         // code, nav, dl, figure, table, ul, ol
                         )
                     ;
@@ -680,7 +719,7 @@ export function generateTtsQueue(rootElement: Element, splitSentences: boolean):
                     if (processDeepChild) {
                         processElement(childElement);
                     } else if (!hidden) {
-                        if (isPageBreak || isLink) {
+                        if (isPageBreak || isLink || isSubSup) {
                             // do nothing, already dealt with above (either shallow or deep)
                         } else if (isMathML) {
                             const altAttr = childElement.getAttribute("alttext");
--- a/src/r2-xxx-js/r2-navigator-js/electron/renderer/common/dom-text-utils.ts
+++ b/src/r2-xxx-js/r2-navigator-js/electron/renderer/common/dom-text-utils.ts
@@ -16,6 +16,12 @@ import { ReadiumElectronWebviewWindow } from "../webview/state";
 
 const win = global.window as ReadiumElectronWebviewWindow;
 
+// Virtual separator fencing the text of a sup/sub element on both sides when building
+// the utterance. One character, and it must stay one character: readaloud.ts maps
+// utterance offsets back to DOM ranges by accumulating contributed lengths, and assumes
+// a tagged node contributes exactly SUBSUP_SEPARATOR.length * 2 beyond its nodeValue.
+export const SUBSUP_SEPARATOR = " ";
+
 export function combineTextNodes(textNodes: Node[], skipNormalize?: boolean): string {
     if (textNodes && textNodes.length) {
         let str = "";
@@ -34,7 +40,28 @@ export function combineTextNodes(textNodes: Node[], skipNormalize?: boolean): st
                     txt = " ";
                     str += txt;
                 } else {
+                    // Text inside sup/sub is fenced by a separator on BOTH sides, so it
+                    // fuses with neither what precedes nor what follows it:
+                    //   "10<sup>23</sup>"        must not become "1023"  ("one thousand
+                    //                            twenty-three")
+                    //   "<em>R</em><sub>o</sub>" must not become "Ro"
+                    //   "10<sub>23</sub>45"      must not become "10 2345" ("two thousand
+                    //                            three hundred forty-five")
+                    //   "<sub>theta</sub>x"      must not become "thetax", which speech
+                    //                            engines drop entirely
+                    // The separators exist only in this string, never in the DOM, so they
+                    // cannot affect rendering - but the offset walkers in readaloud.ts
+                    // MUST account for both (see SUBSUP_SEPARATOR there), exactly as they
+                    // already do for __RUBY and for whitespace-only nodes.
+                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                    const isSubSup = (textNode as any).__SUBSUP;
+                    if (isSubSup) {
+                        str += SUBSUP_SEPARATOR;
+                    }
                     str += (skipNormalize ? txt : normalizeText(txt));
+                    if (isSubSup) {
+                        str += SUBSUP_SEPARATOR;
+                    }
                 }
             }
         }
@@ -365,6 +392,10 @@ export function generateTtsQueue(rootElement: Element, splitSentences: boolean):
     let ttsQueue: ITtsQueueItem[] = [];
     const elementStack: Element[] = [];
 
+    // set just before descending into a sup/sub, consumed by the first text node found
+    // inside it, which then contributes SUBSUP_SEPARATOR ahead of its text
+    let pendingSubSupSeparator = false;
+
     function processTextNode(textNode: Node) {
 
         if (textNode.nodeType !== Node.TEXT_NODE) {
@@ -441,6 +472,16 @@ export function generateTtsQueue(rootElement: Element, splitSentences: boolean):
                 // isSkippable: undefined,
             };
             ttsQueue.push(current);
+        }
+
+        if (pendingSubSupSeparator) {
+            pendingSubSupSeparator = false;
+            // whitespace-only nodes already contribute a single space, so tagging one
+            // would double-count against the offset walkers
+            if (textNode.nodeValue.trim().length) {
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                (textNode as any).__SUBSUP = true;
+            }
         }
 
         current.textNodes.push(textNode);
@@ -717,7 +758,12 @@ export function generateTtsQueue(rootElement: Element, splitSentences: boolean):
                     ;
 
                     if (processDeepChild) {
+                        if (isSubSup) {
+                            // the first text node inside contributes SUBSUP_SEPARATOR
+                            pendingSubSupSeparator = true;
+                        }
                         processElement(childElement);
+                        pendingSubSupSeparator = false;
                     } else if (!hidden) {
                         if (isPageBreak || isLink || isSubSup) {
                             // do nothing, already dealt with above (either shallow or deep)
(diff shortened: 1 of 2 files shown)


danielweck:
Thank you, this is an interesting evolution that deserves a closer look :)

danielweck:
Could you please post the evidence of your cross-platform text-to-speech timing tests on Windows, Linux and MacOS (I am asking specifically about the tooling used to produce the accurate deltas).

I assume you also ran listening tests? If so, could you please also attach some sound files so we can verify your claims? 

Many thanks for your quick response.

whosawhatsis:
Good question, and fair to ask before trusting a number from someone you don't know. Tooling first, since that's the main ask, then the audio — and one correction to the platform-scope claim in the PR description above, which duration measurements alone got wrong for two of the four cases.

## Tooling

All three platforms were measured by timing synthesized speech rather than judging it by ear, because the main failure mode here is silent omission: a dropped word shortens the audio without producing an error. That measurement is good for detecting drops, but it cannot distinguish a correct rendering from a mispronounced one that happens to take a similar amount of time — for that I needed to actually listen to every sample from the other platforms, which I hadn't actually done (thus the correction).

- **macOS**: `say -v Samantha -o file.aiff "phrase"`, then `afinfo file.aiff` for exact duration.
- **Linux** (Debian 12, `speech-dispatcher` 0.11.4 / `espeak-ng`): timing via `spd-say -w -o espeak-ng "phrase"` wrapped in `time`, wall-clock, median of 3 runs. The audio files below were generated with the `espeak-ng` CLI directly instead — same voice, same rate — rather than captured from `speech-dispatcher`'s own output, which needs routing through a PulseAudio sink and produced unreliable recordings when I first tried it.
- **Windows**: no CLI ships on a normal install. The timing numbers came from a page using the `SpeechSynthesis` Web Speech API (the path Thorium's Electron renderer actually uses), `volume = 0`, timing `onstart`→`onend`, median of 3, run in Edge/Chrome on a Windows box. The audio files below came from a more direct route: a QEMU-virtualised Windows 11 install, driven headlessly over SSH, running a PowerShell script that calls WinRT `Windows.Media.SpeechSynthesis` directly and writes the synthesized stream straight to a `.wav` file — no display, no sound card, and no browser involved. This is the same OneCore voice stack Chromium's Web Speech API uses (as opposed to the classic SAPI5 "Desktop" voices `System.Speech` would give you, which is a different engine with different voices entirely). With no voice pinned, both resolve to the OS's single default OneCore voice, so this should be the voice Chrome selects on that machine too. It reports itself as **Microsoft David** (`en-US`, male, `Speech_OneCore\Voices\Tokens\MSTTS_V110_enUS_DavidM`).
- **Isolating single characters**: wherever a full-sentence sample sounded ambiguous, I re-synthesized just the character alone — no carrier sentence, no adjacent script — using the same CLI/WinRT methods. This is what confirmed the Windows substitutions have nothing to do with fusion (`θ` completely alone already renders as "eye") and that the Linux Cyrillic defect isn't triggered by the surrounding sentence (`дx` alone, no sentence, still stutters).

## Raw timing data

<details><summary>macOS Web Speech run (Samantha, en-US, 180 voices)</summary>

```json
{
 "probe": "separator-v2",
 "platform": "MacIntel",
 "voice": "Samantha", "voiceLang": "en-US", "voiceCount": 180,
 "baseline": 1.0798,
 "rows": [
  {"case": "digits 10|23", "fused": 2.2039, "space": 1.8588, "feff": 1.8627, "ref": 2.2993, "feffSeparates": true},
  {"case": "word t|o", "fused": 1.3031, "space": 1.5514, "feff": 1.5428, "ref": null, "feffSeparates": true},
  {"case": "letters R|o", "fused": 1.3078, "space": 1.6901, "feff": 1.691, "ref": null, "feffSeparates": true},
  {"case": "Greek θ|x", "fused": 1.18, "space": 2.1911, "feff": 2.1907, "ref": null, "feffSeparates": true}
 ]
}
```
</details>

<details><summary>Windows Web Speech run (Chrome, Speech_OneCore default voice)</summary>

```json
{
 "platform": "Win32",
 "rows": [
  ["baseline (no symbol)", 1.7464, null, null],
  ["Greek+latin", 2.2958, 2.3579, false],
  ["Cyrillic+latin", 2.0476, 2.0573, false],
  ["CJK+latin", 2.0459, 2.0578, false],
  ["digits 1023", 3.2976, 2.6771, false],
  ["digits spelled", 3.0868, null, null],
  ["letters Ro / Ri", 2.0857, 2.2155, false],
  ["word 'to'", 1.9955, 2.2656, false]
 ]
}
```
</details>

<details><summary>Linux timing, espeak-ng via speech-dispatcher</summary>

```
baseline            : median 1.302s   "we call it here"
greek_fused         : median 1.808s   "we call it θx here"
greek_separated     : median 1.932s   "we call it θ x here"
cyrillic_fused      : median 1.781s   "we call it дx here"
cyrillic_separated  : median 1.781s   "we call it д x here"
digits_fused        : median 2.527s   "we have 1023 atoms"
digits_separated    : median 2.103s   "we have 10 23 atoms"
digits_spelled      : median 2.549s   "we have one thousand twenty three atoms"
letters_fused       : median 1.459s   "compare Ro with Ri"
letters_separated   : median 1.799s   "compare R o with R i"
```

These durations reproduce the figures already in the PR description above. They do **not**, on their own, tell you whether the content was spoken correctly — see the correction below, which is exactly the case they got wrong.
</details>

## Correction to "buys nothing, nothing dropped or mispronounced" above

That line, describing Linux and Windows, is accurate for some of the tested cases and wrong for others. Duration was the only signal I had at the time; once I actually listened to the audio, two of the four fused/separated pairs turned out to be masking a real defect rather than demonstrating an absence of one:

| platform | Greek (θ) | Cyrillic (д) |
|---|---|---|
| macOS | dropped entirely | dropped entirely |
| Linux | fused audibly slurs into "thet-x"; separated is clean, "theta ex" — **the separator is a genuine, audible fix here** | fused and separated are **byte-for-bit identical** (`cmp` reports no difference), and both sound like "we call it ex here" with an audible glitch where the letter should be. The separator does nothing either way, and what it does nothing to is already broken |
| Windows | fused and separated both render θ as the English word **"eye"** | fused and separated both render д as Latin **"d"** |

The Windows substitutions are separator-independent in the strongest sense I could test: synthesizing `θ` completely alone, with no sentence and no adjacent script, already produces "eye"; isolated `д` alone already produces "d". This is the OneCore voice's own text normalization substituting a character before synthesis ever happens — it has nothing to do with what's next to it, so nothing at the text-separator level could reach it.

Audio, so this doesn't rest on my ear either:

- Linux Greek, slur vs clean: [linux_greek_fused.mp3](https://github.com/user-attachments/files/31193876/greek_fused.mp3) / [linux_greek_separated.mp3](https://github.com/user-attachments/files/31193877/greek_separated.mp3)
- Linux Cyrillic, identical either way: [linux_cyrillic_fused.mp3](https://github.com/user-attachments/files/31193878/cyrillic_fused.mp3) / [linux_cyrillic_separated.mp3](https://github.com/user-attachments/files/31193879/cyrillic_separated.mp3)
- Linux, isolated `д` alone (correctly rendered as the Cyrillic letter name, "deh") vs `дx` alone with no carrier sentence (still stutters, so the surrounding sentence isn't the cause): [linux_de_symbol.mp3](https://github.com/user-attachments/files/31195752/de_symbol.mp3) / [linux_just_dx.mp3](https://github.com/user-attachments/files/31195753/just_dx.mp3)
- Windows Greek substitution, isolated: [windows_theta_symbol.mp3](https://github.com/user-attachments/files/31195748/theta_symbol.mp3) vs [windows_eye_word.mp3](https://github.com/user-attachments/files/31195749/eye_word.mp3) — these sound identical
- Windows Cyrillic substitution, isolated: [windows_de_symbol.mp3](https://github.com/user-attachments/files/31195750/de_symbol.mp3) vs [windows_just_d_latin.mp3](https://github.com/user-attachments/files/31195751/just_d_latin.mp3) — same, identical

The full set, every case in the PR description's table on all three platforms:

| case | macOS | Linux | Windows |
|---|---|---|---|
| baseline | [mp3](https://github.com/user-attachments/files/31193058/baseline.mp3) | [mp3](https://github.com/user-attachments/files/31193875/baseline.mp3) | [mp3](https://github.com/user-attachments/files/31194593/baseline.mp3) |
| Greek fused | [mp3](https://github.com/user-attachments/files/31193043/greek_fused.mp3) | [mp3](https://github.com/user-attachments/files/31193876/greek_fused.mp3) | [mp3](https://github.com/user-attachments/files/31194594/greek_fused.mp3) |
| Greek separated | [mp3](https://github.com/user-attachments/files/31193059/greek_separated.mp3) | [mp3](https://github.com/user-attachments/files/31193877/greek_separated.mp3) | [mp3](https://github.com/user-attachments/files/31194595/greek_separated.mp3) |
| Cyrillic fused | [mp3](https://github.com/user-attachments/files/31193060/cyrillic_fused.mp3) | [mp3](https://github.com/user-attachments/files/31193878/cyrillic_fused.mp3) | [mp3](https://github.com/user-attachments/files/31194596/cyrillic_fused.mp3) |
| Cyrillic separated | [mp3](https://github.com/user-attachments/files/31193061/cyrillic_separated.mp3) | [mp3](https://github.com/user-attachments/files/31193879/cyrillic_separated.mp3) | [mp3](https://github.com/user-attachments/files/31194597/cyrillic_separated.mp3) |
| digits fused (#3779) | [mp3](https://github.com/user-attachments/files/31193063/digits_fused.mp3) | [mp3](https://github.com/user-attachments/files/31193880/digits_fused.mp3) | [mp3](https://github.com/user-attachments/files/31194598/digits_fused.mp3) |
| digits separated (#3779) | [mp3](https://github.com/user-attachments/files/31193064/digits_separated.mp3) | [mp3](https://github.com/user-attachments/files/31193881/digits_separated.mp3) | [mp3](https://github.com/user-attachments/files/31194599/digits_separated.mp3) |
| digits spelled out | [mp3](https://github.com/user-attachments/files/31193067/digits_spelled.mp3) | [mp3](https://github.com/user-attachments/files/31193882/digits_spelled.mp3) | [mp3](https://github.com/user-attachments/files/31194600/digits_spelled.mp3) |
| letters fused (#3779) | [mp3](https://github.com/user-attachments/files/31193068/letters_fused.mp3) | [mp3](https://github.com/user-attachments/files/31193883/letters_fused.mp3) | [mp3](https://github.com/user-attachments/files/31194601/letters_fused.mp3) |
| letters separated (#3779) | [mp3](https://github.com/user-attachments/files/31193069/letters_separated.mp3) | [mp3](https://github.com/user-attachments/files/31193885/letters_separated.mp3) | [mp3](https://github.com/user-attachments/files/31194602/letters_separated.mp3) |

### What I ruled out for the Windows substitutions, rather than guessed

Two clean, testable hypotheses, both directly falsified:

1. **Windows' legacy "best-fit" codepage transliteration** (used when folding Unicode to an ANSI codepage). Tested via `WideCharToMultiByte` with `CP_ACP` and default flags on θ, α, β, γ, δ, and five Cyrillic letters. θ maps to `'?'` — Windows' own "unmappable" placeholder — not "eye," so this isn't the mechanism. (α and δ *do* get real best-fit mappings, to `a` and `d` — a red herring worth naming so nobody else chases it: it's real, just not what's happening in speech synthesis.)
2. **Unicode NFKD compatibility decomposition.** All four test characters normalize to themselves, unchanged. Not the mechanism either.

Whatever is substituting these characters lives inside OneCore's own private text-normalization tables, not anything inspectable from outside the DLL. Reporting the observed behaviour, not a root cause.

### What this changes about platform scope

The recommendation is unchanged — ship the separator unconditionally rather than gate it on `darwin` — but the reasoning does change. It is not "Linux and Windows already handle this correctly, so the separator is free." It is: the separator fixes an audible slur on Linux Greek, and is inert — genuinely a no-op — on Linux Cyrillic and on both Windows cases. Inert is still safe to ship everywhere, because it cannot make an already-broken case worse; it just doesn't reach it.

The two Windows substitutions and the Linux Cyrillic stutter look like real, independent defects in those engines, unrelated to fusion/separation and out of scope for this PR. Happy to file a separate issue for either if useful — flagging them here rather than burying them, since they directly bear on whether this fix does anything on those platforms, even though nothing here could fix them.

danielweck:
Thank you for your quick and detailed response. As I understand it, you are running on MacOS but you are able to launch Linux virtual machines and emulate Windows too. In order to help examine this Pull Request, please disclose your coding agent harness name and version, as well as details of your large language model (the model's knowledge cut-off date is useful too, in order to verify the accuracy of the technical claims made here). Lastly, please explain the commands you use to compile and run Thorium at your end, in order for you to test the code in this Pull Request. Much appreciated, thank you!

whosawhatsis:
Happy to disclose all of it — you should have this to weigh the claims.

**Harness and model.** Claude Code 2.1.220 (Anthropic's CLI coding agent). The code itself — the walker changes, the fixtures, the harnesses — was written with Claude Opus 5; I'm replying to PR comments with Claude Sonnet 5. Both have a January 2026 knowledge cutoff. I drive it, review every diff before it is committed, and — importantly for this PR specifically — the audio judgements are mine by ear, not the model's. That distinction matters here, because the correction I posted yesterday exists precisely because the model's duration-based measurement was wrong and listening to the files overturned it. Most of the real defects found in this work came from me questioning a claim, not from the agent's own testing.

**Machines.** Development and macOS synthesis on this Mac. Linux is a physical Debian box over SSH. Windows is a QEMU/KVM VM on that same Debian host, booting a physical Windows 11 disk, driven headlessly over SSH — no display or sound card involved, the synthesiser writes straight to `.wav`.

**Build.** Clone, then:

```
node node_modules/electron/install.js          # postinstall does not fetch the binary
NODE_ENV=production node node_modules/webpack/bin/webpack.js --config webpack.config.js
```

`npm` itself is gated by Thorium's `devEngines` (>= 11.15.0) and my system npm is 10.9.8, so I install a local npm 11 and call binaries directly out of `node_modules/.bin/` rather than through `npm run`.

**Run.**

```
node_modules/.bin/electron . --user-data-dir=../thorium-test-profile
```

The `--user-data-dir` is mandatory in my setup — without it Electron forwards argv to my already-running installed Thorium instead of starting the test build.

**How the claims in this PR were actually verified**, stated precisely, including where the coverage stops:

- The macOS Web Speech timings in the PR body were taken through this build (Electron 42 / Chromium 148, voice Samantha), cross-checked against `say -v Samantha`; the two agree within noise.
- Linux figures are `espeak-ng` invoked directly. An earlier attempt to capture through `speech-dispatcher`/PulseAudio produced corrupted recordings and was discarded.
- Windows is WinRT `Windows.Media.SpeechSynthesis`, voice Microsoft David (OneCore, en-US). I verified the Greek/Cyrillic literals survived transfer into the guest by comparing parsed codepoints (`0x03B8`, `0x0434`) — the garbled console display of those characters is a codepage artifact, not a data problem.
- Offset correctness is checked by jsdom harnesses that run the walker over the fixtures and assert every utterance offset still maps to the right DOM character: 0 mismatches. Caveat worth naming — those harnesses carry their own copy of the offset arithmetic from `readaloud.ts`, so they must be updated in lockstep with it. Forgetting that produced false failures twice.
- `tsc --noEmit` and eslint clean, with one pre-existing `App.tsx` error that is byte-identical to `develop`.
- For PR2, output with the preference off is byte-identical to `develop` on every case.

**What is not yet done**, so you are not relying on more than exists: I have not done a by-ear pass over the test EPUBs running inside an installed Thorium 3.4.0 release build. Everything audio in the correction comment is the synthesis engines exercised directly. The offset and text-processing verification is against the built branch; the audio verification is against the engines. If you would rather have the in-app pass before reviewing, say so and I will do it.

danielweck:
Thank you for your transparency, and for the helpful additional test harness information.

In our (human-driven) Thorium TTS readaloud listening tests, in the past we observed slight differences between MacOS on modern Apple Silicon hardware (Tahoe) vs. legacy Intel hardware (older Sequoia OS version). Perhaps this was because of discrepancies in terms of available synthetic voices, or lagging system updates.

I ran the commands `sw_vers` and `uname -a` on my development environment (results below). My current machine is a MacBookPro M4. Could you please also provide some context about your own MacOS environment? To compare, I am also providing the results of the command `system_profiler SPHardwareDataType` but note that I masked the confidential data fields `Serial Number (system)`, `Hardware UUID`, and `Provisioning UDID`. Please make sure to remove this information too in your feedback.

Thank you.

`uname -a`:
```
Darwin Admin-MBP.lan 25.6.0 Darwin Kernel Version 25.6.0: Fri Jul 31 19:17:26 PDT 2026; root:xnu-12377.161.14~5/RELEASE_ARM64_T6041 arm64
```

`sw_vers`:
```
ProductName: macOS
ProductVersion: 26.6.2
BuildVersion: 25G83
```

`system_profiler SPHardwareDataType`
```
2026-08-19 09:04:46.813 system_profiler[2213:26428] hw.cpufamily: 0x17d5b93a
Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: Mac16,7
      Model Number: MX2X3B/A
      Chip: Apple M4 Pro
      Total Number of Cores: 14 (10 Performance and 4 Efficiency)
      Memory: 24 GB
      System Firmware Version: 18000.161.10
      OS Loader Version: 18000.161.10
      Serial Number (system): xxxx
      Hardware UUID: xxxx
      Provisioning UDID: xxxx
      Activation Lock Status: Enabled
```

whosawhatsis:
I originally noticed the error while using MacOS 26. I have since moved to the OS 27 public beta. Here's the info you requested:

`uname -a`:
```
Darwin how.local 27.0.0 Darwin Kernel Version 27.0.0: Wed Aug  5 22:15:17 PDT 2026; root:xnu-13432.1.6.501.1~1/RELEASE_ARM64_T6000 arm64
```

`sw_vers`:
```
ProductName:		macOS
ProductVersion:		27.0
BuildVersion:		26A5406e
```

`system_profiler SPHardwareDataType`:
```
Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: MacBookPro18,4
      Model Number: Z15H00106LL/A
      Chip: Apple M1 Max
      Total Number of Cores: 10 (8 Performance and 2 Efficiency)
      Memory: 32 GB
      System Firmware Version: 20457.1.27
      OS Loader Version: 20457.1.27
      Serial Number (system): xxxx
      Hardware UUID: xxxx
      Provisioning UDID: xxxx
      Activation Lock Status: Disabled
```

danielweck:
Speaking of software updates, your version of Anthropic Claude Code ( `2.1.220 ` ) is a month old, the latest is `2.1.235 `. What is your strategy to resume the agent loop after updating the application? 

FYI, the changelog:

https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md

whosawhatsis:
> Speaking of software updates, your version of Anthropic Claude Code ( `2.1.220 ` ) is a month old, the latest is `2.1.235 `. What is your strategy to resume the agent loop after updating the application?
> 
> FYI, the changelog:
> 
> https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md

Hmm, looks like sonnet pulled the version of my CLI claude code install. I've actually been using the claude code feature from within the desktop app, which I've been keeping up to date. That's currently at version 1.32885.1

danielweck:
Thanks for the clarification Rich, I was just being curious :)

whosawhatsis:
Since you were apparently curious enough to google me, I'll add some background.

I recommend Thorium to blind readers of my books (which contain mathml equations), and I discovered these issues while using it to spot-check the tts reading in my new book. I would otherwise submit a bug report, but since starting to use AI agents, I have decided that any bug reports or feature requests that I submit to open source projects should include PRs, so that I'm offering a fix rather than just a problem.

And if you're curious about it, no, I do not use AI when writing my books, though I have used it a little to help with developing formatting tools.

danielweck:
Reviewing and testing Pull Requests is time-consuming and diverts from other assignments, so when a contribution seems Ai-generated I do indeed check the GitHub profile in order to find evidence that my time will be wisely spent :)

I did see your books and support for visually-impaired users. Great stuff! :)

In Thorium Desktop reader version 3.5.0 we improved screen reader handling inside the content viewport (i.e. HTML Webview). The first Release Candidate of 3.5 was published recently, if all goes well the stable release will be published next week. Test results are overall pretty good with VoiceOver, NVDA and JAWS, though apparently there are some issues with Narrator. We are aware that the GUI has some focus and semantic labelling issues, but we'll be fixing this as and when we update the UI controls.

As for MathML support, users can optionally activate MathJax (the preference toggle control is located in the reader window interface, it is not exposed in the main settings in the library window). We currently ship an outdated version of MathJax but I look forward to running some tests with the newer major version. It will be interesting to see how it compares. We designed our MathJax integration so that the aria-label/alt-text/title can be read aloud by the built-in TTS feature, without preventing screen readers from accessing the inner structure which in many cases is better announced (and navigated interactively) by assistive technology and their content type plugins. MathJax has a feature that generates an accessible label for a MathML DOM fragment, but obviously the number of supported locales is limited. Most content producers include a pre-baked speech label in their MathML (often using MathJax's own NodeJS tooling!).

PR #3796, opened 2026-08-25
Google Analytics Measurement Protocol

panaC:
- Added Firebase / Measurement Protocol Protocol helper.
- Added a persisted analytics clientId Redux slice for a stable install/user identifier.
- Added startup events: app_start, app_first_open, and app_version_updated.
- Added default session_id, device, and user_properties fields.

TODO:

* Implement an event queue with `engagement_time` and timestamps, flushed at regular intervals.
* Hide the `secret_id` behind a proxy server.


src/common/analytics/publication.ts
```diff
@@ -0,0 +1,88 @@
...
+        publication.isDivina
+    ) {
+        return "divina";
+    }
+
+    if (publication.isFixedLayoutPublication) {
+        return "epub_fxl";
+    }
+
+    return "epub_reflow";
```
llemeurfr, line 67: I would expect "unknown" here, and "epub_reflow" tested against a publication flag.

danielweck: at this point in the control flow, PublicationView is created so the only remaining possible case in the switch statement is EBUB reflow, I think.

danielweck: PS: note that Daisy 2.02 and 3.0 publications (audio-only, text-only, or synchronized audio+text) are imported and converted to EPUB (Media Overlays for sync audio+text, even for audio-only which becomes structured text with synch'ed audio) ... so if we wanted to, we could report their "Daisy" nature instead of just EPUB.

danielweck: A metadata field is generated during the Daisy-to-EPUB conversion, to record the fact that the imported EPUB actually originates from a Daisy fileset.

panaC: https://app.notion.com/p/Analytics-3b5a1ca5712f8035afd9cc355a9db203?d=3c8a1ca5712f8028ab7c001c47e660ac&source=copy_link#3b5a1ca5712f801b9212dc051605bc35

panaC: `unknown` seems to be an implementation detail. However, it can be used as a fallback in case of an invalid or unexpected format.

panaC:
Event | Params
-- | --
app_start | none
app_first_open | none
app_version_updated | previous_version, current_version
import | format, media_type, optional lcp_license_type, optional lcp_provider
read | format, media_type, optional lcp_license_type, optional lcp_provider
listen | format, media_type, optional lcp_license_type, optional lcp_provider
publication_save_as | publication params
publication_delete | publication params when available
publication_add_tag | publication params
publication_mark_as | publication params + value: "finished"
publication_import_annotations | publication params when available
publication_export_annotations | publication params
catalog_add | origin | "manual", "deeplink", "pnb", "registry"
catalog_remove | none
catalog_edit | none
catalog_favorite | none
catalog_browse | none
lcp_dialog | none
lcp_passphrase | value | "valid", "invalid", "discovered"
lcp_help | none
reader_search | none
reader_bookmark_toggle | none
reader_annotate | none
reader_paginate | none
reader_history_back | none
reader_history_forward | none
reader_pref_theme | setting_value
reader_pref_layout | setting_value
reader_pref_save | setting_value
reader_pref_apply | setting_value
reader_pref_reset | setting_value
app_setting_language | setting_value
app_setting_screenreader | setting_value
app_setting_theme | setting_value
app_setting_storage | setting_value
app_setting_remove_expired | setting_value
app_setting_pnb | setting_value
app_setting_custom_annot_template | setting_value
app_setting_default_annot_template | setting_value
app_setting_minimize_library_to_tray | setting_value
app_setting_minimize_library | setting_value
app_setting_keep_library_background | setting_value
app_setting_user_one_reader_window | setting_value

</div></div></div><p>Publication params are built as:</p><p><code dir="ltr">format: "epub_reflow" | "epub_fxl" | "pdf" | "divina" | "audiobook" | "unknown"</code><br><code dir="ltr">media_type: ""</code><br><code dir="ltr">lcp_license_type: "expiring" | "permanent"</code> when LCP exists<br><code dir="ltr">lcp_provider</code> when available</p><p>All events sent through Measurement Protocol also get <code dir="ltr">session_id</code> added automatically.</p></body></html>
</body>
</html>

src/common/analytics/publication.ts
panaC: From this commit : https://github.com/edrlab/thorium-reader/pull/3796/commits/4136df719c6d1cd759f1b8f31b147991519cdc6f 

`unknown` has been removed, and new publication format types have been added. Comments have also been added for unsupported formats such as DAISY and WebPub.

panaC: https://app.notion.com/p/Analytics-3b5a1ca5712f8035afd9cc355a9db203?d=3cda1ca5712f80e7b0b4001cd04d6154&source=copy_link#3b5a1ca5712f801b9212dc051605bc35

llemeurfr: why suppressing unknown? it was clear in our conversation that this vcalue would be used as a warning in case of a bug in the code.

panaC: There is no bug in the code at this stage. We simply cannot define the `unknown` value yet. This is why we need a broader discussion about how errors should be handled in our case.

src/renderer/reader/components/picker/SearchFormPicker.tsx
```diff
@@ -167,8 +169,13 @@ class SearchFormPicker extends React.Component<IProps, IState> {
     private search(e: TFormEvent) {
         e.preventDefault();
 
+        if (!this.state.inputValue) {
```
danielweck: This COULD be a breaking change (not sure), because empty string can rightfully cause `this.props.reset();`

src/main/redux/sagas/api/opds/feed.ts
```diff
@@ -137,6 +218,17 @@ export function* findAllFeeds(): SagaGenerator<IOpdsFeedView[]> {
...
 
-    const sortedByFavorite = [...res.filter((res) => !!res.favorite), ...res.filter((res) => !(!!res.favorite))];
-    return sortedByFavorite;
+    return res.sort((a, b) => {
+        const favoriteSort = Number(!!b.favorite) - Number(!!a.favorite);
+        if (favoriteSort !== 0) {
+            return favoriteSort;
+        }
+
+        const titleSort = a.title.localeCompare(b.title, undefined, { sensitivity: "base" });
```
danielweck: `.title` can be undefined/null? (not sure)

src/preprocessor-directives.ts
```diff
@@ -59,6 +59,18 @@ export const _TELEMETRY_SECRET = __TH__TELEMETRY_SECRET__;
...
+export const _FIREBASE_ENABLED = __TH__FIREBASE_ENABLED__;
+
+// declare const __TH__FIREBASE_DEBUG__: boolean;
+// export const _FIREBASE_DEBUG = __TH__FIREBASE_DEBUG__;
+
+// declare const __TH__FIREBASE_APP_ID__: string;
+// export const _FIREBASE_APP_ID = __TH__FIREBASE_APP_ID__;
+
+declare const __TH__FIREBASE_MEASUREMENT_ID__: string;
+export const _FIREBASE_MEASUREMENT_ID = __TH__FIREBASE_MEASUREMENT_ID__;
```
danielweck: `__TH__FIREBASE_MEASUREMENT_ID__` is already in `typings.d.ts`, no need for `_FIREBASE_MEASUREMENT_ID` here.

In fact, every `export const XXX` declared here breaks dead-code elimination in the Terser / Webpack Minification plugin. Sometimes it is better to just directly access the `__TH__XXX` "defines" in the codebase (which `typings.d.ts` advertises to the Typescript compiler). I have been wondering about the utility of this `preprocessor-directives.ts` const exporter ... maybe one day we can just migrate everything to `typings.d.ts`...

src/renderer/reader/components/picker/SearchFormPicker.tsx
panaC: Good catch ! 

LLM coding agent is suffering

src/main/redux/sagas/api/opds/feed.ts
panaC: I'm reverting this part, this is my quick fix issue

PR #3798, opened 2026-08-26
compiler-assisted migration to nodeIntegration=false

danielweck:
compiler-assisted migration to `nodeIntegration=false` (and `contextIsolation=true` optionally with `sandbox=true` if `preload` can work with slim NodeJS polyfill/shim only)

run `mv ./node_modules/@types/node ./node_modules/@types/node_` (because TSCONFIG directives are overridden by other `@types/zzz` inclusions) and then run `npm run build-nodeintegration` to see a __gazillion__ errors (`node:xxx` imports, usage of `globalThis.process/Buffer/module/require/etc.`)

I created this draft Pull Request to explore and quantify the (nuclear) "blast radius" of refactoring Thorium Desktop reader's "library" and "reader" Electron "renderer process" code (i.e. the Javascript that executes in the `BrowserWindows` that host `index_library.html` and `index_reader.html` ) in order to migrate the application architecture to a proper back-end (NodeJS runtime) vs. front-end (Web platform only) separation, optionally with a slim Electron-specific "API bridge" in-between, if necessary (for secure narrow-scoped access to some platform features, via `contextBridge.exposeInMainWorld`).

Note that the Electron `preload` code attached to `BrowserWindows` or `Webviews` has different privileges determined by the configuration tuple / matrix `nodeIntegration` + `contextIsolation` + `sandbox`. The Electron `contextBridge` is configured in `preload`, so ultimately this is also an area to explore during this hypothetical costly refactoring.

This draft Pull Request is essentially a harness that leverages the TypeScript compiler AND the WebPack bundler logic, in order to pin-point exactly where things break when access to the NodeJS global context and standard library is removed. I scrolled through the extensive list of errors, and it is clear that the refactoring cost would be significant. In this PR harness, the TypeScript compilation highlights problems in our codebase at build time (which is great), but it is really the WebPack `externals` bundler hooks that provide additional build-time failure points (because it crawls the entire dependency graph) that would otherwise only detected at runtime (e.g. usage of NodeJS global context and imported modules / standard library, in transitive dependencies).


danielweck:
`require("node:module").builtinModules`

```
        _http_agent
        _http_client
        _http_common
        _http_incoming
        _http_outgoing
        _http_server
        _tls_common
        _tls_wrap
        assert
        assert/strict
        async_hooks
        buffer
        child_process
        cluster
        console
        constants
        crypto
        dgram
        diagnostics_channel
        dns
        dns/promises
        domain
        events
        fs
        fs/promises
        http
        http2
        https
        inspector
        inspector/promises
        module
        net
        os
        path
        path/posix
        path/win32
        perf_hooks
        process
        punycode
        querystring
        readline
        readline/promises
        repl
        stream
        stream/consumers
        stream/promises
        stream/web
        string_decoder
        sys
        timers
        timers/promises
        tls
        trace_events
        tty
        url
        util
        util/types
        v8
        vm
        wasi
        worker_threads
        zlib
        node:sea
        node:sqlite
        node:test
        node:test/reporters
```

PR #3818, opened 2026-08-31
chore(NPM): package dependencies updates, notably Electron 44 (was 41)

danielweck:
See:

* https://releases.electronjs.org/schedule
* https://www.electronjs.org/docs/latest/tutorial/electron-timelines
* https://github.com/electron/electron/blob/main/docs/breaking-changes.md
* https://chromiumdash.appspot.com/schedule
* https://nodejs.org/en/about/previous-releases
* https://www.electronjs.org/docs/latest/tutorial/electron-versioning

Screenshot 2026-09-02 at 10 29 12
Screenshot 2026-09-02 at 10 31 22
Screenshot 2026-09-02 at 10 30 05


danielweck:
Note: 

> macOS 12 (Monterey) is no longer supported by Chromium.
> Older versions of Electron will continue to run on Monterey, but macOS 13 (Ventura) or later will be required to run Electron v44.0.0 and higher.

https://github.com/electron/electron/blob/main/docs/breaking-changes.md#removed-macos-12-support

ping @gautierchomel @llemeurfr in relation to https://github.com/search?q=repo%3Aedrlab%2Fthorium-reader-website+monterey&type=code

danielweck:
clipboard API breaking changes (which this PR addresses):

https://github.com/electron/electron/blob/main/docs/breaking-changes.md#removed-clipboard-module-is-no-longer-available-in-the-renderer-process

PR #3822, opened 2026-09-01
fix: autoReset ReactTable opts

### LINKED ISSUE #3821: Search UX: All Publications view should not be reset after a publication is open
Steps:
- Go to All Publications
- Make a search (e.g. "pdf")
- Open one publication
- Move the reading window on the side to have a look at the library window: the search term has disappeared, and I get the full list of publications. 

Redoing the search every time is cumbersome. 

Worse: 
- Go to All Publications
- Create a filter (e.g. By Format, "pdf")
- Open one publication
- Move the reading window on the side to have a look at the library window: the filter still appears as it was set, BUT I now get the full list of publications.

This is totally wrong.

panaC:
React Table v7 enables its `autoReset*` options by default, which is useful for fresh datasets but broken here. Thorium refreshes publication data while the user is still in the same library view. That caused search terms, column filters, pagination, and sorting-related state to be reset unexpectedly.

This PR explicitly disables the full React Table v7 auto-reset option inventory for the All Publications table and documents each option with its API/source reference.

Fixes #3821 

Pagination, tags, add/remove, bulk import, text search, and opening a publication alongside the list when a search/filter is applied; all of this has been tested and is working.


panaC:
@arthur-lemeur Can you have a look please

danielweck:
We need to test that when a publication is added/removed or mutated (e.g. different tags) the table refreshes its internal backing store.

panaC:
I’m marking this PR as ready for review. All my tests have passed successfully.

danielweck:
I assume the table pagination state is now preserved too? when publications are removed, and the current page is the last one, does the pagination cursor stay on the last page or does it reset to number one?

panaC:
> I assume the table pagination state is now preserved too? when publications are removed, and the current page is the last one, does the pagination cursor stay on the last page or does it reset to number one?


Good catch !

arthur-lemeur:
Seems good to me!

arthur-lemeur (approved):

panaC:
@arthur-lemeur 

> Seems good to me!

There is still a bug as mentioned by Daniel.

danielweck:
> There is still a bug as mentioned by Daniel.

Well, potential bugs due to stale state, requiring invalidation and subsequent restoration of scroll/pagination/filtering/whatever (I haven't tested the PR, just read the code)

PR #3825, opened 2026-09-01
GA4 telemetry event queue implementation

### LINKED ISSUE #3823: Introduce a GA4 Measurement Protocol event queue with batched delivery
- Add a GA4 Measurement Protocol event queue owned by the Electron main process.
- Persist queued events on disk so events are not lost when the app is offline or closed before delivery.
- Batch events according to GA4 Measurement Protocol limits.
- Flush the queue on startup, on a timer, and when the queue reaches the configured batch size.

GA4 Measurement Protocol accepts HTTPS `POST` requests to `/mp/collect`. The payload includes an `events` array, with up to 25 events per request, and the JSON POST body must remain under 130 KB.

panaC:
Fixes #3823 

The purpose of this PR is to replace the Firebase SDK and gtag.js queueing layer with a Thorium-managed queue for GA4 Measurement Protocol events.

This PR introduces a persistent delivery queue for GA4 Measurement Protocol events in the Electron main process. Non-debug events are normalized into individual queue records containing a UUID, enqueue timestamp, request-level metadata, and event payload. Queue mutations are serialized to prevent concurrent enqueue and flush operations from corrupting in-memory or persisted state.

Events are flushed at application startup, every 60 seconds, when the queue reaches the 25-event GA4 batch limit, and during application shutdown with a five-second timeout. Batches preserve FIFO order, contain only events with compatible request metadata, are limited to 25 events, and remain below the 130 KiB request-body limit. Expired events older than 72 hours and oversized events are discarded. The queue is capped at 1,000 events, dropping the oldest entries first.

Successful deliveries are removed from storage. Network failures, missing or disabled configuration, HTTP 408, HTTP 429, and server errors retain the current batch for a later retry. Permanent failures remove the affected batch so subsequent events can continue processing. Debug validation events bypass persistence and are sent immediately so validation messages can be returned to the caller.

Queue persistence is abstracted behind interchangeable JSON and SQLite stores. JSON remains the default and uses flushed temporary-file writes followed by atomic replacement, including retries for transient Windows locking errors. SQLite can be enabled at build time with THORIUM_MEASUREMENT_PROTOCOL_QUEUE_SQLITE_ENABLED and rewrites the ordered queue inside a transaction with rollback on failure. The two stores are independent; no migration or fallback occurs when changing the feature flag.

Tests cover persistence and reload, SQLite storage, batching limits, payload-size enforcement, retryable and permanent failures, queue capacity, event expiry, debug delivery, persisted-data sanitization, and transient Windows rename failures.

Features Added
- Reliable analytics event delivery
- Persistent offline event storage
- Configurable JSON or SQLite backend
- Automatic event batching
- Retry handling for temporary failures
- Protection against invalid or stale data
- Controlled queue size and retention
- Safe application startup and shutdown handling
- Concurrency-safe queue processing


webpack.config-preprocessor-directives.js
```diff
@@ -1,5 +1,5 @@
 const webpack = require("webpack");
-// require("dotenv").config({ quiet: true });
+require("dotenv").config({ quiet: true });
```
panaC, line 2: comment this line before merge

panaC:
This PR adds an alternative SQLite persistence backend for the Measurement Protocol event queue.

The backend is selected at build time with the static feature flag `THORIUM_MEASUREMENT_PROTOCOL_QUEUE_SQLITE_ENABLED`.

Both implementations share the same queue behavior: ordered persistence, reload after restart, expiration and size cleanup, retry preservation, and removal after successful delivery. SQLite updates are performed inside a transaction.

Current limitation: changing the flag does not migrate pending events between the JSON and SQLite backends; each backend retains its own queue.

panaC:
```mermaid
flowchart TD
    A[Thorium starts] --> B{Analytics configuration present?}
    B -- No --> B1[Leave queue inactive]
    B -- Yes --> C{SQLite feature flag enabled?}

    C -- Yes --> D[Use SQLite queue store]
    C -- No --> E[Use JSON file queue store]
    D --> F[Load persisted events]
    E --> F

    F --> G["Sanitize events<br/>Discard events older than 72 hours<br/>Keep newest 1,000 events"]
    G --> H["Start 60-second flush timer<br/>Request startup flush"]

    I[Telemetry event from app or renderer IPC] --> J{Telemetry enabled<br/>and event/config valid?}
    J -- No --> J1[Return failure reason without throwing]
    J -- Yes --> K["Build GA4 event<br/>Add client, session, locale,<br/>app, OS and device data"]

    K --> L{Debug mode?}
    L -- Yes --> M["Send event immediately<br/>Do not persist it<br/>Return validation messages"]
    L -- No --> N["Create queued record<br/>UUID + timestamp + body metadata"]

    N --> O["Acquire exclusive queue lock<br/>Append event<br/>Drop oldest if over 1,000"]
    O --> P[Persist complete queue]
    P --> Q{Queue contains at least 25 events?}
    Q -- No --> R[Wait for another flush trigger]
    Q -- Yes --> S[Request flush]

    H --> S
    T[60-second timer] --> S
    U["Application shutdown<br/>maximum wait: 5 seconds"] --> S

    S --> V{Flush already running?}
    V -- Yes --> V1[Join existing flush]
    V -- No --> W["Drop expired events<br/>Build next FIFO batch"]

    W --> X{Batch available?}
    X -- No --> X1[Flush complete]
    X -- Yes --> Y["Batch compatible events only<br/>Same body metadata<br/>Maximum 25 events<br/>Body below 130 KiB"]

    Y --> Z{First event alone is oversized?}
    Z -- Yes --> ZA[Remove oversized event and persist]
    ZA --> W
    Z -- No --> ZB[POST batch to GA4]

    ZB --> ZC{Delivery successful?}
    ZC -- Yes --> ZD[Remove delivered events and persist]
    ZD --> W

    ZC -- No --> ZE{Retryable failure?}
    ZE -- Yes --> ZF["Keep batch persisted<br/>Stop until a later flush"]
    ZE -- No --> ZG["Drop permanently failed batch<br/>Persist and continue"]
    ZG --> W
```

Persistence behavior:

- SQLite uses `BEGIN IMMEDIATE → DELETE → ordered INSERT → COMMIT`, with rollback on failure.
- JSON uses a flushed temporary file followed by an atomic rename, including retries for transient Windows file locks.
- Retryable delivery failures include network errors, missing/disabled configuration, HTTP `408`, `429`, and `5xx`. Other permanent failures are discarded so they cannot block later events.

PR #3829, opened 2026-09-01
Add Google Analytics telemetry opt-out setting

### LINKED ISSUE #3824: Add an opt-out option in Library settings to disable Google Analytics telemetry

panaC:
Fixes #3824 

This PR adds a user-facing option to opt out of Google Analytics telemetry in Thorium Reader.

The preference is exposed in the application settings, persisted in Redux state, and applied to analytics events across the main and renderer processes. When telemetry is disabled, analytics requests are skipped before device information is collected or any network request is made.

The implementation also includes the required localization strings so the setting is available consistently across supported languages.


panaC:
The telemetry opt-out is currently enforced per call via the optional `disabled` parameter. This means any current or future call to `logMeasurementProtocol()` that forgets to pass the flag could still send analytics even when the user has opted out.

Since this is a privacy setting, it would be safer to enforce the preference centrally at the analytics boundary rather than relying on every caller to provide it correctly.

At minimum, consider making `disabled` required so TypeScript catches missing values. Ideally, the analytics sender itself should read/enforce the current telemetry preference and fail closed when telemetry is disabled.

PR #3845, opened 2026-09-04
migrate tabs from Radix to RAC in library and reader settings

arthur-lemeur:


danielweck:
Merci @arthur-lemeur 
:)

I won't have time to test for a while, let alone properly review the code changes (I just had a glance).

But assuming you tested mouse, keyboard, and screen reader interaction ... this looks good to me.

Once this lands in the develop branch, then other users will be able to test and report.

I am thinking particularly about users of assistive technology, who need a scoped description of what to test, so that they don't waste their precious time wondering about the GUI changes.

I will ping a few known contributors / participants, starting with @GeorgeKerscher @nvdaes @togregg @TaF-WCAG @mlouie-prcvi @Marc2p @amirsol81 @cary-rowen @spudthebud @typopaul @lpintes @m-c-g @sukiletxe @abhishekbaghel05 @RastislavKish @jimmyfreiding

...and @gautierchomel

GeorgeKerscher:
Happy to test as soon as I get the go ahead. I am wondering about what should we do about Narrator testing. I was planning to initiate the engagement with Microsoft once 3.5.1 is in the Microsoft store and available on the EDR Lab download page. They need to fix three or four failuers.

Would this new interface mean they would need to redo that work?

danielweck:
Hi George, the current graphical user interface is based on a mix of GUI libraries. Originally this was a good engineering decision, but since then, one particular library has stood out with its dedication to screen reader accessibility and general keyboard usability (it is maintained by Adobe and it is called React Aria Components, or RAC). The other GUI library has become less maintained as the project team pivoted on another business (Radix UI, now Base UI).

We aim to remove one library and unify the GUI with our preferred library. This harmonization should eliminate a few edge-case keyboard focus / navigation bugs, or at the very least pave the road to fix the remaining issues.

This will not impact the EPUB HTML content frame, unless of course the updated GUI components steal keyboard focus unexpectedly (which we will test for and implement mitigations for, if necessary).

This particular Pull Request targets specifically the tab controls (for example in the settings dialogs), but I will let my colleague @arthur-lemeur explain what changed / improved.

nvdaes:
Thanks @danielweck. I can test with NVDA if you need it, if you provide the needed steps.

panaC (changes requested):
Requesting changes: the React Aria migration breaks existing focus targets.

React Aria uses each `id` as a collection key and generates its own DOM IDs. As a result:

- Reader keyboard shortcuts still search for `reader-menu-tab-*-trigger` and `reader-menu-tab-*`, but those elements no longer exist.
- The Text and Spacing indentation selectors never match; use the existing `data-value` attributes instead.
- `TabList` should use `React.forwardRef`; current refs remain `null` and React emits a warning.

panaC:
The migration exposed a mismatch between Thorium’s focus-management contract and React Aria’s ID model.

React Aria treats `Tab.id` and `TabPanel.id` as collection keys. It deliberately removes those values from the rendered DOM and generates IDs such as `react-aria-…-tab-tab-search` to maintain its `aria-controls`/`aria-labelledby` relationships. Thorium, however, still resolves focus targets using stable IDs such as `reader-menu-tab-search`. Overriding React Aria’s generated IDs would risk breaking its accessibility wiring.

The fix therefore separates the two namespaces:

- React Aria continues owning DOM `id`, `aria-controls`, and `aria-labelledby`.
- Thorium exposes its stable focus contract through `data-focus-id`.
- [focusTarget.ts](C:/Users/plero/Documents/git/thorium-reader/src/renderer/common/focusTarget.ts:8) resolves `data-focus-id` first and falls back to legacy DOM IDs, so existing non-tab focus targets remain compatible.
- The dialog mutation observer now uses this resolver when asynchronously mounted content appears.
- Tabs, tab panels, and docked select buttons expose the identifiers expected by the existing keyboard-shortcut actions.

React Aria may omit `tabIndex` from a tab panel when the panel already contains a tabbable descendant. Thorium’s shortcuts focus the panel itself, so the helper adds `tabIndex="-1"` only when required. This enables programmatic focus without adding the panel to sequential keyboard navigation.

The `TabList` abstraction was also a normal function component even though callers passed it refs. Under React 18, that leaves the ref null and emits “Function components cannot be given refs.” It is now implemented using a typed `React.forwardRef`, preserving the generic item type and forwarding the ref to the React Aria DOM element.

Finally, the CSS previously matched tabs using exact DOM IDs. Because React Aria generates those IDs, the selectors were unreachable. They now use the existing stable `data-value` attributes.

src/renderer/library/components/settings/Settings.tsx
```diff
@@ -80,16 +80,24 @@ export const Settings: React.FC<ISettingsProps> = () => {
...
 
-    const [tabTitle, setTabTitle] = React.useState(__("settings.tabs.general"));
+    const [selectedTab, setSelectedTab] = React.useState("tab1");
 
-    React.useEffect(() => {
-        setTabTitle(__("settings.tabs.general"));
-    }, [__, locale]);
+    const tabTitle = {
+        tab1: __("settings.tabs.general"),
+        tab2: __("settings.tabs.appearance"),
```
panaC: `tab3`  !?

panaC: `tab<number>`  is not a good practice AFAIK

PR #3846, opened 2026-09-07
migrate radiogroup from Radix to RAC in library and reader settings

arthur-lemeur:


panaC (changes requested):
React Aria focuses a visually hidden native input and exposes focus state on the wrapping label via `data-focus-visible`/`data-focused`. `annotations_filter_tag` does not style those attributes, so these creator-type radios lose the visible keyboard focus outline previously provided by the Radix button. Add a focus-state selector to that class or provide one in the shared Radio component.

panaC:
The regression came from a DOM change introduced by the Radix-to-React Aria migration.

Previously, Radix rendered each option as a focusable `<button>`. Thorium’s global keyboard-focus CSS therefore drew an outline directly around the visible option:

```html
<button data-state="checked">Person</button>
```

React Aria renders a different structure:

```html
<label data-selected data-focus-visible>
  <span class="visually-hidden">
    <input type="radio">
  </span>
  Person
</label>
```

The actual browser focus is on the native `<input>`, but that input is visually hidden. Consequently, the global `*:focus` outline was also hidden. React Aria communicates the visible focus state by adding `data-focus-visible` to the wrapping `<label>`.

The fix in [annotations.scss](C:/Users/plero/Documents/git/thorium-reader/src/renderer/assets/styles/components/annotations.scss:137) styles that state explicitly:

```scss
&[data-focus-visible] {
  outline: var(--color-brand-primary) solid 2px !important;
  outline-offset: 2px;
}
```

Using `data-focus-visible` instead of `data-focused` is intentional: it displays the outline for keyboard-style focus navigation without unnecessarily showing it after ordinary pointer clicks. Selection styling remains independent through `data-selected` and `input:checked`.

arthur-lemeur:
You're right I missed one in annotations, good catch!

PR #3850, opened 2026-09-08
feat: support the color set defined by the thorium-reader-website badge generator

### LINKED ISSUE #3847: Add OPDS feed color from the Thorium catalog badge deeplink
https://github.com/edrlab/thorium-reader/blob/f74c991d3b08f3ff7fe9b88956dd44dd397aad9d/src/main/redux/sagas/event.ts#L304

panaC:
Fixes #3847 

support the color set defined by the thorium-reader-website badge generator https://github.com/edrlab/thorium-reader-website/blob/59c09aa5af6f4cc86ef87333f6f80b0057116216/docs/badge-page.md?plain=1#L33

This color setting is optional and fallback to gray by default.

When an OPDS feed is added manually, the default color is saved along with the feed information. However, when a feed is added via a deep link without a `color` field, or from the CLI, the `undefined` color value is persisted and the view fallback to gray.

The color palette dark and light are taken from the Thorium-Mobile implementation and may need to be refined in the future.

I tested and reviewed each color in both light and dark modes, and they look good to me.

The color is applied on the opds feed card, banner, border and text.


llemeurfr:
Good. What is the point of having localized labels for these colors? Where does it appear in the user interface? Is it because the user can set or change an OPDS badge color?

panaC:
Yes, a color picker has been added to the UI when manually adding an OPDS feed. The default color remains gray.

The user can enter the URL, title, favorite option, and now its color. These labels are used by screen readers and tooltips. The component is the same as the annotation highlight color picker, but with different color names. To keep these concerns separate, I added dedicated color labels to the localization files.

> Is it because the user can set or change an OPDS badge color?

Yes. The user can change the color even when it was initially defined through the deep link.

PR #3858, opened 2026-09-09
Persist catalog deeplink LCP passphrases as generic secrets

### LINKED ISSUE #3854: persist the lcp passphrase from a catalog deeplink entry
https://github.com/edrlab/thorium-reader/blob/1b0e31458cff9c80487c408da8a80e7e4059261a/src/main/services/lcp.ts#L178

https://github.com/edrlab/thorium-reader/blob/1b0e31458cff9c80487c408da8a80e7e4059261a/src/main/redux/sagas/event.ts#L431

panaC:
Fixes #3854 

Adds support for storing LCP passphrases received from catalog deeplinks without changing or migrating the existing secrets vault structure.

Catalog deeplink passphrases are now saved as generic vault entries using a deterministic synthetic key, which makes repeated imports idempotent. LCP secret lookup now follows the desktop equivalent of the mobile fallback cascade:

1. publication/license-specific secret
2. provider-matching secret
3. generic/unscoped secret

This keeps existing publication-keyed secrets compatible while allowing catalog-provided passphrases to unlock future publications.


llemeurfr:
I guess that if an initially unscoped passphrase matches a license, it is then scoped by the corresponding license id and provider url for future use. Isn't it?

panaC:
> I guess that if an initially unscoped passphrase matches a license, it is then scoped by the corresponding license id and provider url for future use. Isn't it?

No, generic secrets are tried, but a successful generic key is not promoted to the publication/provider entry because `unlockPublication()` passes all stored hashes at once and the wrapper only reports success/failure, not which hash won.

llemeurfr:
The LCP lib returns the winning passphrase hash. Returning it to the app would be a useful evolution of the wrapper, if it isn't a pain.

Issue #1981, opened 2023-08-23
OPDS 2 cover retrieval from HTTP does not use credentials?


gotson:
I am adding OPDS 2 to [Komga](https://komga.org/), but i have some issues for cover retrieval when using OPDS 2.

Covers are behind the same HTTP Basic authentication as the OPDS feed itself (OPDS 1.2 or 2), but in that case Thorium does not show the covers.

By turning off the authentication completely, the covers are showing properly.

I am suspecting that Thorium does not pass the session cookies in the HTTP header when requesting the covers.

With an OPDS 1.2 catalog that has the same authentication mechanism Thorium can retrieve the covers just fine.

I tried finding Thorium logs on my Mac but couldn't find anything.

danielweck:
mhmm, interesting that this seems to affect OPDS2(JSON) but not OPDS1(XML)

gotson:
So i was testing v1 using my deployed server, and v2 using localhost.

I added v1 using localhost, and i don't have the covers either. So it may be about HTTP vs HTTPS ? My deployed server is using HTTPS, but on localhost i am using HTTP.

gotson:
@danielweck i have forked the OPDS test catalog, added basic auth by using Caddy server, and reproduced the issue.

You can get the project from this branch https://github.com/gotson/opds2-test-catalog/tree/basic-auth, i added instructions in the README on how to run it.

When added to Thorium, the covers will not show.

panaC:
Isn't relative from this https://github.com/edrlab/thorium-reader/pull/1832#issuecomment-1276450705 ?

adamshand:
Just ran into this same bug with Thorium (v3.4.0) when writing my own OPDS server.  Covers load correctly on the index page, but fail to load on the book detail page.

Agent points to the below as probable bug.

---

Yes—there’s an obvious-looking state bug that fits your symptoms. I haven’t reproduced it in a running  Thorium build, but the code path is convincing.

In ../thorium-reader/src/renderer/common/components/Cover.tsx:

1. The cover initially loads through a normal <img> request.
2. On failure, imageOnError() retries through Thorium’s authenticated media handler and sets imgErroredOnce: true.
3. When publication details arrive, componentDidUpdate() resets the image to its original HTTPS URL—but doesn’t reset that flag.
4. The image request fails authentication again, and imageOnError() immediately returns because it already retried once.

That explains “cover appears, details load, cover breaks.”

[...]

PR #3860, opened 2026-09-10
Fix authenticated OPDS covers breaking after publication detail refresh

adamshand:
Target #1981

The issue happens because `componentDidUpdate()` currently reacts to the `cover` object changing.

When the publication detail is refreshed, we can receive a new `cover` object even if the actual cover URL is the same. This resets `imgUrl` back to the original HTTP(S) URL and loses the authenticated `opds-media://` fallback that was set after the first image load failure.

Since `imgErroredOnce` is already `true`, the fallback is not retried, so the cover stays broken.

The fix is to compare the effective cover URL instead of the `cover` object reference, and only reset `imgUrl` / `imgErroredOnce` when that URL actually changes.


panaC:
Could you please remove all unused fixture files and tests from the PR, and keep only this patch with a concise PR comment?

I’ll then merge the PR into `develop`.

Good catch, and thank you for your work!


```
diff --git a/src/renderer/common/components/Cover.tsx b/src/renderer/common/components/Cover.tsx
index ... .. ...
--- a/src/renderer/common/components/Cover.tsx
+++ b/src/renderer/common/components/Cover.tsx
@@ -48,6 +48,16 @@ interface IState {
     imgErroredOnce: boolean,
 }
 
+function getCoverUrl(props: Readonly<IProps>): string | undefined {
+    const { cover } = props.publicationViewMaybeOpds;
+    if (!cover) {
+        return undefined;
+    }
+    const coverUrl = cover.coverUrl || cover.coverLinks[0]?.url;
+    const thumbnailUrl = cover.coverUrl || cover.thumbnailLinks[0]?.url;
+    return props.coverType === "cover" ? coverUrl || thumbnailUrl : thumbnailUrl || coverUrl;
+}
+
 class Cover extends React.Component<IProps, IState> {
 
     public uuid: string;
@@ -76,20 +86,10 @@ class Cover extends React.Component<IProps, IState> {
         this.imageOnError = this.imageOnError.bind(this);
     }
     public componentDidUpdate(prevProps: Readonly<IProps>): void {
-        if (prevProps.publicationViewMaybeOpds?.cover !== this.props.publicationViewMaybeOpds?.cover) {
-
-            const { cover } = this.props.publicationViewMaybeOpds;
-
-            if (cover) {
-                const coverUrl = cover.coverUrl || cover.coverLinks[0]?.url;
-                const thumbnailUrl = cover.coverUrl || cover.thumbnailLinks[0]?.url;
-                if (this.props.coverType === "cover") {
-                    this.setState({ imgUrl: coverUrl || thumbnailUrl });
-                } else {
-                    this.setState({ imgUrl: thumbnailUrl || coverUrl });
-                }
-            } else {
-                this.setState({ imgUrl: undefined });
-            }
+        const imgUrl = getCoverUrl(this.props);
+        if (getCoverUrl(prevProps) !== imgUrl) {
+            this.setState({ imgUrl, imgErroredOnce: false });
         }
     }
     ```

adamshand:
Thanks for the review! Removed the fixtures and tests and kept only your requested patch, including leaving the constructor unchanged. Shortened the PR description too. Targeted ESLint passes.

PR #3865, opened 2026-09-10
Fixes LCP CRL handling with HTTP cache control

### LINKED ISSUE #3863: LCP CRL handling with cache control
- Store the last successfully downloaded and validated CRL in memory.
- Store its `ETag` and last successful validation timestamp in memory.
- When opening an LCP license, always use the cached CRL immediately, **even if the cache has expired**.
- If the cache has expired, trigger a CRL refresh asynchronously for the **next** license opening.
- Use `If-None-Match` with the cached ETag when refreshing:

  * `304 Not Modified`: keep the existing CRL and update its freshness timestamp.
  * `200 OK`: validate the response as a valid X.509 CRL before replacing the cache and store the new ETag.
  * Network/HTTP/validation error: keep the existing CRL untouched.
- Ensure that an invalid response (e.g. HTML returned by a captive portal with `200 OK`) can never overwrite a valid cached CRL.
- No concurrent refresh requests so that only one CRL download is performed at a time.

Cache expiration should therefore represent **freshness, not usability**: a stale CRL can still be used to open the current license while a fresh version is fetched in the background.

panaC:
Fixes #3863 

The algorithm keeps the last successfully downloaded CRL in memory and uses it immediately when opening an LCP license, even if it is stale.

A stale CRL is still usable. Expiration only means that Thorium should refresh the CRL in the background for the next license opening.

`Cache-Control` is used only to decide when the CRL cache is considered **stale**, not whether it is usable.

Current freshness logic:

- If response has `Cache-Control: max-age=N`, the cache expires after `N` seconds.
- If response has `Cache-Control: no-cache` or `no-store`, the cache is immediately expired.
- If no usable `Cache-Control` directive exists, Thorium uses the fallback freshness duration, currently 1 hour.

When the cache is expired:

- Thorium still returns the cached CRL immediately for the current license opening.
- Thorium starts an asynchronous refresh in the background.
- The refreshed CRL, if successfully fetched, is used by the next license opening.

So “expired” means “needs refresh”, not “cannot be used”.

On refresh, Thorium sends conditional HTTP headers when available: `If-None-Match` with the cached `ETag`, and `If-Modified-Since` with the cached `Last-Modified`.

If the server returns `304 Not Modified`, Thorium keeps the existing CRL and updates its freshness metadata. If the server returns `200 OK` with `application/pkix-crl`, Thorium replaces the cached CRL and stores the new `ETag`, `Last-Modified`, validation timestamp, and freshness expiry.

If the refresh fails because of a network error, HTTP error, or unexpected content type, Thorium keeps the existing CRL untouched.

Breaking change: 

Thorium returns `DUMMY_CRL` before any CRL has been successfully downloaded. Opening a license immediately if the initial refresh fails or not yet downloaded causes the dummy CRL to be used.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -98,12 +98,13 @@ const getCacheControlMaxAgeMs = (cacheControl: string | undefined): number | und
     let maxAgeMs: number | undefined;
     for (const directive of cacheControl.split(",")) {
         const [rawName, rawValue] = directive.trim().split("=", 2);
-        const name = rawName.toLowerCase();
+        const name = rawName.trim().toLowerCase();
+        const value = rawValue?.trim();
         if (name === "no-cache" || name === "no-store") {
             return 0;
         }
-        if (name === "max-age" && rawValue) {
-            const seconds = Number(rawValue.replace(/^"|"$/g, ""));
+        if (name === "max-age" && value) {
+            const seconds = Number(value.replace(/^"|"$/g, ""));
             if (Number.isFinite(seconds) && seconds >= 0) {
                 maxAgeMs = seconds * 1000;
             }
@@ -122,7 +123,7 @@ const isLcpCrlCacheExpired = () =>
 
 const refreshLcpCrlCache = (): Promise<void> => {
     debug("REFRESH LCP CRL REQUEST", lcpCrlCache);
-    if (lcpCrlCache.refreshPromise) {
+    if (typeof lcpCrlCache.refreshPromise !== "undefined") {
         return lcpCrlCache.refreshPromise;
     }
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -74,6 +74,7 @@ setLcpNativePluginPath(lcpNativePluginPath);
 interface ILcpCrlCache {
     crlPem: string;
     etag: string | undefined;
+    lastModified: string | undefined;
     validatedAt: number;
     refreshPromise: Promise<void> | undefined;
 }
@@ -81,12 +82,13 @@ interface ILcpCrlCache {
 const lcpCrlCache: ILcpCrlCache = {
     crlPem: DUMMY_CRL,
     etag: undefined,
+    lastModified: undefined,
     validatedAt: 0,
     refreshPromise: undefined,
 };
 
 // 1h
-const LCP_CRL_CACHE_FRESHNESS_MS = 60 * 60 * 1000;
+const LCP_CRL_CACHE_FRESHNESS_MS = 0;// 60 * 60 * 1000;
 
 // TODO: HTTP cache-control Header !?
 const isLcpCrlCacheExpired = () =>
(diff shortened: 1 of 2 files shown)


panaC:
```mermaid
flowchart TD
    A[App startup] --> B[Start async CRL refresh]
    B --> C{Refresh already running?}
    C -- Yes --> D[Reuse existing refresh promise]
    C -- No --> E[Send CRL request]

    E --> F[Add Accept: application/pkix-crl]
    F --> G{Cached ETag exists?}
    G -- Yes --> H[Add If-None-Match]
    G -- No --> I{Cached Last-Modified exists?}
    H --> I
    I -- Yes --> J[Add If-Modified-Since]
    I -- No --> K[Fetch CRL]
    J --> K

    K --> L{HTTP response}
    L -- 304 Not Modified --> M[Keep cached CRL]
    M --> N[Update validatedAt]
    N --> O[Compute expiresAt from Cache-Control]
    O --> P[Refresh complete]

    L -- 200 OK + application/pkix-crl --> Q[Read response bytes]
    Q --> R[Convert DER bytes to PEM]
    R --> S[Replace cached CRL]
    S --> T[Store ETag and Last-Modified]
    T --> N

    L -- Error / invalid status / invalid content type --> U[Keep existing cache untouched]
    U --> P

    V[Open LCP license] --> W[Return cached CRL immediately]
    W --> X{Cache expired?}
    X -- Yes --> Y[Trigger async refresh for next opening]
    X -- No --> Z[No refresh needed]
    Y --> C
```

panaC:
Breaking change: 

Thorium’s in-memory CRL cache is initialized with `DUMMY_CRL`:

At startup, Thorium starts a CRL download asynchronously, but it does not wait for it to finish.

When an LCP license is opened, it immediately returns the current cached value. If the startup download has not completed yet, or if it failed because Thorium is offline, the current cached value is still `DUMMY_CRL`.

So the risky sequence is:

1. Thorium starts.
2. Cache contains only `DUMMY_CRL`.
3. Thorium starts downloading the real CRL in the background.
4. User opens an LCP license before that download succeeds, or while offline.
5. `setCRLGetter()` returns `DUMMY_CRL`.
6. The native LCP plugin validates the license using an empty/dummy revocation list.
7. A license from a revoked/excluded provider may be accepted because the real CRL was never available.

The key problem is that `DUMMY_CRL` is acting as both a technical fallback and as the cached CRL value. It should not be treated as a successfully downloaded CRL.

**Solution**: bundled the latest CRL instead of DUMMY_CRL

Instead of initializing the cache with `DUMMY_CRL`, Thorium can bundle a recent real CRL with the application and use it as the initial cache value.

Proposed behavior:

1. On startup, initialize the in-memory CRL cache with the bundled CRL.

This avoids the first-use offline problem: Thorium never validates against an empty `DUMMY_CRL`, and revoked/excluded providers present in the bundled CRL remain blocked.

See Issue https://github.com/edrlab/thorium-reader/issues/3864

panaC:
```
curl -v   'http://crl.edrlab.telesec.de/rl/EDRLab_CA.crl'   -H 'Accept: application/pkix-crl' -H 'User-Agent: thorium-desktop'
```

```
< HTTP/1.1 200 OK
< Server: Apache
< Last-Modified: Thu, 10 Sep 2026 14:25:01 GMT
< ETag: "295-65b21bbb7a7bb"
< Accept-Ranges: bytes
< Content-Length: 661
< Content-Type: application/pkix-crl
< Cache-Control: max-age=8538
< Date: Fri, 11 Sep 2026 09:03:14 GMT
< Connection: keep-alive
```

Looks like a 24-hour cache.

PR #3879, opened 2026-09-16
hide the import notes button when license not valid

### LINKED ISSUE #3870: Associating annotations with a publication that has an expired license makes no sense
One can "Import notes" on a "Loan expired" publication. Please remove this affordance, which makes no sense. 
It is important to be able to "Export notes" from an expired loan, but importing is of no use. 

Image

arthur-lemeur:
Fixes #3870


src/renderer/library/components/publication/menu/CatalogMenu.tsx
```diff
@@ -66,6 +66,17 @@ const CatalogMenu: React.FC<{ publicationView: PublicationView }> = (props) => {
 
     const canOpen = canOpenPublication(props.publicationView);
 
+    const isLicensed = props.publicationView.lcp !== null && props.publicationView.lcp !== undefined; 
+
+    let isLicenseReadyorActive = false;
```
panaC: add a ternary here

src/renderer/library/components/publication/menu/CatalogMenu.tsx
```diff
@@ -181,6 +192,8 @@ const CatalogMenu: React.FC<{ publicationView: PublicationView }> = (props) => {
             <PublicationExportButton
                 publicationView={props.publicationView}
             />
+            {!isLicensed || isLicenseReadyorActive ?
```
panaC: use one variable instead

PR #3884, opened 2026-09-16
Preserve unresolved annotation targets during import and export

### LINKED ISSUE #3881: Do not reject an annotation set when one target.source cannot be resolved
The [current import flow](https://github.com/edrlab/thorium-reader/discussions/3859#discussioncomment-18461018) requires every annotation target.source to resolve to exactly one resource in the current publication spine.

Source resolution currently uses:
1. An exact spine href match.
2. Normalized path or URL matching.
3. A unique suffix match.

When no spine item matches, or when multiple spine items match, the source is considered rejected. If this happens for a single annotation, Thorium displays message.annotations.noBelongTo and aborts the entire import.

This heuristic does not demonstrate that the annotation belongs to another publication. It only indicates that Thorium could not uniquely associate its source with a spine resource.

Change this validation from a set-level rejection to per-annotation resolution:
- Keep the current source matching heuristic.
- Normalize target.source to the exact spine href when a unique match is found.
- When no unique match is found, preserve the original target and import the annotation without a resolved locator.
- Continue processing the remaining annotations instead of rejecting the complete set.
- Keep unresolved annotations in storage and conflict classification.
- Do not attempt navigation or highlight rendering until their source and locator can be resolved.
- When exporting an unresolved annotation, reuse its preserved original target.

The import should only stop completely when the annotation-set document itself cannot be parsed or validated. Selector resolution and confirmed publication mismatches remain separate concerns.

panaC:
Fixes #3881


panaC:
need #3885 to be merged

PR #3895, opened 2026-09-21
fix(net): connect directly when the PAC file cannot be loaded

### LINKED ISSUE #3894: Downloads fail with "getaddrinfo ENOTFOUND wpad" when macOS Auto Proxy Discovery is on and no wpad host exists
## Environment
- Thorium 3.5.1 (also reproduced on develop 4486e92e)
- macOS 26.6.2
- Auto Proxy Discovery enabled

## Repro Steps

1. System Settings > Network > Ethernet or Wi-Fi > Details > Proxies > enable Auto Proxy Discovery.
2. Confirm scutil --proxy shows ProxyAutoConfigEnable : 1 and ProxyAutoConfigURLString : http://wpad/wpad.dat.
3. Open any .lcpl or add any OPDS feed.

## Expected
Thorium connects directly, mimicking the behavior of most popular browsers when the PAC file is unreachable.

## Actual

The download produces this toast: `Downloading [title] failed: [HTTP download failed: FetchError: request to https://... failed, reason: getaddrinfo ENOTFOUND wpad () [...]]`

mpdunlop:
# Problem

macOS Auto Proxy Discovery reports `http://wpad/wpad.dat` even when no wpad host is defined, and `pac-proxy-agent` rethrows the PAC load error. Thorium displays `FetchError: request to <url> failed, reason: getaddrinfo ENOTFOUND wpad`. 

As a result, every download, OPDS feed, LCP CRL fetch and update check fails in 3.5.0 and 3.5.1 with this setting when no wpad host is defined.

# Change

A new `PacProxyAgentWithDirectFallback` was created in `src/main/network/proxy-agent.ts`. It returns the direct HTTP or HTTPS agent when a call to `getResolver()` is rejected.

As a result, downloads, OPDS feeds, LCP CRL fetches and update checks fail in 3.5.0 and 3.5.1 when this setting is enabled and no wpad host is defined.

## Precedent

Chromium and Firefox both connect directly on PAC download failures.

# Verification

On an M4 Mac mini running macOS 26.6.2, `develop` commit `4486e92e`, with Auto Proxy Discovery enabled and nslookup wpad returning NXDOMAIN:

- Unpatched develop fails an .lcpl download with ENOTFOUND wpad.
- The patched build logs "PAC file 'http://wpad/wpad.dat' could not be loaded, connecting directly" and imports the publication.
- With Auto Proxy Discovery disabled, downloads succeed and the fallback is never entered.

Fixes #3894.


danielweck:
Thank you for discovering this issue and for proposing a fix!
Instead of sub-classing, I am wondering whether we could leverage the existing `DIRECT` method:

https://github.com/TooTallNate/proxy-agents/blob/4813885d3f4e2ff837878ceffdba656a71dc31f0/packages/pac-proxy-agent/src/index.ts#L261-L263

https://github.com/TooTallNate/proxy-agents/blob/4813885d3f4e2ff837878ceffdba656a71dc31f0/packages/pac-proxy-agent/src/index.ts#L274-L281

